PR #3, opened 2023-01-26
cleanup some lints

bnewbold:
Nothing urgent here, just getting CI to "green" with current `go vet -a ./...` linting.

The "comment out a big function which isn't used" is definitely debatable. Would be fine adding a CI override comment or something there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,8 @@ test: build ## Run all tests
 
 .PHONY: lint
 lint: ## Run style checks and verify syntax
-	go vet -a ./...
+	go vet -asmdecl -assign -atomic -bools -buildtag -cgocall -copylocks -httpresponse -loopclosure -lostcancel -nilfunc -printf -shift -stdmethods -structtag -tests -unmarshal -unreachable -unsafeptr -unusedresult ./...
 	test -z $(gofmt -l ./...)
-	# TODO: golangci-lint
 
 .PHONY: fmt
 fmt: ## Run syntax re-formatting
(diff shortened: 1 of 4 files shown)


carstore/bs.go
```diff
@@ -124,7 +124,7 @@ func (uv *userView) Get(ctx context.Context, k cid.Cid) (blocks.Block, error) {
 		Where("usr = ? AND cid = ?", uv.user, k.String()).
 		First(&info).Error; err != nil {
 		if err == gorm.ErrRecordNotFound {
-			return nil, ipld.ErrNotFound{k}
```
whyrusleeping, line 127: I'm really not a fan of forcing keys in cases like this, definitely a style thing but it just feels cleaner not to

whyrusleeping:
Let's go with this vet call for now:
```
go vet -asmdecl -assign -atomic -bools -buildtag -cgocall -copylocks -httpresponse -loopclosure -lostcancel -nilfunc -printf -shift -stdmethods -structtag -tests -unmarshal -unreachable -unsafeptr -unusedresult ./...
```

whyrusleeping:
I also find errcheck (https://github.com/kisielk/errcheck) helpful, but running it in CI is usually a bad time expenditure (tons of false positives)

carstore/bs.go
bnewbold: what about in the case of multiple keys? like:

```
err := shared.TokenPump{TokenSource: dec, TokenSink: enc}.Run()
```
(`cmd/gosky/main.go:722`)

bnewbold:
Got a spurious test failure: https://github.com/bluesky-social/indigo/actions/runs/4020248065/jobs/6907952478

```
panic: read tcp 127.0.0.1:57326->127.0.0.1:8231: use of closed network connection
```
passed on second run (re-ran job)

whyrusleeping (approved):

PR #9, opened 2023-02-03
update lexicons (2023-02-02)

bnewbold:
This is a bigger change, deprecates/deletes some custom code, and isn't complete, so passing along for review.

The main thing is updates of api/bsky/* and api/atproto/* based on today's upstream Lexicon files. I didn't run lexgen on the new com.atproto.admin and com.atproto.reports schemas, because lexgen failed on some new corner cases and features. I did update lexgen to handle a subset of new patterns (like CAR file output, with corresponding mimetype). I also used the `atproto` and `bsky` package names, respectively, getting rid of `schemagen`.

One the destructive changes here is removing a bunch of "peering" lexicons, which are not in upstream. The codegen'd files are in git history, so we could pull these back in the future. It seems like these lexicons were not wired up, though maybe there just isn't any test coverage touching on dynamic typing code paths.

There is also a breaking change going from `com.atproto.sync.getRoot` to `com.atproto.sync.getHead`; will need to update (or deprecate) some hand-rolled code to finish that transition.

I didn't (yet) update ./gen/main.go to do CBOR gen on any new types.


whyrusleeping:
Changes to lexgen look good to me, assuming the rest is just running the codegen, yeah?

bnewbold:
@whyrusleeping: yeah, basically a clean codegen (wiping all existing codegen code first), then small tweaks to get things to compile.

@devinivy: right! should have thought to do that before: https://github.com/bluesky-social/indigo/issues/10

bnewbold:
(rebased and merged manually)

PR #15, opened 2023-02-12
start refactoring away old `api/*.go` clients

bnewbold:
I did some big chunks of `gosky` and a bit of `stress`. Was able to get rid of the `BskyApp` client type.

The `testscript` worked, but i'm not confident that there are not runtime bugs in the other parts of `gosky` (I didn't test all the individual commands).

There is a bunch more of this to do; the `ATProto` client type is used all over. A good start would be refactoring the existing methods on that client (in `api/atproto.go`) to use the lexgen stuff and work with `xrpc.Client`. That is, turn them in to high-level helpers, particularly for account+session management, and record creation/deletion. And maybe post creation, because that is so common in existing code with just text+createdAt (not the full-featured record type).

Sending this for review for your thoughts on the above, as much as for review of the changes.

### DIFF
--- a/api/bsky.go
+++ b/api/bsky.go
@@ -1,17 +1,5 @@
 package api
 
-import (
-	"context"
-	"fmt"
-
-	bsky "github.com/bluesky-social/indigo/api/bsky"
-	"github.com/bluesky-social/indigo/xrpc"
-)
-
-type BskyApp struct {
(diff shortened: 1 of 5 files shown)


bnewbold:
merged manually, not sure why this didn't close automatically

PR #23, opened 2023-02-16
MST/repo interop tests; updates to implementation from typescript

bnewbold:
~The tests here work with atproto (Typescript) and adenosine (Rust). They are mostly not working here in golang, mostly because of differences in DAG-CBOR CID formatting (it seems).~

~Also probably going to add some repo-layer tests before this gets merged.~

Updated and now MST tests are passing. Some `testing` (integration) tests are failing in CI, but run ok for me locally. Hrm.

Updated code from typescript implementation. Added cross-language interop tests, with exact CID values. Removed some redundant code and unused struct fields (not sure why `go vet` or similar didn't complain about those).

### DIFF
--- a/mst/mst.go
+++ b/mst/mst.go
@@ -2,103 +2,53 @@
 // This implementation is a port of the Typescript implementation in the
 // `atproto` git repo.
 //
-// The hash function is SHA-256, and "fanout=16", which corresponds to counting
-// "leading zeros" for tree level in 4-bit chunks.
+// ## Notes copied from TS repo
+//
+// This is an implementation of a Merkle Search Tree (MST)
+// The data structure is described here: https://hal.inria.fr/hal-02303490/document
+// The MST is an ordered, insert-order-independent, deterministic tree.
+// Keys are laid out in alphabetic order.
+// The key insight of an MST is that each key is hashed and starting 0s are counted
+// to determine which layer it falls on (5 zeros for ~32 fanout).
+// This is a merkle tree, so each subtree is referred to by it's hash (CID).
+// When a leaf is changed, ever tree on the path to that leaf is changed as well,
+// thereby updating the root hash.
+//
+// For atproto, we use SHA-256 as the key hashing algorithm, and ~16 fanout
+// (4-bits of zero per layer).
+//
+// NOTE: currently keys are strings, not bytes. Because UTF-8 strings can't be
+// safely split at arbitrary byte boundaries (the results are not necessarily
+// valid UTF-8 strings), this means that "wide" characters not really supported
+// in keys, particularly across programming language implementations. We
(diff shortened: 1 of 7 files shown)


bnewbold:
I've rebased this and it is ready for review/signoff. To summarize the changes:

- re-ordered functions and files to match upstream Typescript implementation. Eg, created new `mst_util.go` file to match `mst/util.ts` file
- also updated implementation details to match upstream Typescript. Eg, now properly "trims" top-of-tree, and generates "empty" nodes in tree between layers when needed
- removed flexibility around MST "fanout". This is now effectively hardcoded at fanout=16 (4bit "zero" counting), to match what we are doing in prod
- add cross-language interop tests to verify bit-for-bit compatibility with Typescript implementation
- removed a couple redundant or unused functions and fields on structs

mst/mst.go
```diff
@@ -270,35 +439,113 @@ func (mst *MerkleSearchTree) Add(ctx context.Context, key string, val cid.Cid, k
...
 
 		checkTreeInvariant(updated)
-		newRoot, err := create(ctx, mst.cst, updated, keyZeros, mst.fanout)
-		if err != nil {
-			return nil, fmt.Errorf("creating new tree after split: %w", err)
-		}
+		newRoot := NewMST(mst.cst, cid.Undef, updated, keyZeros)
 
-		// why invalidate?
+		// we have mutated this node without recomputing CID, so invalidate it
```
whyrusleeping, line 444: its technically a no-op, we set validPtr to be whether or not the cid we give it is defined, and here we're giving it an explicitly undefined cid

mst/mst.go
```diff
@@ -346,109 +598,105 @@ func (mst *MerkleSearchTree) Delete(ctx context.Context, k string) (*MerkleSearc
...
 	}
 
-	return cid.Undef, ErrNotFound
-}
-
-func (mst *MerkleSearchTree) createParent(ctx context.Context) (*MerkleSearchTree, error) {
-	layer, err := mst.getLayer(ctx)
-	if err != nil {
-		return nil, err
+	// TODO(bnewbold): same as Typescript, but shouldn't this error instead of returning null?
```
whyrusleeping, line 673: Its a bit weird, but to match typescript the closest ive let the zero value of NodeEntry.Kind be 'undefined', so we can have the same explicitly undefined checks that the typescript code does.
If i was implementing this from scratch and not from referenced i'd definitely use typed errors

whyrusleeping (approved):
quick pass over, this looks good. If you want me to dive in deeper i can, but idk if thats necessary at this point.

bnewbold:
Updated comment, rebased on current main. Will merge if/when CI checks pass.

PR #27, opened 2023-02-21
bigsky deployment infra

bnewbold:
- dockerfile
- github action to build and upload container

### DIFF
--- a/cmd/bigsky/Dockerfile
+++ b/cmd/bigsky/Dockerfile
@@ -29,3 +29,7 @@ COPY --from=build-env /bigsky /
 EXPOSE 2470
 
 CMD ["/bigsky"]
+
+LABEL org.opencontainers.image.source=https://github.com/bluesky-social/indigo
+LABEL org.opencontainers.image.description="ATP Big Graph Server (BGS)"
+LABEL org.opencontainers.image.licenses=MIT
--- a/cmd/bigsky/Dockerfile
+++ b/cmd/bigsky/Dockerfile
@@ -14,7 +14,8 @@ RUN apk add --no-cache build-base make go
 ADD . /dockerbuild
 WORKDIR /dockerbuild
 
-RUN go build -o /bigsky ./cmd/bigsky
+# timezone data for alpine builds
+RUN go build -tags timetzdata -o /bigsky ./cmd/bigsky
 
 ### Run stage
 FROM alpine:3.17
@@ -25,6 +26,10 @@ ENTRYPOINT ["dumb-init", "--"]
 WORKDIR /
 RUN mkdir -p data/bigsky
 COPY --from=build-env /bigsky /
+
+# small things to make golang binaries work well under alpine
+ENV GODEBUG=netdns=go
+ENV TZ=Etc/UTC
 
 EXPOSE 2470
 
--- a/cmd/bigsky/Dockerfile
+++ b/cmd/bigsky/Dockerfile
@@ -1,0 +1,31 @@
+# Run this dockerfile from the top level of the indigo git repository like:
+#
+#   podman build -f ./cmd/bigsky/Dockerfile -t bigsky .
+
+### Compile stage
+
+# trouble with 'No such file or directory' when building using 'golang:1.19'
+# images, I think because of library errors when copied to 'alpine' image below
+# https://megamorf.gitlab.io/2019/09/08/alpine-go-builds-with-cgo-enabled/
+# TODO: lock to a specific version of go toolchain
+FROM alpine:3.17 AS build-env
+RUN apk add --no-cache build-base make go
+
+ADD . /dockerbuild
+WORKDIR /dockerbuild
+
+RUN go build -o /bigsky ./cmd/bigsky
+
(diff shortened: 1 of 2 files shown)


bnewbold:
I have some other tweaks for test-ability and deploy-ability of the bigsky application itself, I will probably put all that in a separate PR and have this one be just the Dockerfile. I've been developing using `podman` instead of `docker` locally and it is working fine, but i'm mostly agnostic about which we actually go with.

It feels like we could simplify things by having the container include all the deployable executables, and let deployment scripts keep the `ENTRYPOINT` but override the `CMD` on a per-deployment basis to select which service to run. On the other hand, that might be complex/confusing for third party folks who just want to run a single service. Hrm.

bnewbold:
Split out config var names and actual code changes to a separate PR: https://github.com/bluesky-social/indigo/pull/29

cmd/bigsky/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN go build -o /bigsky ./cmd/bigsky
+
+### Run stage
+FROM alpine:3.17
+
+RUN apk add --no-cache --update dumb-init
+ENTRYPOINT ["dumb-init", "--"]
+
+WORKDIR /
+RUN mkdir -p data/bigsky
```
Jacob2161, line 26: We'll mount a volume like `--volume=/data:/data` into the container which will mask any directories created  in here. We should also be able to specify a `BIGSKY_DATA_DIR=/data/bigsky` or similar env I think, so that it's not hard coded.

cmd/bigsky/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+FROM alpine:3.17 AS build-env
+RUN apk add --no-cache build-base make go
+
+ADD . /dockerbuild
+WORKDIR /dockerbuild
+
+RUN go build -o /bigsky ./cmd/bigsky
+
+### Run stage
+FROM alpine:3.17
```
Jacob2161, line 20: We're building bigsky with cgo so are there any gotchas with running in alpine w/musl? I can't remember. I seem to recall wanting to do a few things like `export GODEBUG=netdns=go` when using CGO. But if things are working okay we may be able to look into this later.

Jacob2161: Maybe we need to install the tzdata package too? We should probably be building with `-tags timetzdata` https://pkg.go.dev/time/tzdata

Also, there's nothing keeping us on alpine I don't think. We can afford 100-200 MB container images if we run into annoyances. I do like Alpine though.

Jacob2161:
> I've been developing using `podman` instead of `docker` locally and it is working fine

I'm excited about podman I've just judged it to be not quite ready for prime time, after having tried to use it a few times.

The Docker runtime/daemon has some disadvantages but it really does Just Work at this point. Podman is still catching up but I think we should definitely expect to migrate to it at some point.

Jacob2161 (approved):
:+1:

bnewbold:
> > I've been developing using `podman` instead of `docker` locally and it is working fine
> 
> I'm excited about podman I've just judged it to be not quite ready for prime time, after having tried to use it a few times.
> 
> The Docker runtime/daemon has some disadvantages but it really does Just Work at this point. Podman is still catching up but I think we should definitely expect to migrate to it at some point.

Yeah, I wouldn't recommend running `podman` in prod, or for doing container builds. I've run in to issues with it trying to do docker-compose like stuff. I am liking it on my dev laptop locally for simple cases, but we should consider docker the canonical tool I think.

cmd/bigsky/Dockerfile
bnewbold: Yup, have an env var to move this. And the current implementation will create the directory structure if needed on boot (`mkdir -p` equivalent). Partially just put that in as documentation.

`bigsky` mostly uses this for "carstore" (repo storage on disk as CAR files), but by default it will also use sqlite database at this location. It will only look for postgresql if configured to do so via database connection string.

cmd/bigsky/Dockerfile
bnewbold: These are good tips, i'll look in to these build flags and config vars. When deploying `beemo` to render I noticed the `netdns` thing but didn't dig in to it.

I'm a heavy debian user, and often like to have a regular linux distro so I have debug tools at hand in production if needed. In this case with a single binary it seemed nice and easy to keep image size small, and I mostly copied this from other templates that use alpine. And small size would be appealing for self-hosters. Can split between "our deployment dockerfile" and "public dockerfile" in the future if needed.

Jacob2161:
Would be cool to get this in, even if it's not perfect, so we get containers building. I'm sure there will be things to fix but shouldn't be anything major.

bnewbold:
- added container metadata (like for PDS in atproto repo)
- removed the build+push action for this branch, only do it on main (that was temporary while developing)
- add some golang build flags and env vars for running under alpine

This branch has no code changes, is reviewed, and is ready to merge. I'm only holding off until we resolve `main` being broken right now.

PR #29, opened 2023-02-21
standardize config and app setup; rename `cmd/pds` to `cmd/laputa`

bnewbold:
We had discussed these changes briefly. Probably worth a one-over.

I'm fairly opinionated about using the `DATABASE_URL` convention, don't care as much about the other variables.

### DIFF
--- a/cmd/gosky/util/util.go
+++ b/cmd/gosky/util/util.go
@@ -150,6 +150,13 @@ func ReadAuth(fname string) (*xrpc.AuthInfo, error) {
 	return &auth, nil
 }
 
+// Supports both previous "dbtype=" prefixed DSNs, and URI-style database config strings, for both sqlite and postgresql.
+//
+// Examples:
+// - "sqlite=dir/file.sqlite"
+// - "sqlite://file.sqlite"
+// - "postgres=host=localhost user=postgres password=password dbname=pdsdb port=5432 sslmode=disable"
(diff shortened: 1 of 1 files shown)


cmd/beemo/main.go
```diff
@@ -113,7 +118,7 @@ func pollNewReports(cctx *cli.Context) error {
 	xrpcc.Auth.Did = auth.Did
 	xrpcc.Auth.Handle = auth.Handle
 
-	adminToken := cctx.String("admin-token")
+	adminToken := cctx.String("admin-password")
```
whyrusleeping, line 121: is it a password?

bnewbold: Yeah, and it was confusing initially when I thought it was a token. The way it works is HTTP `Basic` auth (not `Bearer`) with username `admin` and this as the password.

We are planning to switch and use regular auth (the JWT tokens, as HTTP `Bearer` auth), plus an admin flag on accounts, but this is how it works today.

Some interfaces, like redsky, do say "token" though.

bnewbold:
Updated the `DATABASE_URL` helper to accept the `sqlite=` and `postgres=` (DSN) syntax from before. I don't have test coverage for this, but did test all the variations locally. Will end up with test coverage when I add postgres tests in CI.

@whyrusleeping said in slack "other than that it seems good"

whyrusleeping (approved):

bnewbold:
I merged and pushed this, the ordering was weird and github was confused and marked as closed.

PR #30, opened 2023-02-23
Add a basic healthcheck endpoint

Jacob2161:
Just need some endpoint to healthcheck against so the instance can be put into service by the load balancer. Obviously we want a more complete healthcheck whenever possible.


bnewbold:
here is my take with quick DB ping: https://github.com/bluesky-social/indigo/pull/32

Jacob2161:
See #32

PR #32, opened 2023-02-24
add /xrpc/_health checks to bigsky and pds

bnewbold:
These do a trivial connection to database to ensure things are actually running.


whyrusleeping (approved):
LGTM

PR #34, opened 2023-02-27
add 'make all' to actually generate executables

whyrusleeping:


Makefile
```diff
@@ -8,6 +8,16 @@ help: ## Print info about all commands
...
 
+.PHONY: all
+all: 
+	go build ./cmd/gosky
+	go build ./cmd/laputa
+	go build ./cmd/bigsky
+	go build ./cmd/beemo
+	go build ./cmd/lexgen
+	go build ./cmd/stress
+	go build ./cmd/fakermaker
```
bnewbold, line 19: I would just put these under the existing `make build` instead of `make all`?

Would there be any benefit to also doing `go build ./...` at the start of the list, to ensure that any hypothetical new cmds get at least built before we add them to this list? I think that would cache all the builds so it isn't any duplicate work. Probably over thinking things, might be wrong.

bnewbold (approved):
We could put these under `make build` instead, but i'm also fine if you want to keep it this way.

Makefile
whyrusleeping: That would work, though I'll probably also want to alias all->build to satisfy my muscle memory

PR #35, opened 2023-02-28
Remove deprecated manual helper methods from `api/atproto.go` and `api/bsky.go`

bnewbold:
A handful of small refactors to use `xrpc.Client` and codegen'd type directly, then was able to nuke these files entirely.

This gives me confidence about our ability to run codegen, get things to compile and tests to pass, and expect it to actually work with Typescript PDS implementation.


whyrusleeping (approved):
NICE

PR #36, opened 2023-02-28
bigsky Dockerfile: include ca-certificates

bnewbold:
This is to ensure that we have correct (or at least reasonable) SSL root certs when the BGS makes outbound connections to crawl PDS instances.


Jacob2161 (approved):
:+1:

PR #38, opened 2023-03-01
Add 'repo ops' back into events

whyrusleeping:
This isnt ready to merge yet, still waiting on the upstream lexicon updates, but it *is* backwards compatible with the existing lexicons so it works against staging still

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -160,51 +160,58 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
 		since = &sval
 	}
 
+	ctx := c.Request().Context()
+
 	// TODO: authhhh
 	conn, err := websocket.Upgrade(c.Response().Writer, c.Request(), c.Response().Header(), 1<<10, 1<<10)
 	if err != nil {
 		return fmt.Errorf("upgrading websocket: %w", err)
 	}
 
-	evts, cancel, err := bgs.events.Subscribe(func(evt *events.RepoStreamEvent) bool { return true }, since)
+	evts, cancel, err := bgs.events.Subscribe(ctx, func(evt *events.RepoStreamEvent) bool { return true }, since)
 	if err != nil {
 		return err
 	}
 	defer cancel()
 
 	header := events.EventHeader{Op: events.EvtKindRepoAppend}
-	for evt := range evts {
-		wc, err := conn.NextWriter(websocket.BinaryMessage)
-		if err != nil {
-			return err
-		}
+	for {
+		select {
+		case evt := <-evts:
+			wc, err := conn.NextWriter(websocket.BinaryMessage)
+			if err != nil {
+				return err
+			}
 
-		var obj util.CBOR
+			var obj util.CBOR
 
-		switch {
-		case evt.Append != nil:
-			header.Op = events.EvtKindRepoAppend
-			obj = evt.Append
-		case evt.Error != nil:
-			header.Op = events.EvtKindErrorFrame
-			obj = evt.Error
-		case evt.Info != nil:
-			header.Op = events.EvtKindInfoFrame
-			obj = evt.Info
-		default:
-			return fmt.Errorf("unrecognized event kind")
-		}
+			switch {
+			case evt.Append != nil:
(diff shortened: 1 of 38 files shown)


cmd/gosky/main.go
```diff
@@ -823,6 +824,29 @@ var resetPasswordCmd = &cli.Command{
 	},
 }
 
+var updateHandleCmd = &cli.Command{
```
bnewbold, line 827: It would be nice to pull out commits like this into separate PRs. And the `--auth` revert to a seperate commit.

indexer/keymgr.go
```diff
@@ -48,7 +48,7 @@ func (km *KeyManager) getKey(ctx context.Context, did string) (*did.PubKey, erro
 		return nil, err
 	}
 
-	pubk, err := doc.GetPublicKey("#signingKey")
```
bnewbold, line 51: Another (important!) fix that would be good in a separate PR or at least commit

bgs/bgs.go
```diff
@@ -116,10 +124,13 @@ func (bgs *BGS) HandleHealthCheck(c echo.Context) error {
...
-	Handle string `gorm:"uniqueIndex"`
-	Did    string `gorm:"uniqueIndex"`
-	PDS    uint
+	ID        bsutil.Uid `gorm:"primarykey"`
+	CreatedAt time.Time
+	UpdatedAt time.Time
+	DeletedAt gorm.DeletedAt `gorm:"index"`
+	Handle    string         `gorm:"uniqueIndex"`
+	Did       string         `gorm:"uniqueIndex"`
+	PDS       uint
```
bnewbold, line 133: Will the gorm auto-migrate code handle this sort of non-trivial migration? Or will we need to wipe state in staging to deploy this (`DROP DATABASE` and rebuild)?

As we move to having stuff deployed we should both add "deployment concerns" sections or comments to PR messages, so it is clear how and when they can go out.

cmd/gosky/util/util.go
```diff
@@ -152,6 +152,8 @@ func SetupDatabase(dburl string) (*gorm.DB, error) {
 	var dial gorm.Dialector
 	// NOTE(bnewbold): might also handle file:// as sqlite, but let's keep it
 	// explicit for now
+
+	openConns := 99
```
bnewbold, line 156: I think this is pretty high for postgresql by default? I know there have been improvements in recent years, but I think the postgresql design philosophy is to keep connections to a couple dozen at max and go with `pg_bounce` if needed. Probably fine on our beefy deployed instances. I would have gone with 20 or 30.

util/uid.go
```diff
@@ -0,0 +1,3 @@
+package util
+
+type Uid uint
```
bnewbold, line 3: :+1: big fan of this sort of thing

bnewbold (approved):
Looks reasonable over all. I am curious what you think of that gorm migration: can we just deploy this as-is to staging?

My comments here are mostly opinions (about postgresql connection counts and git workflow).

util/uid.go
whyrusleeping: Yeah, it makes messing things up a lot harder lol

cmd/gosky/util/util.go
whyrusleeping: this is really just copy/paste from my estuary code, picked it there on recommendation of some perf guides somewhere, totally not opinionated about it

bgs/bgs.go
whyrusleeping: yeah this is actually not even a migration, its all the same fields, just one of them is getting a named type instead of a bare `uint`

cmd/gosky/main.go
whyrusleeping: Yeah, i'm moving towards that, but things are still in the phase of needing a bunch of things to land all at the same time to stay up to date.

bgs/bgs.go
bnewbold: The `ID` field is new, and the primary key; will it be auto-generated on existing rows in the database?

Also all the timestamp fields are new, and don't seem to be nullable. Will they default to "now"?

PR #47, opened 2023-03-06
fix typo introduced by PR #40 where io package was used instead of os

erka:


erka:
Hey @whyrusleeping. Is it possible to run checks for pull-requests too?

bnewbold:
@erka yeah we can enable actions for `pull_request` events.

PR #48, opened 2023-03-07
ci: run actions on pull_request

bnewbold:
This seems to be the correct and safe way to enable github actions CI runs for third party forks (when they are submitted as a PR).

The important detail is that the `.github/*` action stuff does not run from the fork, it runs from the previous `main` (or whatever branch) that the fork branched off from. This should prevent exfiltration of secrets?

Still possible for folks to, eg, mine cryptocurrency on our dime, I think. The current behavior from the github UI side should be that first-time contributors need to be approved before the CI job runs.

### DIFF
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -1,5 +1,9 @@
 name: golang
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -1,9 +1,15 @@
 name: golang
+
 on:
   pull_request:
   push:
     branches:
       - main
+
+concurrency:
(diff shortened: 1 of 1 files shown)


bnewbold:
Hrm, this seems to be running CI actions twice :crying_cat_face:

.github/workflows/golang.yml
```diff
@@ -1,5 +1,5 @@
 name: golang
-on: [push]
+on: [push, pull_request]
```
erka, line 2:
```suggestion
on:
  push:
    branches:
    - main
  pull_request:
    branches:
    - main

concurrency:
  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
  cancel-in-progress: true
```
Please try this

bnewbold:
I checked against what `ripgrep` does (https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/ci.yml), and also merged in erka's concurrency control config.

Seems to work correctly now, thanks for the tips!

PR #49, opened 2023-03-07
Get signing key from PLC using new did doc semantics

devinivy:
In https://github.com/bluesky-social/did-method-plc/pull/5 there were some updates to how the did doc is laid out in PLC.  I was experiencing an issue with signature checks, and this small update seems to have done the trick.

For good measure, here's the error that was surfacing:
```
2023-03-07T10:59:57.689-0500    ERROR   indexer indexer/crawler.go:158  failed to perform repo crawl of "did:plc:wqzurwm3kmaig6e6hnc2gqwo": importing fetched repo: process new repo: cb errored (54/54): signature check failed: no key found by that ID:
```

PR #50, opened 2023-03-07
Get service from PLC using new did doc semantics

dholms:
Inspired by https://github.com/bluesky-social/indigo/pull/49

We also change the semantics of how the PDS is specified


bnewbold:
I haven't gone back to look at all the other PLC changes to see if there are other things that need to change, but this LGTM.

dholms:
Last related thing is the new canonical way of finding the handle is:

"first value in `alsoKnownAs` that starts with `at://`"

PR #52, opened 2023-03-08
Fix typo in HACKING.md

akiomik:


whyrusleeping:
realistically we should just remove this whole section and recommend people use the `gosky readStream` utility

akiomik:
I see, if you don't need to fix the typo, you could close this.

bnewbold:
I think this is still potentially useful for debugging, especially until the event schema fully stabilizes. I'll merge and we can pull that section out separately.

PR #55, opened 2023-03-08
use dbcid type for carstore cid references

whyrusleeping:
This goes from 59 bytes per cid in the db to 36, a pretty hefty savings.

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -63,7 +63,7 @@ type UserInfo struct {
 type CarShard struct {
 	gorm.Model
 
-	Root      string
+	Root      util.DbCID
 	DataStart int64
 	Seq       int `gorm:"index"`
 	Path      string
@@ -71,8 +71,8 @@ type CarShard struct {
 }
 
 type blockRef struct {
-	ID     uint   `gorm:"primarykey"`
-	Cid    string `gorm:"index"`
+	ID     uint       `gorm:"primarykey"`
+	Cid    util.DbCID `gorm:"index"`
 	Shard  uint
 	Offset int64
 	//User   uint `gorm:"index"`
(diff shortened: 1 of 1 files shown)


bnewbold (commented):
Nice! Not sure if you wanted a full review yet, but early notes.

A particularly good win on the `blockRef` table (because all the other columns are already somewhat compact), and because there is an index on this value (so saves on index space).

There is a bit of a tradeoff with ease of manual SQL debugging (will get bytes back not simple string), but seems justified in this case as we expect BGS to get scale pretty large pretty fast. 

What is the migration path for existing databases? Eg, it would be possible to iterate over the rows and rewrite from string to bytes, but I don't see any code implementing that. Possible i'm confused. We still have a window here where we can get away with just nuking the BGS database on deploy. But we should at least mention this kind of thing in the PR message.

While we are touching these schemas, could we set explicit sizes for the `int` and `uint` columns, aka `int32` or `int64`? Eg, to be more explicit about whether we expect to get a postgres `integer` (32bit) or `bigint` (64bit). Could also add the column type as an annotation if you don't want to change the struct types.

whyrusleeping:
I dont think we should worry about migrating existing databases, just nuke them and start over

Not sure its necessary to be explicit about the bitwidth of the types, gorm handles it all under the hood and ive never had a problem with it. 
Are there any cases we specifically want a 32 bit integer?

whyrusleeping:
As for the manual debugging issues, I think our solution should probably be to just add extra utilities for debugging to gosky as we need them.

bnewbold:
I default to 64bit ints in databases ("BIGINT") unless i'm trying to shave bytes in a specific well-understood context.

I feel pretty strongly that we should understand the specific SQL schema being generated (column names and types) and not have the ORM abstract it away too much. In this case it seems like it should be an easy thing to add annotations.

Once this is running in prod my understanding is that nuking the entire database would be fairly disruptive to the ecosystem, even if we don't wipe the carstore. At a minimum it would drop the recovery window for subscribers, require re-polling all dids/handles, rediscovering (or being polled by) all PDS instances. A totally reasonable disaster recovery thing, but a non-trivial service degradation? That indicates to me that we are going to need to handle more sophisticated migrations in the future without nuking, and that will probably mean knowing exactly what the underlying SQL schema is before and after the migration, in addition to things like indices, views, tablespaces, and other operational levers for scaling and reliability.

whyrusleeping:
Sure, but we arent in prod yet. Thats why i'm doing this PR now, so we can move fast and not waste time dealing with migrations later

bnewbold (approved):

PR #57, opened 2023-03-10
repo v2 support

bnewbold:
A couple high-level notes:

- reasonably confident in MST changes. not very confident in repo node changes (eg, `repo/` and `repomgr/` sections). I basically just updated the structs and got things to compile, the entire `repo` package could probably be simplified
- there is at least one call-out for closer review of some code I ripped out (in repomgr)
- I haven't verified that the signature process validates against atproto repo signatures (or vica versa)
- the MST changes also include new constraints on keys (paths): total length <= 256 bytes; must have a single '/'; small subset of ASCII chars allowed in keys
- I implemented the []byte / string mapping the same way as the typestring package, which means most of the logic is still string-based not bytes-based; conversation happens at the last moment when serializing/deserializing


repo/repo.go
```diff
@@ -17,38 +19,57 @@ import (
...
+}
+
+// returns SHA-256 (as direct bytes, not hex-encoded string bytes) of the
+// DAG-CBOR representation of object. This is what gets signed.
+func (uc *UnsignedCommit) BytesForSigning() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	if err := uc.MarshalCBOR(buf); err != nil {
+		return []byte{}, err
+	}
+	h := sha256.Sum256(buf.Bytes())
```
whyrusleeping, line 69: I believe the signing library hashes the input already, need to verify

whyrusleeping: Confirmed (go-did/key.go.(PrivKey).Sign) hashes the input with sha256 before signing internally.
I'm open to doing that elsewhere, but it seems less error prone there

repomgr/repomgr.go
```diff
@@ -957,33 +967,7 @@ func (rm *RepoManager) processNewRepo(ctx context.Context, user util.Uid, r io.R
...
-			return err
-		}
-
-		if err := membs.Put(ctx, blk); err != nil {
-			return err
-		}
-
-		membs = util.NewReadThroughBstore(robs, membs)
-	}
+	// TODO(bnewbold): @Why should review the code that was removed here
```
whyrusleeping, line 970: This is good, this was the hack necessary for the meta object being a separate thing. I'm only sad that i wasnt the one to delete it ;)

whyrusleeping (changes requested):
This all looks good except for the hashing of the signable content

repo/repo.go
bnewbold: Ok, I will pull this hash out.

mst/mst.go
```diff
@@ -191,8 +191,9 @@ func (mst *MerkleSearchTree) getEntries(ctx context.Context) ([]NodeEntry, error
 func entriesFromNodeData(ctx context.Context, nd *NodeData, cst cbor.IpldStore) ([]NodeEntry, error) {
 	layer := -1
 	if len(nd.Entries) > 0 {
+		// NOTE(bnewbold): can compute the layer on the first KeySuffix, because for the first entry that field is a complete key
 		firstLeaf := nd.Entries[0]
-		layer = leadingZerosOnHash(firstLeaf.KeySuffix)
+		layer = leadingZerosOnHash(string(firstLeaf.KeySuffix))
```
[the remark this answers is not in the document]
bnewbold: Yup! As just discussed, keeping it as directly similar to the Typescript implementation for now, will refactor to be more idiomatic golang in probably not too long.

repomgr/repomgr.go
bnewbold: Ok, I'll pull this comment out.

bnewbold:
@whyrusleeping verification change LGTM, thanks for adding the test.

I'm ready for this to merge if you are.

whyrusleeping (approved):

PR #60, opened 2023-03-11
Labelmaker v0 / "first phase"

bnewbold:
This is an updated/refactored version of the trivial keyword-based labelmaker. I tested it locally to work directly with Typescript PDS ("subscribeAllRepos"), and can see websocket events coming out of this labelmaker end-to-end.

I'm submitting this for review to @whyrusleeping mostly w/r/t how i'm linking in with existing BGS slurper and events code. Eg, should we refactor the event processing code to be more generic (across repo events and label events)? Especially things like the non-trivial persist implementaitons.

Also, for any feedback on the label event schema.

Things that will happen in next phases (which I will start building on top of this branch this weekend), but not in this PR:

- calls out to APIs during the label process, eg for NSFW image labeling and SQRL integration
- concurrent label processing, especially in the context of API calls
- possibly removing or changing repo persistence of labels. might use an existing PDS implementation for simplicity (and to remove need for a local disk for carstore)
- Lexicon additions
- productionized: docs, Dockerfile, metrics, logging

### DIFF
--- a/labeling/labelers.go
+++ b/labeling/labelers.go
@@ -3,16 +3,15 @@ package labeling
 import (
 	"strings"
 
-	"github.com/bluesky-social/indigo/api"
-	bskyapi "github.com/bluesky-social/indigo/api/bsky"
+	bsky "github.com/bluesky-social/indigo/api/bsky"
 )
 
 // simple record labeling (without pre-fetched blobs)
 type SimplePostLabeler interface {
-	labelPost(p api.PostRecord) []string
+	labelPost(p bsky.FeedPost) []string
 }
 type SimpleActorProfileLabeler interface {
-	labelActorProfile(ap bskyapi.ActorProfile) []string
+	labelActorProfile(ap bsky.ActorProfile) []string
 }
 
 type KeywordLabeler struct {
@@ -30,11 +29,11 @@ func (kl KeywordLabeler) labelText(txt string) []string {
 	return []string{}
 }
 
-func (kl KeywordLabeler) labelPost(p api.PostRecord) []string {
+func (kl KeywordLabeler) labelPost(p bsky.FeedPost) []string {
 	return kl.labelText(p.Text)
 }
 
-func (kl KeywordLabeler) labelActorProfile(ap bskyapi.ActorProfile) []string {
+func (kl KeywordLabeler) labelActorProfile(ap bsky.ActorProfile) []string {
 	txt := ap.DisplayName
 	if ap.Description != nil {
 		txt += *ap.Description
--- a/labeling/labelers.go
+++ b/labeling/labelers.go
@@ -1,0 +1,43 @@
+package labeling
+
+import (
+	"strings"
+
+	"github.com/bluesky-social/indigo/api"
+	bskyapi "github.com/bluesky-social/indigo/api/bsky"
+)
+
+// simple record labeling (without pre-fetched blobs)
+type SimplePostLabeler interface {
+	labelPost(p api.PostRecord) []string
+}
+type SimpleActorProfileLabeler interface {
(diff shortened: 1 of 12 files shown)


labeling/labelers.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"strings"
+
+	bsky "github.com/bluesky-social/indigo/api/bsky"
+)
+
+// simple record labeling (without pre-fetched blobs)
+type SimplePostLabeler interface {
+	labelPost(p bsky.FeedPost) []string
```
whyrusleeping, line 11: Can you emit a label for an actor based on one of their posts?

labeling/service.go
```diff
@@ -0,0 +1,268 @@
...
+
+	go levtman.Run()
+
+	return s, nil
+}
+
+func (s *Server) SubscribeBGS(ctx context.Context, bgsUrl string, useWss bool) {
+	// subscribe our RepoEvent slurper to the BGS, to receive incoming records for labeler
+	log.Infof("subscribing to BGS: %s (SSL=%v)", bgsUrl, useWss)
+	s.bgsSlurper.SubscribeToPds(ctx, bgsUrl, useWss)
```
whyrusleeping, line 109: oh yeah probably could rename that method

labeling/labelers.go
bnewbold: Yes. This initial version only outputs labels on the specific record, but the overall architecture allows outputting labels on either on records or on repos (DIDs). For example, if this was the Nth label on a record in the same repo we could have logic that would output a repo-level label.

An open question is how/whether to label pieces of infrastructure, like a PDS. There is an idea of a way to label DNS names (so, a PDS at a stable domain). If infrastructure has stable DIDs we could label those, but I don't think we've thought that through all the way (aka, how to we prevent infra from just cycling through DIDs; do we really want to require stable DIDs for infra in the first place).

labeling/service.go
bnewbold: Yes, maybe `SubscribeToRepoHost`. But low priority.

labeling/labelers.go
```diff
@@ -0,0 +1,42 @@
...
+}
+type SimpleActorProfileLabeler interface {
+	labelActorProfile(ap bsky.ActorProfile) []string
```
whyrusleeping, line 14: Why have an interface with unexported methods?

labeling/labelers.go
whyrusleeping: Okay, separate question is how can i load the blobs from this post in order to use them as context for label decisions?
Also should probably wire contexts through all of this (esp if we're going to be making external API calls at some point)

bnewbold: Fetching blobs is presumably going to be a multi-step process of resolving the DID to PDS (cachable), and then fetching the blob with `getBlob`. The atproto branch with getBlob implementation hasn't landed, it is here: https://github.com/bluesky-social/atproto/pull/606

I did a manual rebase locally of that branch to test things, and ran in to the mentioned issue with records seemingly missing blob metadata.

You can see my branch with external blob (image) fetches here: https://github.com/bluesky-social/indigo/pull/61

labeling/labelers.go
bnewbold: This keyword labeling is kind of throw-away just as an example for testing and integration, but I can expose these methods.

labeling/service.go
bnewbold: https://github.com/bluesky-social/indigo/issues/64 so this doesn't get lost

labeling/labelers.go
bnewbold: made public

bnewbold:
Merging as per discussion

PR #61, opened 2023-03-12
labelmaker: external API calls

bnewbold:
Making calls out to APIs to help with labeling determinations.

This round of work will include a basic (not production ready) NSFW image filter, and SQRL API. Everything is still single-threaded, not concurrent.

### DIFF
--- a/labeling/keyword_labeler.go
+++ b/labeling/keyword_labeler.go
@@ -1,6 +1,10 @@
 package labeling
 
 import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	bsky "github.com/bluesky-social/indigo/api/bsky"
@@ -15,15 +19,15 @@ type SimpleActorProfileLabeler interface {
 }
 
 type KeywordLabeler struct {
-	keywords []string
-	value    string
+	Keywords []string `json:"keywords"`
+	Value    string   `json:"value"`
 }
 
 func (kl KeywordLabeler) LabelText(txt string) []string {
 	txt = strings.ToLower(txt)
-	for _, word := range kl.keywords {
+	for _, word := range kl.Keywords {
 		if strings.Contains(txt, word) {
-			return []string{kl.value}
+			return []string{kl.Value}
 		}
 	}
 	return []string{}
@@ -39,4 +43,26 @@ func (kl KeywordLabeler) LabelActorProfile(ap bsky.ActorProfile) []string {
 		txt += *ap.Description
 	}
 	return kl.LabelText(txt)
+}
+
+func LoadKeywordFile(fpath string) ([]KeywordLabeler, error) {
+
+	var kwl []KeywordLabeler
+
+	jsonFile, err := os.Open(fpath)
+	if err != nil {
+		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
+	}
+	defer jsonFile.Close()
+
+	raw, err := io.ReadAll(jsonFile)
+	if err != nil {
+		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
+	}
+
+	if err = json.Unmarshal(raw, &kwl); err != nil {
+		return kwl, fmt.Errorf("failed to parse Keyword file: %v", err)
+	}
+
+	return kwl, nil
 }
--- a/labeling/keyword_labeler.go
+++ b/labeling/keyword_labeler.go
@@ -43,17 +43,17 @@ func LoadKeywordFile(fpath string) ([]KeywordLabeler, error) {
 
 	jsonFile, err := os.Open(fpath)
 	if err != nil {
-		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
+		return nil, fmt.Errorf("failed to load JSON file: %v", err)
 	}
 	defer jsonFile.Close()
 
 	raw, err := io.ReadAll(jsonFile)
 	if err != nil {
-		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
+		return nil, fmt.Errorf("failed to load JSON file: %v", err)
 	}
 
 	if err := json.Unmarshal(raw, &kwl); err != nil {
-		return kwl, fmt.Errorf("failed to parse Keyword file: %v", err)
+		return nil, fmt.Errorf("failed to parse Keyword file: %v", err)
 	}
 
 	return kwl, nil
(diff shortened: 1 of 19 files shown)


bnewbold:
Rebased this and it is ready for review.

I tested with https://github.com/bluesky-social/indigo/pull/67 cherry-picked on top of this and it worked; this branch passes tests but won't actually work for blobs until that fix lands.

The makefile changes in https://github.com/bluesky-social/indigo/pull/65 are also helpful for testing this code.

labeling/micro_nsfw_img.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, fmt.Errorf("micro-NSFW-img request failed  statusCode=%d", res.StatusCode)
+	}
+
+	respBytes, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read micro-NSFW-img resp body: %v", err)
+	}
+
+	var nsfwScore MicroNSFWImgResp
+	if err = json.Unmarshal(respBytes, &nsfwScore); err != nil {
```
whyrusleeping, line 103: best practice to do `if err := json....` here, that way you arent accidentally changing an external scope value

labeling/service.go
```diff
@@ -129,33 +147,181 @@ func (s *Server) wantAnyRecords(ctx context.Context, ra *events.RepoAppend) bool
...
 	log.Infof("labeling record: %v", uri)
 	var labelVals []string
+	var blobs []lexutil.Blob
 	switch nsid {
 	case "app.bsky.feed.post":
 		post, suc := rec.(*appbsky.FeedPost)
 		if !suc {
 			return []string{}, fmt.Errorf("record failed to deserialize from CBOR: %s", rec)
 		}
+		/* XXX(bnewbold): debugging broken post record
```
whyrusleeping, line 173: delete?

labeling/service.go
```diff
@@ -129,33 +147,181 @@ func (s *Server) wantAnyRecords(ctx context.Context, ra *events.RepoAppend) bool
...
+			}
+		}
 	case "app.bsky.actor.profile":
 		profile, suc := rec.(*appbsky.ActorProfile)
 		if !suc {
 			return []string{}, fmt.Errorf("record failed to deserialize from CBOR: %s", rec)
 		}
+		// XXX(bnewbold): debugging broken profile record
+		profileJson, _ := json.Marshal(profile)
+		log.Infof("labeling profile: %v", string(profileJson))
```
whyrusleeping, line 209: cleanup if it works now

labeling/service.go
```diff
@@ -129,33 +147,181 @@ func (s *Server) wantAnyRecords(ctx context.Context, ra *events.RepoAppend) bool
...
+		}
+		if profile.Banner != nil {
+			blobs = append(blobs, *profile.Banner)
+		}
+	}
+
+	log.Infof("will process %d blobs", len(blobs))
+	for _, blob := range blobs {
+		if blob.Cid == "" {
+			return []string{}, fmt.Errorf("received stub blob (CID undefined)")
```
whyrusleeping, line 241: any particular reason not to just return nil? In an error case you arent going to use the value

labeling/service.go
```diff
@@ -129,33 +147,181 @@ func (s *Server) wantAnyRecords(ctx context.Context, ra *events.RepoAppend) bool
...
+		labelVals = append(labelVals, blobLabels...)
+	}
+	return dedupeStrings(labelVals), nil
+}
+
+func (s *Server) downloadRepoBlob(ctx context.Context, did string, blob *lexutil.Blob) ([]byte, error) {
+	var blobBytes []byte
+
+	if blob.Cid == "" {
+		return []byte{}, fmt.Errorf("invalid blob to download (CID undefined)")
```
whyrusleeping, line 269: similarly, just return `nil, err`

labeling/service.go
```diff
@@ -129,33 +147,181 @@ func (s *Server) wantAnyRecords(ctx context.Context, ra *events.RepoAppend) bool
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return []byte{}, fmt.Errorf("failed to fetch blob from PDS. did=%s cid=%s statusCode=%d", did, blob.Cid, resp.StatusCode)
+	}
+
+	blobBytes, err = io.ReadAll(resp.Body)
+	if err != nil {
+		return blobBytes, nil
```
whyrusleeping, line 293: return the error

labeling/util.go
```diff
@@ -45,3 +45,16 @@ func loadKey(kfile string) (*did.PrivKey, error) {
 		Type: out,
 	}, nil
 }
+
+// NOTE(bnewbold): I don't know of a more idiomatic way to do this in golang
+func dedupeStrings(in []string) []string {
```
whyrusleeping, line 50: this is fairly typical. If its performance sensitive or in a fast path there are better ways, but simple is better

labeling/service.go
bnewbold: whoops!

labeling/service.go
bnewbold: good catch!

labeling/util.go
bnewbold: removed comment

labeling/micro_nsfw_img.go
bnewbold: updated this an similar cases

labeling/service.go
bnewbold: updated this and similar cases

bnewbold:
Updated all comments. Good eye on the fat-fingered incorrect returns in error handling!

labeling/keyword_labeler.go
```diff
@@ -1,42 +1,60 @@
...
+
+	jsonFile, err := os.Open(fpath)
+	if err != nil {
+		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
+	}
+	defer jsonFile.Close()
+
+	raw, err := io.ReadAll(jsonFile)
+	if err != nil {
+		return kwl, fmt.Errorf("failed to load JSON file: %v", err)
```
whyrusleeping, line 52: definitely not blocking, but its always a bit awkward when i see a non-nil value along with an error return. It signals to me that you intend to use the error *and* the value, when the intention here is to ignore the value in the error case. It draws the eye unnecessarily

whyrusleeping (approved):

PR #62, opened 2023-03-12
add 'too big' event flag to events

whyrusleeping:
We don't have a lexicon for this yet, but this does prevent things from breaking outright for now.


bnewbold:
Looks fine to me. I assume you'll update if needed when the lexicon change gets settled.

PR #63, opened 2023-03-12
Misc debugging improvements

whyrusleeping:


cmd/bigsky/main.go
```diff
@@ -35,7 +35,7 @@ var log = logging.Logger("bigsky")
 
 func init() {
 	//logging.SetAllLoggers(logging.LevelDebug)
-	logging.SetAllLoggers(logging.LevelWarn)
+	logging.SetAllLoggers(logging.LevelInfo)
```
bnewbold, line 38: I made a note for myself to look in to this. Want to ensure that we don't clobber `GOLOG_LOG_LEVEL` env var, but also have basically all commands default to `info` verbosity.

events/events.go
```diff
@@ -140,7 +141,10 @@ type ErrorFrame struct {
 	Message string `cborgen:"message"`
 }
 
-func (em *EventManager) AddEvent(ev *RepoStreamEvent) error {
+func (em *EventManager) AddEvent(ctx context.Context, ev *RepoStreamEvent) error {
```
bnewbold, line 144: I think this will impact the labelmaker version, but I can handle that change in my code de-dupe branch. May need to check that there are no build errors after rebasing this branch.

bnewbold: I tested rebasing locally, no build problem.

bnewbold (approved):
A couple small things (println -> log, just delete some commented lines). Otherwise LGTM.

PR #65, opened 2023-03-12
makefile: dev runners; fast `make build`; lexgen

bnewbold:
Switches `make build` back to `go build ./...` for speed. `make binaries` builds all the executables, and `make all` is an alias of that. Maybe a bit non-standard; `go build ./...` could be `make check` instead? But this is my current muscle memory.

Adds a `make lexgen` helper which looks in `../atproto/lexicons/` by default, but can be overridden by `LEXDIR` environment variable.

Adds `make run-dev-bgs` to match current behavior of `make run-pds` in `atproto` repo. This is quite helpful to me for getting local integeration dev work going. Note that the `GOLOG_LOG_LEVEL` environment variable can't currently be definted in `.env` config file because of import ordering.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,7 @@ help: ## Print info about all commands
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "    \033[01;32m%-20s\033[0m %s\n", $$1, $$2}'
 
 .PHONY: build
-build: ## Build everything (to Go cache, not binaries)
-	go build ./...
-
-.PHONY: binaries
-binaries: ## Build all executables
+build: ## Build all executables
 	go build ./cmd/gosky
 	go build ./cmd/laputa
 	go build ./cmd/bigsky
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,17 +2,21 @@
...
 	@echo "Commands:"
 	@echo
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "    \033[01;32m%-20s\033[0m %s\n", $$1, $$2}'
 
-.PHONY: all
-all: build
-
 .PHONY: build
-build: ## Build all executables
+build: ## Build everything (to Go cache, not binaries)
```
whyrusleeping, line 15: but why?

bnewbold: `go build ./...` was running much faster for me on my laptop, and would find errors in all code, including any binaries not yet including the link of specific binaries to build. 

When I tried to time this just now I couldn't reproduce the different in build times though, it was only a second or two of different. So I can rip this part of the changes out if you think not worth it.

whyrusleeping: yeah, i'm just really used to `make build` building everything and being able to run the binaries. Maybe we can add a `make compile` or `make check-compile` type thing to do what you're after? Its kinda misleading to have build not produce any artifacts

bnewbold:
Rebased with `make check` and added labelmaker to the build list

PR #68, opened 2023-03-14
merged EventManager

bnewbold:
This combines the RepoAppend and LabelBatch event managers (and MemPersisters) into a single implementation. The database persister and some other code is not sharable.

After implementing this, I realized that the "kind" integers probably can't be a single numeric space because there would be a coordination problem with folks proposing random new lexicons and having conflicts. So may have to refactor this again before label stuff stabilizes.

The `Append -> RepoAppend` rename was for clarity, but it got large and disruptive. I don't feel super strongly about that if you want me to rever it.

The `XRPCStreamEvent` (to replace `RepoStreamEvent`) was just the first name I thought of, can do a quick search/replace if you prefer something else.


whyrusleeping (approved):
LGTM!

PR #70, opened 2023-03-15
Add a concurrent event processor helper

whyrusleeping:
Havent integrated this yet, but I figured it would be good to send up for a review.
Pretty straightforward to make this generic so we can use it for labels when we need to as well.

### DIFF
--- a/events/parallel.go
+++ b/events/parallel.go
@@ -1,0 +1,89 @@
+package events
+
+import (
+	"context"
+	"sync"
+)
+
+type ConsumerPool struct {
+	maxConcurrency int
+	maxQueue       int
+
+	do func(*RepoAppend)
+
+	feeder chan *consumerTask
+
+	lk     sync.Mutex
+	active map[string][]*consumerTask
+}
+
+func NewConsumerPool(maxC, maxQ int, do func(*RepoAppend)) *ConsumerPool {
+	p := &ConsumerPool{
+		maxConcurrency: maxC,
+		maxQueue:       maxQ,
+
+		do: do,
+
+		feeder: make(chan *consumerTask),
+		active: make(map[string][]*consumerTask),
+	}
+
+	for i := 0; i < maxC; i++ {
+		go p.worker()
+	}
+
+	return p
+}
+
+type consumerTask struct {
+	key string
+	val *RepoAppend
+}
+
+func (p *ConsumerPool) Add(ctx context.Context, key string, val *RepoAppend) error {
+	t := &consumerTask{
+		key: key,
+		val: val,
+	}
+	p.lk.Lock()
+
+	a, ok := p.active[key]
+	if ok {
+		p.active[key] = append(a, t)
+		p.lk.Unlock()
+		return nil
+	}
+
+	p.active[key] = []*consumerTask{}
+	p.lk.Unlock()
+
+	select {
+	case p.feeder <- t:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (p *ConsumerPool) worker() {
+	for work := range p.feeder {
+		for work != nil {
+			p.do(work.val)
+
+			p.lk.Lock()
+			rem, ok := p.active[work.key]
+			if !ok {
+				log.Errorf("should always have an 'active' entry if a worker is processing a job")
+			}
+
+			if len(rem) == 0 {
+				delete(p.active, work.key)
+				work = nil
+			} else {
+				work = rem[0]
+				p.active[work.key] = rem[1:]
+			}
+			p.lk.Unlock()
+		}
+	}
+}
--- a/events/parallel.go
+++ b/events/parallel.go
@@ -9,7 +9,7 @@ type ConsumerPool struct {
 	maxConcurrency int
 	maxQueue       int
 
-	do func(*RepoAppend)
+	do func(*XRPCStreamEvent)
 
 	feeder chan *consumerTask
 
@@ -17,7 +17,7 @@ type ConsumerPool struct {
 	active map[string][]*consumerTask
 }
 
-func NewConsumerPool(maxC, maxQ int, do func(*RepoAppend)) *ConsumerPool {
+func NewConsumerPool(maxC, maxQ int, do func(*XRPCStreamEvent)) *ConsumerPool {
 	p := &ConsumerPool{
 		maxConcurrency: maxC,
 		maxQueue:       maxQ,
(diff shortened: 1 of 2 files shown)


events/parallel.go
```diff
@@ -0,0 +1,89 @@
...
+type ConsumerPool struct {
+	maxConcurrency int
+	maxQueue       int
+
+	do func(*RepoAppend)
+
+	feeder chan *consumerTask
+
+	lk     sync.Mutex
+	active map[string][]*consumerTask
```
bnewbold, line 17: What is the key for? To keep separate lists per PDS being consumed from? I would have expected something like `map[util.Uid][]int64`, that is, a list of outstanding seq from all the PDSes being consumed from.

bnewbold (commented):
I'm not sure i'm understanding the structure here right. Is the intent of the `key` to have separate queues per PDS?

What I would have expected is for the workers to receive events from a single channel, along with end-of-channel (or context error) for termination. Process locally in goroutine, and then public that it is done working to a "done" channel. So the worker goroutines themselves never try to take a mutex lock.

The progress state that would to be tracked is two lists of sequence ids per PDS (or other upstream infra host): a list of received-but-not-persisted sequence numbers ("inflight"), and a list of processed-but-not-persisted sequence numbers ("outoforder"). Some goroutine, which could be a main event thread or some other thread, would consume from the "done" channel. If the most recent "done" for a given PDS is the oldest in the "inflight", that gets popped, all the "outoforder" are checked against the head of "inflight", and the consume state updated for that PDS. If it isn't the oldest "inflight", it gets inserted into the "outoforder" queue (while maintaining sort order, so maybe actually a `map[int64]bool` or something). If we are clever about goroutines this persist goroutine might not even require any mutex, but a mutex is also fine.

Can we send entire `XRPCStreamEvent` objects instead of `RepoAppend`? They have some extra context (like the source repo), and would make this generic across RepoAppend and LabelBatch.

There should be a helper on `ConsumerPool` to do graceful shutdown, that is, stop pushing any new tasks but wait for any outstanding to complete and shutdown.

events/parallel.go
whyrusleeping: The key is the did of the user, to ensure we dont process events for a given repo in parallel

whyrusleeping:
We can send whatever type you want on this, i was going to make it just generic but didnt want to futz with that at first.

> What I would have expected is for the workers to receive events from a single channel, along with end-of-channel (or context error) for termination. Process locally in goroutine, and then public that it is done working to a "done" channel. 

Thats mostly correct, youre free to push into a 'done' channel as you like in the callback, but for most cases i'm imagining its not necessary (gathering results into a single place via a channel isnt doing much for you since you've lost ordering anyways).

The tricky bits here is just ensuring that multiple events for a given key are processed in the same order they arrive.

ericvolp12:
I implemented a similar worker pattern for consuming events from the repoSync socket [here](https://github.com/ericvolp12/bsky-experiments/blob/main/pkg/events/event_callbacks.go#L107).
One thing I ran into which you might want to be aware of, the Merkel tree on a `RepoRecord` isn't threadsafe and I was getting data race warnings from Go:
```


|  |2023-04-29 01:50:07 | Goroutine 72 (running) created at:
  |   | 2023-04-29 01:50:07 | /app/pkg/events/event_callbacks.go:94 +0x47 |  
  |   | 2023-04-29 01:50:07 | github.com/ericvolp12/bsky-experiments/pkg/events.NewBSky.func1() |  
  |   | 2023-04-29 01:50:07 | /app/pkg/events/worker.go:60 +0x37b |  
  |   | 2023-04-29 01:50:07 | github.com/ericvolp12/bsky-experiments/pkg/events.(*BSky).worker() |  
  |   | 2023-04-29 01:50:07 | /app/pkg/events/worker.go:77 +0x171 |  
  |   | 2023-04-29 01:50:07 | github.com/ericvolp12/bsky-experiments/pkg/events.(*BSky).ProcessRepoRecord() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/repo/repo.go:333 +0x35d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/repo.(*Repo).GetRecord() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:487 +0x38d |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:467 +0x84 |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).Get() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:895 +0x71 |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).findGtOrEqualLeafIndex() |  
  |   | 2023-04-29 01:50:07 | /go/pkg/mod/github.com/bluesky-social/indigo@v0.0.0-20230428230924-93b159a0ea0d/mst/mst.go:183 +0x272 |  
  |   | 2023-04-29 01:50:07 | github.com/bluesky-social/indigo/mst.(*MerkleSearchTree).getEntries() |  
```

Not 100% sure if it's relevant but thought I'd call it out in case you run into something similar.

PR #71, opened 2023-03-16
fix issue with gap filling in index crawler

whyrusleeping:
The source of the bug is the one line change in the carstore, the repomanager changes are the part that could use the closest look, basically i'm allowing the first commit in the gap fill process to dip into the carstore for the blocks it needs for the diff, later diffs are isolated and must pull blocks from the car file.


bnewbold:
I'll take a look at this later tonight or tomorrow.

whyrusleeping:
@bnewbold i'd still like review here but i'm going to merge this so we can get it deployed

PR #72, opened 2023-03-16
some extra tweaks required for fully handling 'too big' slices

whyrusleeping:
The last PR handled this for playback but missed catching when they happen on the fly, also dropped some log warnings to infos so I could see what was even going wrong.


whyrusleeping:
The log messages i demoted are still relevant, they are mostly failures to resolve certain user references, which i *think* are mostly scenes, but hard to say for sure.

bnewbold (approved):
LGTM

PR #73, opened 2023-03-17
Feat/blob syncing

whyrusleeping:


bgs/blobs.go
```diff
@@ -0,0 +1,29 @@
+package bgs
```
bnewbold, line 1: Should you pull this out into a top-level `blobstore` package? Seems like it could be reused by the PDS.

events/dbpersist.go
```diff
@@ -63,6 +63,12 @@ func (p *DbPersistence) Persist(ctx context.Context, e *XRPCStreamEvent) error {
 
 	evt := e.RepoAppend
 
+	// TODO: hack hack hack
```
bnewbold, line 66: Do you mean to ship this, even as a temporary thing?

I would pass through a "too large", or drop event, or remove all ops instead of this truncation.

events/repostream.go
```diff
@@ -11,14 +11,18 @@ import (
...
 )
 
-type LiteStreamHandleFunc func(op repomgr.EventKind, path string, did string, rcid *cid.Cid, rec any) error
+type LiteStreamHandleFunc func(op repomgr.EventKind, seq int64, path string, did string, rcid *cid.Cid, rec any) error
 
 func ConsumeRepoStreamLite(ctx context.Context, con *websocket.Conn, cb LiteStreamHandleFunc) error {
 	return HandleRepoStream(ctx, con, &RepoStreamCallbacks{
 		RepoAppend: func(evt *RepoAppend) error {
+			if evt.TooBig {
+				log.Errorf("skipping too big events for now: %d", evt.Seq)
```
bnewbold, line 20: More context would be helpful in this log line: at least the repo and commit (which are in the RepoAppend).

util/client.go
```diff
@@ -0,0 +1 @@
+package util
```
bnewbold, line 1: ?

bnewbold (approved):
A couple small notes, if you resolve those the rest looks good.

Based on conversation this morning (Friday), doesn't seem entirely clear whether we want/need BGS to be holding all the blobs, but there are other relevant updates in here, and the blobstore code would be reusable for PDS anyways.

bgs/blobs.go
whyrusleeping: Yeah, that makes sense.

events/dbpersist.go
whyrusleeping: Yeah, the issue is that we currently don't have an official way of recovering from this. If we drop them then a consumer will get zero events instead of 'some', and for the purposes of the search engine work we're pushing on I would rather have some than none in the short term

util/client.go
whyrusleeping: meant to delete, thanks for pointing out

PR #76, opened 2023-03-20
add datadog trace exporting to bigsky

whyrusleeping:
I *think* this is right, but its annoyingly hard to test outside of having a full datadog setup locally


cmd/bigsky/main.go
```diff
@@ -135,6 +138,21 @@ func run(args []string) {
...
+				log.Fatal(err)
+			}
+			defer exporter.Stop()
+
+			trace.RegisterExporter(exporter)
+
+			// For demoing purposes, always sample.
+			trace.ApplyConfig(trace.Config{
+				DefaultSampler: trace.AlwaysSample(),
+			})
```
bnewbold, line 153: Feels to me like this should be a CLI/env variable from the get-go (aka, `--datadog-trace-fraction` or something like that; can default to 1.0).

cmd/bigsky/main.go
```diff
@@ -135,6 +138,21 @@ func run(args []string) {
 			otel.SetTracerProvider(tp)
 		}
 
+		if cctx.Bool("datadog") {
```
bnewbold, line 141: Should at least have this be specified in the flag list, I don't think this args should be hidden. Also putting it in the flags lets an `EnvVars` be configured, which is how we will probably configure this up when deployed.

Do we not need to specify a sidecar endpoint or anything? Is that handled automatically by the datadog library? Would be good to document any such env vars in `./cmd/bigsky/README.md`, even if just for Jake and myself.

cmd/bigsky/main.go
```diff
@@ -17,9 +17,11 @@ import (
 	"github.com/bluesky-social/indigo/notifs"
 	"github.com/bluesky-social/indigo/plc"
 	"github.com/bluesky-social/indigo/repomgr"
+	"go.opencensus.io/trace"
 
 	_ "net/http/pprof"
 
+	datadog "github.com/DataDog/opencensus-go-exporter-datadog"
```
bnewbold, line 24: This repo says:

> This repository is deprecated and is in maintenance mode.

But it does also say that bugs are getting fixed, and the alternative linked to seems more complex and "beta" status. Did you chose this one intentionally?

whyrusleeping:
this isnt the right way of doing things, jake found a better option

PR #78, opened 2023-03-21
Additional dockerfiles and github actions: beemo and labelmaker

bnewbold:
Labelmaker will get deployed in the next week or so.

Beemo is currently running fine on render, but at some point we might want it deployed as a container. I could disable that container build by default if we want to hold off on it.

LMK if I should switch these up to use AWS instead of ghcr.io.

At some point maybe we could abstract these to be less duplication between files, but seems fine for now.

### DIFF
--- a/.github/workflows/container-beemo-ghcr.yaml
+++ b/.github/workflows/container-beemo-ghcr.yaml
@@ -1,0 +1,52 @@
+name: container-beemo-ghcr.yaml
+on:
+  push:
+    branches:
+      - main
+env:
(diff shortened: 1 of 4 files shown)


Jacob2161 (approved):
1. Agreed, the copy/pasta is getting heavy with these across all the repos/projects. Build once and push to multiple registries seems good enough. Only issue w/that is making sure the names are correct as they differ a bit between AWS and GitHub.
2. It'd be good to get all images pushing to both GitHub and AWS going forward. Not a blocker though.

PR #81, opened 2023-03-21
add some metrics and set up otlp exporting

whyrusleeping:


cmd/bigsky/main.go
```diff
@@ -64,6 +66,9 @@ func run(args []string) {
 		&cli.BoolFlag{
 			Name: "jaeger",
 		},
+		&cli.BoolFlag{
+			Name: "otlp-export",
```
Jacob2161, line 70:
```suggestion
			Name: "otlp-export",
			EnvVars: []string{"OTLP_EXPORT"},
```

cmd/bigsky/main.go
```diff
@@ -110,9 +115,15 @@ func run(args []string) {
 		&cli.StringFlag{
 			Name: "disk-blob-store",
 		},
+		&cli.StringFlag{
+			Name:    "env-type",
```
Jacob2161, line 119:
```suggestion
			Name:    "env-type",
			Usage: "environment type (test, staging, production)",
```

Jacob2161 (changes requested):
I've already added the receiver part for datadog-agent.

With this change (I added one env var)  we should be able to deploy with this for e.g. production:

```bash
BGS_ENVIRONMENT=production
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
OTLP_EXPORT=true
```

whyrusleeping:
this needs to be rebased and updated to use the latest opentrace metrics package

whyrusleeping:
this was merged in a different PR

PR #82, opened 2023-03-22
HTTP retry helper; add as default for XRPC, beemo, labelmaker, fakermaker

bnewbold:
The initial motivation for this was for beemo to not crash on the very first HTTP 5xx it encounters, but to do a couple retries. An example where we expect occasional 5xx errors is when deploying some of our own services, or running complex database migrations.

I initially was going to hand-roll something simple like https://brandur.org/fragments/go-http-retry, with fixed backoffs, but it seems like the hashicorp lib is fairly light and slots in as a regular `http.Client`, which is nice.

Curious your thoughts on when/where to do this instead of just crashing and having the service restart. I think there are reasonable arguments in a variety of directions; having automatic retries everywhere for 429 and 502/503 has served me well. This could, technically, result in duplicate requests when the backend handled things fine and it was the load-balancer which returned a 5xx error.

I also didn't copy over all the transport configs you have in `gosky`, but we could include most of those. I don't think we should force HTTP/2 in the general case.

### DIFF
--- a/cmd/beemo/main.go
+++ b/cmd/beemo/main.go
@@ -27,10 +27,12 @@ import (
 var log = logging.Logger("beemo")
 
 func main() {
-	run(os.Args)
+	if err := run(os.Args); err != nil {
+		log.Fatal(err)
+	}
 }
 
-func run(args []string) {
+func run(args []string) error {
 
 	app := cli.App{
 		Name:    "beemo",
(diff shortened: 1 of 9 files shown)


bnewbold:
@whyrusleeping said "seems good but I would maybe do a longer timeout? Though i guess if its always run on a server somewhere that is plenty".

I think for general use (eg, CLIs) the timeout is already pretty long. Going to merge as-is, can revist/bump as needed.

PR #85, opened 2023-03-22
error out if we dont read the entire response body

whyrusleeping:
probably should also handle chunked encoding..? Though i'm not convinced that should be allowed in xrpc

### DIFF
--- a/xrpc/xrpc.go
+++ b/xrpc/xrpc.go
@@ -117,9 +117,16 @@ func (c *Client) Do(ctx context.Context, kind XRPCRequestType, inpenc string, me
 
 	if out != nil {
 		if buf, ok := out.(*bytes.Buffer); ok {
-			_, err := io.CopyN(buf, resp.Body, resp.ContentLength)
-			if err != nil {
-				return fmt.Errorf("reading response body: %w", err)
+			if resp.ContentLength < 0 {
+				_, err := io.Copy(buf, resp.Body)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


bnewbold:
In golang, for clients receiving a response, this field can be `-1` if the content length isn't known. For the golang field, this might be because the content was de-compressed (configurable, and indicated by a flag), or presumably because the header just wasn't set by the server. Docs: https://pkg.go.dev/net/http#Response

So, we at least need to handle that case with a regular `Copy`, possibly with an overall size limit.

I did a quick hunt and was surprised that it isn't obvious that the net/http library would raise an error on content-length mismatches automatically. I see `ErrShortBody` but it is deprecated. The `CopyN()` technique should solve short reads when there is a valid `ContentLength` field.

As-is, this client code will ignore "too long" bodies. I sort of feel like we should thow or or at least log a warning on that condition, if we are experiencing issues with body sizes. I think the server can sometimes notice that the body wasn't entirely read, but not all conditions (depending on whether the body had already been transmitted?).

I guess depends on what specific problem we are trying to solve. I am surprised that this isn't handled natively by the HTTP library.

devinivy:
The other case I would be interested in here is what happens if the response is compressed, which some servers and intermediaries perform automatically for responses over a certain size.  In that case the content-length header would match the compressed rather than the uncompressed size.

bnewbold (approved):
This whole thing continues to be crazy, but these changes seem correct for now, and an improvement.

Issue #56, opened 2023-03-09
remaining did:plc changes


bnewbold:
Already handled a couple changes in https://github.com/bluesky-social/indigo/pull/50 and https://github.com/bluesky-social/indigo/pull/49.

@dholms mentioned:

> Last related thing is the new canonical way of finding the handle is: "first value in alsoKnownAs that starts with at://"

We should update indigo to match that.

bnewbold:
This is implemented in `indigo/syntax/identifier`

PR #88, opened 2023-03-23
properly deal with handle changes we dont get an event for

whyrusleeping:


bgs/bgs.go
```diff
@@ -400,6 +401,18 @@ func (s *BGS) syncUserBlobs(ctx context.Context, pds *models.PDS, user bsutil.Ui
 	return nil
 }
 
+func handleFromDidDoc(doc *did.Document) (string, error) {
```
bnewbold, line 404: Related to https://github.com/bluesky-social/indigo/issues/56, we should technically be checking that the AlsoKnownAs value starts with `at://`. That is, iterate through the list until we hit a URI starting with `at://` and take that one.

bgs/bgs.go
```diff
@@ -477,6 +484,59 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
...
+						return nil, err
+					}
+
+					// TODO: it is a bit weird that we keep the handle in multiple locations...
+					if err := s.db.Model(models.ActorInfo{}).Where("uid = ?", ou.ID).Update("handle", ohandle).Error; err != nil {
+						return nil, err
+					}
+
+					// Now that thats resolved, try creating the user again?
+					return s.lockedCreateUser(ctx, handle, did, peering.ID)
```
bnewbold, line 522: Small thing, but I think we should log even the happy path when this happens (not just error paths). Eg, we should log that it seems like handle changed, and include the prior handle and DID.

bgs/bgs.go
```diff
@@ -494,12 +554,12 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
...
 		Did:    did,
-		PDS:    peering.ID,
+		PDS:    pdsid,
 	}
 
 	if err := s.db.Create(&u).Error; err != nil {
-		// some debugging...
-		return nil, fmt.Errorf("failed to create other pds user: %w", err)
+		log.Warnf("failed to create new user record: %s", err)
+		return nil, ErrRevalidateUserHandle
```
bnewbold, line 562: It isn't obvious to me why we would always try to re-validate if the row insertion failed. Eg, couldn't the database error be a deadlock or connection dropped or something other than a constraint violation? Feels like this needs something like the `if !errors.Is(err, gorm.ErrRecordNotFound)` just above.

bnewbold (approved):
Left some comments, the rest looks good.

bgs/bgs.go
whyrusleeping: yeah, There didnt seem to be an easy way to check that when i looked through the gorm docs otherwise i would have

bnewbold: From https://github.com/go-gorm/gorm/blob/master/errors.go, it looks like `ErrDuplicatedKey` would at least handle the "there is already such a handle" UNIQUE constraint violation. If that is the primary situation where the creation would fail and we need to re-validate.

whyrusleeping:
I think i've accidentally made this redundant in #106

PR #89, opened 2023-03-24
capture git revision at build time

### LINKED ISSUE #31: capture git revision at build time, and inject as needed
Eg, CLI `--version`; HTTP client `User-Agent`; etc.

The git invocation I have used before is:

```
git describe --tags --long --always
```

Need to take care in docker builds to grab this while `.git` is available.

erka:
Fixes #31


erka:
Hey @bnewbold. 

Please take a look on my changes.

I suggest to add the `v0.0.0` tag to have something like `v0.0.0-0-g9272eac`

bnewbold:
This is looking good to me, thanks! You even updated the `User-Agent` strings.

I'm not sure about the `internal/engine` package name, though this is basically bikeshedding. I think @whyrusleeping does not like `internal` package namespaces. Maybe under `util/` or `meta/`? Let's see what he thinks.

whyrusleeping:
maybe just call it build?

bnewbold:
`github.com/bluesky-social/indigo/build` or `github.com/bluesky-social/indigo/version` would both make sense to me. `hashicorp/consul` does `version/` and we sometimes copy them.

We will probably iterate on this a bit, but a good start as-is. We will probably do a `v0.1.0` tagged release soon.

erka:
I renamed it to `version`. `build` is great option, but it is a common pattern for compiled assets. Also we have `build` in the `.gitignore`.

PR #90, opened 2023-03-25
handle 'genesis' events such as rebases

whyrusleeping:
I'm sure this will change slightly after the lexicon refactor lands, but this is still strictly better for this codepath

### DIFF
--- a/events/repostream.go
+++ b/events/repostream.go
@@ -13,12 +13,38 @@ import (
 
 type LiteStreamHandleFunc func(op repomgr.EventKind, seq int64, path string, did string, rcid *cid.Cid, rec any) error
 
-func ConsumeRepoStreamLite(ctx context.Context, con *websocket.Conn, cb LiteStreamHandleFunc) error {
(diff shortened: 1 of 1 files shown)


events/repostream.go
```diff
@@ -13,12 +13,38 @@ import (
 
 type LiteStreamHandleFunc func(op repomgr.EventKind, seq int64, path string, did string, rcid *cid.Cid, rec any) error
 
-func ConsumeRepoStreamLite(ctx context.Context, con *websocket.Conn, cb LiteStreamHandleFunc) error {
+type GetRepoFunc func(ctx context.Context, did string, oldest *string, newest string) ([]byte, error)
```
bnewbold, line 16: I'd be curious to see an example implementation of this signature. Seems like it would want a reference to a cache of DID lookups (`did:plc` or `did:web`), or at least a DID PLC API client.

Not a blocking thing, we can add those optimizations later.

bnewbold (approved):

whyrusleeping:
gonna close this, its been effectively redone in other PRs

Issue #83, opened 2023-03-22
Lex refactor needs


bnewbold:
Cross-reference to main upstream (atproto) branches: https://github.com/bluesky-social/atproto/pull/658

Things that will require new lexgen feature support:

- [x] `type="cid-link"`, encodes as an object like `{ "$link": "<cid>" }`. For binary CIDs, aka "Link" in IPLD data model)
- [x] `type="bytes"`, encodes as an object like `{ "$bytes": "<base64-encoded>" }` for embedded bytes (which specific base64 config?)
- [x] no image/video/audio, just `type="blob"` now. `blob` gains a `size` field (integer, size in bytes)
- [x] `type="number"` changed to `type="float"` (though not yet used, and might change how that works)
- [ ] string types, which will work like `type="string" format="at-uri"`

The string types are not strictly needed to get lexgen to run, but we probably might as well wire up placeholder validation for functions in codegen code. The formats are at least: `handle`, `nsid`, `at-identifier` (handle or DID), `uri` (generic), `datetime`, `at-uri`, `did`, `cid`. I can port over the validation functions and their tests. I guess we haven't actually implemented the other validators yet, like `maxLength`. Maybe we punt on validation until a second pass (separate PR).

There are some semantics/benavioral changes, like idempotency of batch mutations and `deleteRecord` not failing if record already does not exist. I don't think the existing golang PDS is complete enough to care about those.

Once lexgen runs successfully, presumably there will be a bunch of broken stuff that we'll just need to get to compile.

cc: @whyrusleeping

whyrusleeping:
can we close this now?

bnewbold:
The string formatting stuff hasn't been touched yet, AFIAK. Might factor in to an iteration of lexgen? Also things like `maxLength` (both strings and arbitrary arrays), max sizes for blobs (in bytes), max grapheme lengths.

We could create a new issue to track that, but this issue has good summary info.

bnewbold:
closing this old/stale issue; lex refactor was forever ago.

https://github.com/bluesky-social/indigo/pull/449 is a PR with some partial/draft work on incorporating `indigo/atproto/syntax` package in to lexgen.

"grapheme length validation" will probably end up in `indigo/atproto/data` or `indigo/atproto/lexicon`.

PR #93, opened 2023-03-27
Lex refactor branch ("last big breaking lexicon changes")

whyrusleeping:
Working towards #83 mostly

bigsky and laputa compile, havent tested them yet. Still a fair emount left to do


bnewbold:
I guess I spoke too soon: the tests pass locally for me (including in a fresh checkout), but not in CI. It is possible that golang 1.19 vs. 1.20 is the issue (I've updated on my laptop; this PR hasn't landed: https://github.com/bluesky-social/indigo/pull/91). But more likely some timing issue.

bnewbold:
Some more progress, all tests are passing now. BGS interop with typescript lex-refactor branch is not working yet.

I hacked in the subscription changes pretty quickly, I suspect that more refactoring might be needed in BGS, repomgr, etc. The previous `RepoAppend` object could be both a `repo_append` or a `rebase`, based on the `Event` field value; now the different event types are different structs. Maybe we should pass around entire `XRPCStreamEvent` structs? This is just a vague impression, I just "got things to compile" without digging in to how the rebase vs. commit code paths work.

Tweaked a couple lexgen things relating to required-but-nullable fields. This is only implemented on structs for now, not handler function signatures or query params.

Created an issue to track post-refactor follow-ups: https://github.com/bluesky-social/indigo/issues/95

I have hit a couple different errors trying to do BGS interop with typescript implementation. One error looks like:

```
2023-03-29T21:16:33.850-0700	ERROR	indexer	indexer/crawler.go:160	failed to perform repo crawl of "did:plc:lzkaig7u5entvhqgr4cgabwm": importing fetched repo (curHead: bafyreia7y77ta5i3sazxkkpfmjl5j2adznebxjbps4odgmryrl2v3qffji): process new repo (current head: bafyreia7y77ta5i3sazxkkpfmjl5j2adznebxjbps4odgmryrl2v3qffji): walkTree: can only handle dag-cbor objects in repos (bafkreif6df2wrudglcyzdcx5qcjvv6e6lmipfzghvdzxscl5klgbu4lu6y is 85):
```

I think that is the indexer code breaking on the new CID links in records, which in the case of blobs are "raw" CID pointers to file bytes (not dag-cbor multicodec). The only other current use of the `cid-link` datatype is in repo subscription objects, but in the future there might be more CIDs in records pointing to either dag-cbor (in local or remote repos), or raw CIDs (pointing to arbitrary content).

Another error is:

```
2023-03-29T21:50:03.697-0700	ERROR	bgs	bgs/fedmgr.go:168	failed handling event from "localhost:2583" (252): handle user event failed: signature check failed: invalid signature
```

I'm not sure what the deal is with that. We have an existing test in `testing/sig_test.go` for basic signature validation.

New users don't seem to get picked up (or at least logged) by BGS; labelmaker event consumers sees them and ignores.

bnewbold:
Ah, as a hint, also seeing the signature failure against typescript when using 'main' branch on both.

whyrusleeping:
Got the embed/union issue figured out. That test still isnt passing yet but its better.

PR #105, opened 2023-04-03
fakermaker integration tests, both internal and against typescript (running locally in dev mode)

bnewbold:
Refactored fakermaker to move internal methods in to `fakedata/` package.

Then use that to write integration tests.

Some code duplication here, though note that it can be helpful to comment out parts of the complete fakermaker run when pieces are not supported.

The `const` variable under `./testing` will probably need to change.

### DIFF
--- a/fakedata/accounts.go
+++ b/fakedata/accounts.go
@@ -51,10 +51,13 @@ func ReadAccountCatalog(path string) (*AccountCatalog, error) {
 		if err := decoder.Decode(&usr); err != nil {
 			return nil, fmt.Errorf("parse AccountContext: %w", err)
 		}
-		if usr.AccountType == "celebrity" {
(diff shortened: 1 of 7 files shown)


fakedata/accounts.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	defer catFile.Close()
+
+	decoder := json.NewDecoder(catFile)
+	for decoder.More() {
+		var usr AccountContext
+		if err := decoder.Decode(&usr); err != nil {
+			return nil, fmt.Errorf("parse AccountContext: %w", err)
+		}
+		if usr.AccountType == "celebrity" {
```
whyrusleeping, line 54: nit: use a switch, likely we will want to model other types soon

PR #106, opened 2023-04-03
handle updates and pds migrations

whyrusleeping:
basic handling for handle updates and pds migrations.
Theres likely more cases to be handled (migrateTo can be nil, what does that imply?) but this should be an improvement for now.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -322,7 +322,6 @@ func (bgs *BGS) lookupUserByDid(ctx context.Context, did string) (*User, error) 
 }
 
 func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *events.XRPCStreamEvent) error {
-	fmt.Println("HANDLE BGS FED EVENT", env.RepoHandle != nil, env.RepoCommit != nil)
 	ctx, span := otel.Tracer("bgs").Start(ctx, "handleFedEvent")
 	defer span.End()
 
@@ -377,7 +376,6 @@ func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *event
 
 		return nil
 	case env.RepoHandle != nil:
-		fmt.Println("handling handle update in bgs!!!!!")
 
 		// TODO: ignoring the data in the message and just going out to the DID doc
 		if _, err := bgs.createExternalUser(ctx, env.RepoHandle.Did); err != nil {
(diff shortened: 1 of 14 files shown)


whyrusleeping:
The changes I had to make in fedmgr.go feel weird. This smells like it should just be some more basic passthrough, but theres a lot of structure there now thats doing nothing.

bgs/bgs.go
```diff
@@ -372,6 +372,19 @@ func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *event
 			}
 		}
 
+		return nil
+	case env.RepoHandle != nil:
```
bnewbold, line 376: If we handle these the same for now, it could be a single compound `env.RepoHandle != nil || env.RepoMigrate != nil`.

I think you should at least check and log whether the account changed due to the event (eg, log "handle was updated due to event" or "got event, but DID doc didn't indicate handle update", along with the expected and found handles).

bnewbold (commented):
Generally looks fine.

I'd really like to see integration tests as we go, eg before merging to main or deploying to staging. I think the BGS integration tests already have a lot of the framework in place for this one (eg, multiple PDS instances, etc).

bgs/bgs.go
whyrusleeping: yeah, we could, but it requires adding the whole set of conditionals for figuring out which struct to pull the DID from, which increases the code length anyways.

bnewbold: Ah, yeah, sorry for conflicting requests. I think the logging is more important.

pds/server.go
```diff
@@ -356,6 +358,15 @@ func (s *Server) RunAPI(listen string) error {
 
 	e.HTTPErrorHandler = func(err error, ctx echo.Context) {
 		fmt.Printf("HANDLER ERROR: (%s) %s\n", ctx.Path(), err)
+
+		// TODO: need to properly figure out where http error codes for error
+		// types get decided. This spot is reasonable, but maybe a bit weird.
+		// reviewers, please advise
```
bnewbold, line 364: I have been doing error codes in XRPC endpoint handlers like: https://github.com/bluesky-social/indigo/blob/main/labeler/xrpc_handlers.go#L317

We can catch and handle all possible errors that way, which actually feels alright in the golang err handling paradigm. I think adding error types for every possible error here would be laborious, and having it work more than one way would also be brittle.

I would catch here things that are really 5xx errors and give them better messages. Eg, database connection errors, or any upstream HTTP connection errors (for proxy requests or other instances we might need to do upstream requests, like PLC).

bnewbold (approved):
LGTM, thanks for the test. Left a comment about HTTP error codes

pds/server.go
whyrusleeping: I'm gonna do this better in a later refactor, this particular error gets returned a lot of places, and should very often be a 404 (but not always). Probably need some helper functions or something

PR #107, opened 2023-04-03
lex: another interop test (unicode, blobs, etc)

bnewbold:
This matches up with the test vectors in: https://github.com/bluesky-social/atproto/pull/734

There was a single previous test file that i've renamed and split into interop and generic test cases. I kept the old interop test case versions (which didn't end up merged in `atproto`) around in a separate file.

Note that for golang I declared actual struct types (schemas) for these types, instead of doing generic parsing.

The "array" case (with arrays-of-arrays-of-bytes) does not work with golang cbor-gen currently, so I only have JSON tests for that.

I didn't do the "poorly formatted" test case yet, I wanted to get the basic tests in place quickly for now. It might be helpful to add a blob in that test case as well, to test generic (schema-less / struct-less) blob CID extraction.

### DIFF
--- a/lex/util/cbor_gen_test.go
+++ b/lex/util/cbor_gen_test.go
@@ -25,6 +25,815 @@ func (t *basicSchema) MarshalCBOR(w io.Writer) error {
 	}
 
 	cw := cbg.NewCborWriter(w)
+	fieldCount := 8
+
+	if t.Absent == nil {
+		fieldCount--
+	}
+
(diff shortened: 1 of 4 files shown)


whyrusleeping (approved):
This looks good, thanks!
Which types arent cborable? I'll patch that upstream

bnewbold:
@whyrusleeping the non-CBOR-able parts are `ipldNestedSchemaInnerInner`: https://github.com/bluesky-social/indigo/blob/main/lex/util/lex_interop_test.go#L45

existing test code which is commented out: https://github.com/bluesky-social/indigo/blob/main/lex/util/lex_interop_test.go#L308

PR #110, opened 2023-04-04
Improve cbor type inclusion detection during codegen

whyrusleeping:
Previously we werent detecting that the richtext facet features union needed to have the special union cbor type encoding set on it. This was because the 'walk' function we use to recursively set this field didnt cross schema boundaries, and even if we did that shallowly, we could have already processed the referenced schema and not actually passed the flag down recursively. So now as a hack i'm just calling the walk-set function twice. This is obviously bad, and I think we should just rewrite the codegen code instead of trying to improve it, it was written before several large lexicon refactors and its super messy as a result.


testing/feedpost_test.go
```diff
@@ -91,6 +91,7 @@ func TestFeedPostParse(t *testing.T) {
 				"$type": "app.bsky.embed.images",
 				"images": [
 					{
+						"$type": "",
```
bnewbold, line 94: Huh. This doesn't look right to me, it should either have a value or be omitted.

I wouldn't think that `EmbedImages_Image` ever needs `$type`, because it is not a top-level record type, and is not included in an enum itself. `EmbedImages` is in an enum, so it does need `$type` in that context.

bnewbold (changes requested):
I'm suspicious of the empty `$type` in the existing test JSON.

I think this PR needs to include a regression test for the specific CBOR bytes that failed (there are some existing tests that can be copied that do CBOR/JSON cross-validation). It also seems like we should ensure in this PR that we are hitting any remaining FeedPost corner cases by generating test cases that exercise all the possible enum variants.

Those test cases will also help later if we do end up rewriting lexgen.

testing/feedpost_test.go
whyrusleeping: Yeah, agreed. Will investigate (by the same token i think the empty alt field seems wrong to me)

whyrusleeping:
Empty `$type` has been remedied, also resulted in some cleanups on a bunch of other types. It's still not perfect but its good enough to get search processing going again

cmd/lexgen/main.go
```diff
@@ -131,6 +131,7 @@ func main() {
 		} else {
 			defmap := lex.BuildExtDefMap(schemas, []string{"com.atproto", "app.bsky"})
 
+			lex.FixRecordReferences(schemas, defmap, prefix)
```
bnewbold, line 134: A comment here would be nice, because it looks like a typo but is intentional

gen/main.go
```diff
@@ -30,7 +30,7 @@ func main() {
 		bsky.EmbedExternal{}, bsky.EmbedExternal_External{},
 		bsky.EmbedImages_Image{}, bsky.GraphFollow{}, bsky.ActorProfile{},
 		bsky.EmbedRecord{}, bsky.FeedLike{}, bsky.RichtextFacet{},
-		bsky.RichtextFacet_ByteSlice{}, bsky.RichtextFacet_Features_Elem{},
```
bnewbold, line 33: Interesting; we didn't need to serialize that type as part of post records?

bnewbold (approved):
I suspect we are going to need to come back to this, but looks good for now, especially to get search unblocked.

@dholms it would be nice for confidence to have interop tests similar to https://github.com/bluesky-social/atproto/pull/734, which exercise all the corner cases of feed records; they have gotten pretty complex. Of the core record types, i'm not nearly as concerned about likes, follows, profiles, etc. I could lead on this, but not until after team week.

gen/main.go
whyrusleeping: We do, but since they are unions they have to be codegenned by the lexgen command instead

PR #115, opened 2023-04-06
gosky: command to output did:key for an input key file

bnewbold:
Small hack that was needed for manually registering a did:plc with a recovery key.


whyrusleeping (approved):
LGTM

PR #117, opened 2023-04-09
vote dir does not work at least

mattn:


whyrusleeping:
I think this has since been removed in a broader fix of this command i did in #118

PR #119, opened 2023-04-11
add helper command for generating invite codes

whyrusleeping:
I'm not using the lexgen stuff because the auth is different for this route. Not quite sure how to properly deal with that yet so i'm yoloing a one-off http request

### DIFF
--- a/api/atproto/admindefs.go
+++ b/api/atproto/admindefs.go
@@ -20,9 +20,11 @@ type AdminDefs_ActionReversal struct {
 
 type AdminDefs_ActionView struct {
 	Action            *string                       `json:"action" cborgen:"action"`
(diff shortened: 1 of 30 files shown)


bnewbold:
There should be support in the XRPC client for admin passwords and use on that endpoint specifically:

- https://github.com/bluesky-social/indigo/blob/main/xrpc/xrpc.go#L21
- https://github.com/bluesky-social/indigo/blob/main/xrpc/xrpc.go#L101

... but need to update `com.atproto.account.createInviteCode` to `com.atproto.server.createInviteCode` on that second line.

existing config example:

```
		&cli.StringFlag{
			Name:     "admin-password",
			Usage:    "admin authentication password for PDS",
			Required: true,
			EnvVars:  []string{"ATP_AUTH_ADMIN_PASSWORD"},
		},
```

PR #123, opened 2023-04-14
gitignore default creds

jbn:
Maybe other people are more diligent than me but I almost committed them with my first code exploration.


bnewbold:
Thanks!

PR #124, opened 2023-04-16
Improve handling of bad records

whyrusleeping:
And some other misc improvements.

### DIFF
--- a/events/repostream.go
+++ b/events/repostream.go
@@ -36,20 +36,23 @@ func ConsumeRepoStreamLite(ctx context.Context, con *websocket.Conn, cb LiteStre
 					if err != nil {
 						e := fmt.Errorf("getting record %s (%s) within seq %d for %s: %w", op.Path, *op.Cid, evt.Seq, evt.Repo, err)
 						log.Error(e)
-						return nil
+						continue
 					}
 
 					if lexutil.LexLink(rc) != *op.Cid {
+						// TODO: do we even error here?
 						return fmt.Errorf("mismatch in record and op cid: %s != %s", rc, *op.Cid)
 					}
 
 					if err := cb(ek, evt.Seq, op.Path, evt.Repo, &rc, rec); err != nil {
-						return err
+						log.Errorf("event consumer callback (%s): %s", ek, err)
+						continue
 					}
 
 				case repomgr.EvtKindDeleteRecord:
 					if err := cb(ek, evt.Seq, op.Path, evt.Repo, nil, nil); err != nil {
-						return err
+						log.Errorf("event consumer callback (%s): %s", ek, err)
+						continue
 					}
 				}
 			}
--- a/indexer/crawler.go
+++ b/indexer/crawler.go
@@ -2,6 +2,7 @@ package indexer
 
 import (
 	"context"
+	"fmt"
 
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
 	"github.com/bluesky-social/indigo/models"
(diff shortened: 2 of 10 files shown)


repomgr/repomgr.go
```diff
@@ -885,28 +885,30 @@ func processOp(ctx context.Context, bs blockstore.Blockstore, op *mst.DiffOp) (*
...
+		outop := &RepoOp{
 			Kind:       kind,
 			Collection: parts[0],
 			Rkey:       parts[1],
 			RecCid:     &op.NewCid,
-			Record:     rec,
-		}, nil
+		}
+
+		rec, err := lexutil.CborDecodeValue(blk.RawData())
```
whyrusleeping, line 900: I guess we could move the record decoding up outside of the repomgr, and just include the bytes in the RepoOp from here

events/repostream.go
```diff
@@ -36,20 +36,23 @@ func ConsumeRepoStreamLite(ctx context.Context, con *websocket.Conn, cb LiteStre
 					if err != nil {
 						e := fmt.Errorf("getting record %s (%s) within seq %d for %s: %w", op.Path, *op.Cid, evt.Seq, evt.Repo, err)
 						log.Error(e)
-						return nil
+						continue
 					}
 
 					if lexutil.LexLink(rc) != *op.Cid {
+						// TODO: do we even error here?
```
bnewbold, line 43: I would say we should error out here. If the ops metadata is inconsistent with repo slice, that is pretty bad and we should error out and not partially process.

Maybe even mark repo as "dirty"? Something to think about this week.

indexer/crawler.go
```diff
@@ -20,22 +21,29 @@ type CrawlDispatcher struct {
...
+		concurrency: concurrency,
+	}, nil
 }
 
 func (c *CrawlDispatcher) Run() {
 	go c.mainLoop()
 
-	for i := 0; i < 3; i++ {
+	for i := 0; i < c.concurrency; i++ {
 		go c.fetchWorker()
```
bnewbold, line 47: Will probably want to make this runtime configurable (not just hardcode 10 as below), but we can do that later.

A TODO for me this week will be reviewing shutdown behavior. Not change in behavior here though.

repomgr/repomgr.go
```diff
@@ -885,28 +885,30 @@ func processOp(ctx context.Context, bs blockstore.Blockstore, op *mst.DiffOp) (*
 			return nil, err
 		}
 
-		rec, err := lexutil.CborDecodeValue(blk.RawData())
```
bnewbold, line 888: Semantically, I think we should (must?) validate that the record as acceptable here, but it should be lexicon-agnostic. We probably don't have a helper for that yet? What I mean is things like:

- actually decodes as CBOR at all
- meets additional DAG-CBOR requirements (eg, no tag types other than 42 "link")
- meets any Lexicon requirements, eg no weird CID types (raw-inline data CID)

Basically BGS needs to be conservative about low-level details (below app schema) and permissive about higher-level stuff (above app pschema).

In this PR, would be good to at least do CBOR decode, if we have a generic way to do that. Maybe call `CborTypeExtract`? all records should (must) have a `$type`.

bnewbold: Oh, now I see that you moved the `CborDecodeValue` below

repomgr/repomgr.go
bnewbold: That sounds roughly correct. Let's have a quick discussion later today about what `repomgr` is responsible for.

bnewbold (approved):
Mostly left forward-looking comments, which we can follow-up on later this week.

Overall LGTM to deploy in current setup (for search)

events/repostream.go
whyrusleeping: yeah, youre probably right

PR #125, opened 2023-04-18
dont try to stringify a nil thing

whyrusleeping:


bnewbold:
for the record, I did review and approve this, in-person

PR #127, opened 2023-04-20
some more work on debugging issues in event streams

whyrusleeping:


api/atproto/syncsubscribeRepos.go
```diff
@@ -11,7 +11,7 @@ func init() {
 
 type SyncSubscribeRepos_Commit struct {
 	Blobs  []util.LexLink               `json:"blobs" cborgen:"blobs"`
-	Blocks util.LexBytes                `json:"blocks" cborgen:"blocks"`
+	Blocks util.LexBytes                `json:"blocks,omitempty" cborgen:"blocks"`
```
bnewbold, line 14: this is a manual hack, and would get clobbered by running `lexgen`, I assume? Ok for now, making lexgen "clean" is a goal but will be a separate pass of work.

cmd/gosky/debug.go
```diff
@@ -231,3 +242,152 @@ var debugStreamCmd = &cli.Command{
 		return nil
 	},
 }
+
+// TODO: WIP - turns out to be more complicated than i initially thought
```
bnewbold, line 246: feels like this is almost a separate tool, though don't want to proliferate those too much. maybe do move code for this specific command to a separate file at least? it is pretty big

bnewbold (approved):
looks like the second commit is mostly just detecting the fork condition, but not doing anything yet, right? seems ok. we are really starting to pile up TODOs to handle in the next ~week.

api/atproto/syncsubscribeRepos.go
whyrusleeping: Yeah, lexgen needs work. I guess i can add a TODO comment here to help us not miss it in future diffs

PR #128, opened 2023-04-21
fix loop closure issue

whyrusleeping:
I thought that our linter would have caught this, but apparently it only catches certain shapes of this bug.
Probably worth auditing the rest of the codebase for this problem at some point (or just wait for the go team to fix it like jake said they might)

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -152,6 +152,16 @@ func (bgs *BGS) StartDebug(listen string) error {
 	})
 	http.Handle("/prometheus", prometheusHandler())
 
+	http.HandleFunc("/feddbg/conns", func(w http.ResponseWriter, r *http.Request) {
+		b, err := json.Marshal(bgs.slurper.GetActiveList())
+		if err != nil {
+			http.Error(w, err.Error(), 500)
+			return
+		}
+
+		w.Write(b)
+	})
+
 	return http.ListenAndServe(listen, nil)
 }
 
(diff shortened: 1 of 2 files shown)


bgs/fedmgr.go
```diff
@@ -92,6 +92,8 @@ func (s *Slurper) RestartAll() error {
 	}
 
 	for _, pds := range all {
+		pds := pds
+		s.active[pds.Host] = &pds
```
whyrusleeping, line 96: This line here was fixing another bug where we werent tracking active subscriptions started at bootup

bgs/bgs.go
```diff
@@ -152,6 +152,16 @@ func (bgs *BGS) StartDebug(listen string) error {
 	})
 	http.Handle("/prometheus", prometheusHandler())
 
+	http.HandleFunc("/feddbg/conns", func(w http.ResponseWriter, r *http.Request) {
```
bnewbold, line 155: What this endpoint does is self explanatory, but the path name and plan going forward are not. It's a new public web HTTP endpoint on a critical service, and I can't tell why it needs to be there.

A clearer path might be `/debug/upstream-conns`.

bnewbold (changes requested):
It is kind of confusing and takes a while to review these when they roll up multiple fixes and addition in a single commit. The loop fix would have been a fast/trivial single-line PR to review and ship quickly on it's own.

whyrusleeping:
The loop fix was discovered by adding a way to query what services were actually being slurped

bgs/bgs.go
whyrusleeping:
```suggestion
	http.HandleFunc("/debug/upstream-conns", func(w http.ResponseWriter, r *http.Request) {
```

whyrusleeping:
And the debug endpoint should not be public, @Jacob2161 can we make sure that one is only available on the servers localhost?

PR #135, opened 2023-04-27
lexgen failure case for richtext CBOR `$type` field missing `#link` suffix

bnewbold:
This is a small test case which fails with regular lexgen output.

I manually tweaked the lexgen output to show what needs to change, and ran cborgen, and the test passes after. The fix needed here is to get lexgen to output the correct `const=` values by default.

Handing this PR off to @whyrusleeping for now. I took a 3-minute attempted at the lexgen change and it didn't work, requires deeper understanding of lexgen internals.


bnewbold:
I re-based this and the problem seems to have been fixed at some point in the past... 6 months (!). would be good to merge this as a regression test.

PR #136, opened 2023-04-27
WIP: websocket connection improvements

bnewbold:
I stalled out on this SSL, `wss://`, `https://` cleanup attempt. Wanted to standardize on storing URLs for each crawl target, with http/https being stored and ws/wss getting inferred. But was more complex than expected.

Posting in case we want to harvest some of the changes out, or to come back to later.


bnewbold:
Something like this would probably still be helpful, but this is quite stale so closing for now.

PR #138, opened 2023-04-27
lex: regenerate APIs, moving manual hacks into code generator

bradfitz:
This way we can run "make lexgen" and not have to manually tweak the output.


bradfitz:
cc @whyrusleeping

whyrusleeping (approved):
LGTM, thanks!

PR #139, opened 2023-04-27
all: rename "test_files" dirs to Go idiomatic "testdata".

bradfitz:
The Go "testdata" convention (as respected by the cmd/go tool) is briefly documented in https://pkg.go.dev/cmd/go


bradfitz:
cc @whyrusleeping

PR #140, opened 2023-04-27
lex: add some minimal docs to the generated code

bradfitz:
Not perfect, but a step.


lex/gen.go
```diff
@@ -88,7 +88,6 @@ func (s *Schema) Name() string {
 
 type outputType struct {
 	Name      string
-	DefName   string
```
bradfitz, line 91: (this was never used)

bradfitz:
cc @bnewbold @whyrusleeping

bnewbold (approved):
LGTM, thanks!

Presumably we will end up with package-level doc files with additional context and links to "canonical" lexicon JSON schema files (location TBD).

PR #143, opened 2023-04-29
mst: add a node kind type, tweak some docs

bradfitz:


bradfitz:
cc @whyrusleeping 

I'm just reading code at the moment, cleaning up little things along the way.

mst/mst.go
```diff
@@ -39,7 +40,6 @@
 // If the first leaf in a tree is `bsky/posts/abcdefg` and the second is `bsky/posts/abcdehi`
 // Then the first will be described as `prefix: 0, key: 'bsky/posts/abcdefg'`,
 // and the second will be described as `prefix: 16, key: 'hi'.`
-
```
bradfitz, line 42: (So the comment above is the Go package doc... needs to be attached to the package line)

whyrusleeping (approved):
LGTM, thanks!

mst/mst.go
Abdullahitajuddeen: hi please i need your help my phone number is +2348069056081

PR #144, opened 2023-04-29
mst: remove an alloc in isValidMstKey

bradfitz:
goos: darwin
    goarch: arm64
    pkg: github.com/bluesky-social/indigo/mst
    IsValidMstKey-8   301.1n ± 4%   262.8n ± 4%  -12.69% (p=0.000 n=10)

                    │   before   │               after                │
                    │    B/op    │   B/op     vs base                 │
    IsValidMstKey-8   32.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                    │   before   │                after                │
                    │ allocs/op  │ allocs/op   vs base                 │
    IsValidMstKey-8   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)

### DIFF
--- a/mst/mst_test.go
+++ b/mst/mst_test.go
@@ -502,3 +502,12 @@ func diffOpEq(aa, bb *DiffOp) bool {
 	}
 	return true
 }
+
+func BenchmarkIsValidMstKey(b *testing.B) {
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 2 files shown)


bradfitz:
cc @bnewbold @whyrusleeping

whyrusleeping (approved):
Hot

PR #145, opened 2023-04-30
Join string slices in makeParams to fix broken urlencodes

ericvolp12:
Currently the `makeParams` function will take a string slice and URL encode it as `[hello+world]` instead of `hello,world` which makes the XRPC API upset (HTTP:400 errors) since it wants a comma separated list of strings. This change tries to cast the params interface into a string slice, and if it matches, it does a join instead of just printing the debug value string into the query params.

Tested in the [GraphBuilder](https://github.com/ericvolp12/bsky-experiments/blob/main/pkg/events/helpers.go#L20) and hasn't broken anything, not sure if/what the testing plan is here.

That function previously threw 400s trying to talk to the XRPC API.

That project is using a [module replace](https://github.com/ericvolp12/bsky-experiments/blob/main/go.mod#L107) to test my commit in place of the upstream indigo package.

### DIFF
--- a/xrpc/xrpc.go
+++ b/xrpc/xrpc.go
@@ -45,16 +45,16 @@ const (
 )
 
 func makeParams(p map[string]interface{}) string {
-	var parts []string
+	params := url.Values{}
 	for k, v := range p {
 		if s, ok := v.([]string); ok {
-			parts = append(parts, fmt.Sprintf("%s=%s", k, url.QueryEscape(strings.Join(s, ","))))
+			params.Add(k, strings.Join(s, ","))
 		} else {
-			parts = append(parts, fmt.Sprintf("%s=%s", k, url.QueryEscape(fmt.Sprint(v))))
+			params.Add(k, fmt.Sprint(v))
 		}
 	}
 
-	return strings.Join(parts, "&")
+	return params.Encode()
 }
 
 func (c *Client) Do(ctx context.Context, kind XRPCRequestType, inpenc string, method string, params map[string]interface{}, bodyobj interface{}, out interface{}) error {
--- a/xrpc/xrpc.go
+++ b/xrpc/xrpc.go
@@ -44,7 +44,9 @@ const (
 	Procedure
 )
 
-func makeParams(p map[string]interface{}) string {
+// makeParams converts a map of string keys and any values into a URL-encoded string.
+// If a value is a slice of strings, it will be joined with commas.
+func makeParams(p map[string]any) string {
 	params := url.Values{}
 	for k, v := range p {
 		if s, ok := v.([]string); ok {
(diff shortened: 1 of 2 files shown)


ericvolp12:
I added some unit tests for the `makeParams` functions, just covering a few cases though.

xrpc/xrpc.go
```diff
@@ -45,12 +45,16 @@ const (
 )
 
 func makeParams(p map[string]interface{}) string {
-	var parts []string
+	params := url.Values{}
 	for k, v := range p {
-		parts = append(parts, fmt.Sprintf("%s=%s", k, url.QueryEscape(fmt.Sprint(v))))
+		if s, ok := v.([]string); ok {
+			params.Add(k, strings.Join(s, ","))
```
bradfitz, line 51: so we want `p=x,y,z` instead of `p=x&p=y&p=z`? Can you at least document that on `func makeParams`, about what the values can be and what it does, while you're here?

Also, might as well change `interface{}` to `any` since this module requires Go 1.20+ anyway.

xrpc/xrpc_test.go
```diff
@@ -0,0 +1,64 @@
...
+				t.Errorf("Expected '%s', got '%s'", tc.expected, result)
+			}
+		})
+	}
+}
+
+func TestXRPC(t *testing.T) {
+	// Run the tests
+	TestMakeParams(t)
+}
```
bradfitz, line 64: delete this. Go does that automatically. This would run the tests twice.

xrpc/xrpc_test.go
```diff
@@ -0,0 +1,64 @@
...
+			},
+			expected: "key1=value1&key2=value2%2Cvalue3",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			result := makeParams(tc.input)
+			if result != tc.expected {
+				t.Errorf("Expected '%s', got '%s'", tc.expected, result)
```
bradfitz, line 55: use %q instead of '%s'

also, Go convention is got first, then want.  "got %q; want %q"

xrpc/xrpc.go
ericvolp12: The case I was getting at is we want `p=x,y,z` instead of `p=[x+y+z]` which is how it currently handles slices of strings. I'll try converting this to use `any`.

ericvolp12:
ah my backticks got swallowed by git in the commit message, my bad

xrpc/xrpc_test.go
ericvolp12: done, thanks

xrpc/xrpc_test.go
ericvolp12: done, thanks

xrpc/xrpc.go
ericvolp12: Added a comment clarifying behavior.

bradfitz (approved):
LGTM

bradfitz:
(But note that I'm not a maintainer here; leaving for somebody else.)

whyrusleeping (approved):
Nice, Thank you so much! This LGTM too

whyrusleeping:
Thanks @bradfitz !

PR #146, opened 2023-04-30
gosky,stress: add positional argument docs

raggi:


whyrusleeping (approved):
LGTM,

os.Exit in a helper is a bit magic for my tastes, but the alternative is much less clean, and I think this is likely fine (unless we decide we want to start running commands in unit tests for some reason, but at that point we will just fix it then)

raggi:
Oh turns out that's actually not so bad, the CLI lib has an error type for it, just took another pass.

whyrusleeping (approved):
Nice, looks great

PR #149, opened 2023-05-01
mst: remove alloc in leadingZerosOnHash

bradfitz:
goos: darwin
    goarch: arm64
    pkg: github.com/bluesky-social/indigo/mst
    LeadingZerosOnHash-8   72.28n ± 3%   58.76n ± 2%  -18.71% (p=0.000 n=10)

                         │   before   │               after                │
                         │    B/op    │   B/op     vs base                 │
    LeadingZerosOnHash-8   48.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                         │   before   │                after                │
                         │ allocs/op  │ allocs/op   vs base                 │
    LeadingZerosOnHash-8   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)

### DIFF
--- a/mst/mst.go
+++ b/mst/mst.go
@@ -198,7 +198,7 @@ func entriesFromNodeData(ctx context.Context, nd *NodeData, cst cbor.IpldStore) 
 	if len(nd.Entries) > 0 {
 		// NOTE(bnewbold): can compute the layer on the first KeySuffix, because for the first entry that field is a complete key
 		firstLeaf := nd.Entries[0]
-		layer = leadingZerosOnHash(string(firstLeaf.KeySuffix))
+		layer = leadingZerosOnHashBytes(firstLeaf.KeySuffix)
 	}
 
 	entries, err := deserializeNodeData(ctx, cst, nd, layer)
--- a/mst/mst_test.go
+++ b/mst/mst_test.go
@@ -511,3 +511,27 @@ func BenchmarkIsValidMstKey(b *testing.B) {
 		}
 	}
 }
+
+func TestLeadingZerosOnHashAllocs(t *testing.T) {
+	var sink int
+	const in = "some.key.prefix/key.bar123456789012334556"
+	var inb = []byte(in)
+	if n := int(testing.AllocsPerRun(1000, func() {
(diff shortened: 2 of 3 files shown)


bradfitz:
cc @bnewbold @whyrusleeping

whyrusleeping (approved):
LGTM, would like @bnewbold to give a quick look too

bradfitz:
BTW, I'm just being lazy and spamming both of you on all these changes as I haven't really learned (or done git archaeology) to learn who "owns" what. If there's some heuristics I should follow on assigning owners, let me know, otherwise I'll keep spamming both of you :)

bnewbold (approved):
Great! I can remember not finishing this part of the "MST paths as bytes not strings" refactor. Thanks for bringing a benchmark along for the ride. Definitely a hot path as we have larger and larger repos.

If you are poking at the MST code, it is probably obvious but this is basically a line-by-line port from our typescript implementation, and is really not idiomatic golang.

There is also a Rust implementation I wrote from scratch, which is clean and understandable, but is immutable data structure style (rebuilds the entire tree from scratch on mutation) and doesn't support partial MST trees (aka, repo "slices" or "diffs").

https://gitlab.com/bnewbold/adenosine/-/blob/main/adenosine/src/mst.rs

We do have some decent cross-language test vectors which make development and refactoring reasonable here. We are lacking some "nastier" test cases, like invalid/malformed CBOR; out-of-order MST node lists (invalid, should throw error); wrong CID types (only SHA256 should be allowed); etc.

bnewbold:
@bradfitz go ahead and keep spamming us both! roughly why has done more of the carstore/BGS/repomgr/gosky stuff and I did the last pass on MST/labeling/cmds, but it is pretty mixed at this point.

bradfitz:
> If you are poking at the MST code, it is probably obvious but this is basically a line-by-line port from our typescript implementation, and is really not idiomatic golang.

yeah, @whyrusleeping and @Jacob2161 both nudged me into looking at it. I read the paper and the code and am so far just doing some cleanups and benchmarks before I do anything too invasive.

PR #151, opened 2023-05-02
mst: optimize deserializeNodeData allocs, isValidMstKey CPU

bradfitz:
isValidMstKey used regexp, which isn't super fast in Go. Write it out by hand instead, making it much faster:

                    │    before    │                after                │
                    │    sec/op    │   sec/op     vs base                │
    IsValidMstKey-8   276.15n ± 1%   27.49n ± 4%  -90.05% (p=0.000 n=10)

Also, allocate less in deserializeNodeData, reusing a scratch buffer between entries and only doing the []byte to string conversion once.

### DIFF
--- a/mst/mst_util.go
+++ b/mst/mst_util.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
-	"regexp"
 	"strings"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


bradfitz:
cc @bnewbold @whyrusleeping

bradfitz:
@Jacob2161 sent me some pprof snapshots. That regexp was using like 3.5% of your CPU.

And GC is ~25% of CPU, which means y'all are allocating a bunch, which the heap pprof confirms, and showed a fair bit in deserializeNodeData (but mostly everywhere). The goal of this PR was really removing the regexp. The deserializeNodeData stuff happened by accident when I was looking at the callers.

bnewbold:
MST "key" (aka "path" in repo-level nomenclature) changes: sort of wistful for regex, but of course fine. Cross-language test coverage for this behavior exists at a higher-level, via `TestAllowedKeys`.

The `isValidMstKey` function does correctly reject empty key (zero-length string / zero-length bytes), though the way we wrote it doesn't call that out very explicitly. This PR isn't touching those lines so no change needed here.

deserializeNodeData changes: yay! LGTM

bradfitz:
> sort of wistful for regex, but of course fine

I feel ya. I spent a decade with Perl and love me some regexps. Go's implementation has gotten better over time but it's still not super fast. There are a bunch of fast paths it could have but doesn't.

bnewbold:
A win with regex is that we can just copy the string around between different language implementations.

A loss is that, actually, you can't always do that, there are a bunch of dialects and corner-cases. Even beyond semantics/"correctness", different langs/impl will have different complexity explosions and denial-of-service vectors. oh well!

bradfitz:
I suppose I could've at least moved the literal regexp into a test and fuzzed and/or test vector'ed that the custom implementation matches the regexp.

bradfitz:
I sent https://github.com/bluesky-social/indigo/pull/155 as a peace offering for wistful sadness.

PR #154, opened 2023-05-03
Add code for search service

whyrusleeping:
Dockerfile build doesnt quite work yet


bnewbold (approved):

bnewbold:
@Jacob2161 the AWS container upload failed. That action file is basically a copy of the `bigsky` one. Do we need to add/configure a directory in the AWS container service or something?

Jacob2161:
> @Jacob2161 the AWS container upload failed. That action file is basically a copy of the `bigsky` one. Do we need to add/configure a directory in the AWS container service or something?

Looks like it's just that the repo didn't exist. I created it and the job is re-running now.

PR #157, opened 2023-05-03
mst: add BenchmarkDiffTrees

bradfitz:
To establish a baseline for future DiffTrees optimizations.


bradfitz:
cc @bnewbold @whyrusleeping

bnewbold:
now we are cooking! nice to start getting some benchmarks in here

PR #158, opened 2023-05-03
mst: make countPrefixLen match JavaScript's UTF-16-edness

bradfitz:
And optimize it a bit, so the Go compiler can avoid some bounds checks.

benchstat:

    CountPrefixLen-8   78.85n ± 1%   38.45n ± 5%  -51.23% (p=0.000 n=10)

                     │   before   │             after              │
                     │    B/op    │    B/op     vs base            │
    CountPrefixLen-8   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
    ¹ all samples are equal

                     │   before   │             after              │
                     │ allocs/op  │ allocs/op   vs base            │
    CountPrefixLen-8   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
    ¹ all samples are equal

### DIFF
--- a/mst/mst_test.go
+++ b/mst/mst_test.go
@@ -599,3 +599,37 @@ func BenchmarkDiffTrees(b *testing.B) {
 		b.Fatal("diffs not equal")
 	}
 }
+
+var countPrefixLenTests = []struct {
+	a, b string
+	want int
+}{
+	{"", "", 0},
+	{"a", "", 0},
+	{"", "a", 0},
+	{"a", "b", 0},
+	{"a", "a", 1},
+	{"ab", "a", 1},
+	{"a", "ab", 1},
+	{"ab", "ab", 2},
+	{"abcdefghijklmnop", "abcdefghijklmnoq", 15},
+}
+
+func TestCountPrefixLen(t *testing.T) {
+	for _, tt := range countPrefixLenTests {
+		if got := countPrefixLen(tt.a, tt.b); got != tt.want {
+			t.Errorf("countPrefixLenTests(%q, %q) = %d, want %d", tt.a, tt.b, got, tt.want)
+		}
+	}
+}
+
+func BenchmarkCountPrefixLen(b *testing.B) {
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		for _, tt := range countPrefixLenTests {
+			if got := countPrefixLen(tt.a, tt.b); got != tt.want {
+				b.Fatalf("countPrefixLenTests(%q, %q) = %d, want %d", tt.a, tt.b, got, tt.want)
+			}
+		}
+	}
+}
--- a/mst/mst_util.go
+++ b/mst/mst_util.go
@@ -170,25 +170,18 @@ func serializeNodeData(entries []nodeEntry) (*nodeData, error) {
 	return &data, nil
 }
 
-func min(a, b int) int {
-	if a <= b {
-		return a
-	}
-	return b
-}
-
-// how many leading chars are identical between the two strings?
+// how many leading bytes are identical between the two strings?
 // Typescript: countPrefixLen(a: string, b: string) -> number
 func countPrefixLen(a, b string) int {
-	aa := []byte(a)
-	bb := []byte(b)
-	count := min(len(aa), len(bb))
-	for i := 0; i < count; i++ {
-		if aa[i] != bb[i] {
+	// This pattern avoids panicindex calls, as the Go compiler's prove pass can
+	// convince itself that neither a[i] nor b[i] are ever out of bounds.
+	var i int
+	for i = 0; i < len(a) && i < len(b); i++ {
+		if a[i] != b[i] {
 			return i
 		}
 	}
-	return count
+	return i
 }
 
 // both computes *and* persists a tree entry; this is different from typescript


bradfitz:
cc @bnewbold @whyrusleeping

bnewbold:
We should really be working with bytes everywhere internally in MST. The main motivation for this decision was specifically this prefix length counting, in the context of Unicode characters in MST paths. Javascript does UTF-16, while much of the rest of the world does UTF-8, so you get different counts and different MST structures for the same input.

We belt-and-suspenders'd it by:

* changing the MST node DAG-CBOR schema to be byte array not string, for the key suffixes
* ensuring we do this counting on bytes
* validating the MST paths (aka keys) to be a subset of ASCII

We might relax the constraint on MST paths in the future, so we should try to ensure we could work with hypothetical unicode internally, I think? Doing that informally but not explicitly testing, etc, might be a code smell.

The part that is a hold-out from the refactor is having strings pretty much anywhere in MST internals. I think even the interface the MST exposes to the world should probably be bytes not strings, and let the "repo" package handle the decoding/encoding.

bradfitz:
Good background, but FWIW this change doesn't change any semantics around unicode. It's purely a performance/cleanup change. 

Go `string` and `[]byte` both just store arbitrary bytes. A Go `string` doesn't mean unicode or utf-8 or anything in particular. The only difference is mutability and whether they're `comparable` (a string supports ==, being a map key, etc)

(https://go.dev/blog/strings says more, if this is new information.)

bradfitz:
@bnewbold, I've updated this PR to match JavaScript's behavior, while also being faster. Maybe this is more interesting to you now? 😄

bradfitz:
But warning: this still isn't safe for non-ASCII keys in MST yet. The callers will need to be updated. Notably, this only returns the matching `"k"` value for the prefix length that JavaScript encodes. But this doesn't build such non-ASCII trees correctly yet because you can't just slice a UTF-8 encoded string by this number to get to the right suffix. That'll require updating too. But currently all the MST stuff requires ASCII keys, so I think this is still safe meanwhile.

bnewbold:
Hey, sorry, I think I miscommunication about this. What we identified previously (back in March) was there there was ambiguity with strings w/r/t UTF-16 and UTF-8 counting. The resolution of this ambiguity was to switch from using "string" to "bytes" in the MST node objects, and to specify that strings should be encoded as UTF-8 at the interface boundary (aka, either at the MST package interface, or even better have the MST package interface take bytes, and the "repo" package take care of UTF-8 encoding/decoding).

So, we don't want to replicate UTF-16 encoding here. Instead, a while back we updated the *Typescript* code to explicitly encode to bytes as UTF-8, and that is what the network/wire/file spec is.

The somewhat confusing `TestPrefixLenWide` is a legacy from when I was reviewing this MST code and discovered the strings problem, and we kept that around to demonstrate the issue.

bradfitz:
So is the plan for the common prefix length integer to be defined as the common UTF-8 bytes or Unicode codepoints?

Do you worry about prefixes that don't have any JSON representation?

That is, is `commonPrefixLen("fooΓ", "fooΔ")` equal to 4 or 3?

Because if the former, e.g. `"Γ"` and `"Δ"` both encode to two bytes of UTF-8, both starting with the byte `0xCE` but that has no JSON representation as a single byte.

But maybe you're always working with DAG-CBOR or other binary formats and you don't need it to round-trip through a Unicode or UTF-8 string. It'd work in Go but I'm not sure how painful working with binary strings is in JS. Maybe you stay in ArrayBuffer or whatever world enough as it is?

bnewbold:
Common prefix length is defined as bytes, and strings get encoded as UTF-8, so: UTF-8 bytes.

For now, we effectively limit the repo paths to be a subset of ASCII, so we don't really even need an answer about encoding unicode characters. The string examples you give ("fooΓ", "fooΔ") would both be rejected as invalid repo paths today. But we do want to leave the door open to unicode in repo paths in the future, so we do have an answer.

The answer is that "paths as strings" is a repo-layer concept. When the repo contents are encoded in to an MST tree, the first step is to convert the repo paths in to an MST key, and that key is bytes. So the answer to the above would be "4 bytes common prefix".

The changes in the typescript implementation are described in this PR: https://github.com/bluesky-social/atproto/pull/635

Basically, we switched the MST node object from having "strings" to "byte arrays". I think they use [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) instead of native strings in that implementation.

A good analogy is something like BoltDB, a key/value store. The keys are bytes, and the function signature looks like:

```
func (b *Bucket) Get(key []byte) []byte
func (b *Bucket) Put(key []byte, value []byte) error
```

It is really common to use strings encoded in UTF-8 as keys, but calling code needs to handle that. Similar, atproto MST trees should have `[]byte` in all the function signatures. The calling code in other packages needs to deal with encoding/decoding of strings to UTF-8 bytes.

bradfitz:
Gotcha. But repeating what I said before: in Go, a `string` doesn't mean any particular encoding or imply Unicode or imply UTF-8. It's just a read-only view of an immutable `[]byte`. So writing `[]byte` in a signature might be a win for docs, but it's not required for correctness. Sometimes `string` (or a typed variant thereof) is better for other reasons due to it being comparable, immutable, etc.

bradfitz:
I've reverted this PR back to just the optimization it was previously (with no behavior changes).

bnewbold (approved):
Looks good! Thanks for bearing with the review/conversation.

PR #161, opened 2023-05-04
implement repo takedowns

whyrusleeping:
Adds a basic admin route to fully take down a given repo, including marking the user as 'taken down' and fully deleting all associated data.

### DIFF
--- a/bgs/admin.go
+++ b/bgs/admin.go
@@ -3,26 +3,14 @@ package bgs
 import (
 	"strconv"
 
-	"github.com/bluesky-social/indigo/util"
 	"github.com/labstack/echo/v4"
 )
-
-func (bgs *BGS) handleAdminDeleteRecord(e echo.Context) error {
-	puri, err := util.ParseAtUri(e.QueryParam("uri"))
-	if err != nil {
-		return err
-	}
-
-	_ = puri
-
-	panic("TODO")
-}
 
 func (bgs *BGS) handleAdminBlockRepoStream(e echo.Context) error {
 	panic("TODO")
 }
 
-func (bgs *BGS) handleAdminDisableNewSlurps(e echo.Context) error {
+func (bgs *BGS) handleAdminSetSubsEnabled(e echo.Context) error {
 	enabled, err := strconv.ParseBool(e.QueryParam("enabled"))
 	if err != nil {
 		return err
@@ -32,5 +20,8 @@ func (bgs *BGS) handleAdminDisableNewSlurps(e echo.Context) error {
 }
 
 func (bgs *BGS) handleAdminTakedownRepo(e echo.Context) error {
-	panic("TODO")
+	did := e.QueryParam("did")
+	ctx := e.Request().Context()
+
+	return bgs.TakedownRepo(ctx, did)
 }
--- a/bgs/admin.go
+++ b/bgs/admin.go
@@ -19,9 +19,16 @@ func (bgs *BGS) handleAdminSetSubsEnabled(e echo.Context) error {
 	return bgs.slurper.SetNewSubsDisabled(!enabled)
 }
 
-func (bgs *BGS) handleAdminTakedownRepo(e echo.Context) error {
+func (bgs *BGS) handleAdminTakeDownRepo(e echo.Context) error {
 	did := e.QueryParam("did")
 	ctx := e.Request().Context()
 
-	return bgs.TakedownRepo(ctx, did)
+	return bgs.TakeDownRepo(ctx, did)
+}
+
+func (bgs *BGS) handleAdminReverseTakedown(e echo.Context) error {
+	did := e.QueryParam("did")
+	ctx := e.Request().Context()
+
+	return bgs.ReverseTakedown(ctx, did)
 }
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -199,7 +199,8 @@ func (bgs *BGS) Start(listen string) error {
 
 	admin := e.Group("/admin", bgs.checkAdminAuth)
 	admin.POST("/subs/setEnabled", bgs.handleAdminSetSubsEnabled)
-	admin.POST("/repo/takedown", bgs.handleAdminTakedownRepo)
+	admin.POST("/repo/takeDown", bgs.handleAdminTakeDownRepo)
+	admin.POST("/repo/reverseTakedown", bgs.handleAdminReverseTakedown)
 
 	return e.Start(listen)
 }
@@ -282,7 +283,11 @@ type User struct {
 	Handle    string         `gorm:"uniqueIndex"`
 	Did       string         `gorm:"uniqueIndex"`
 	PDS       uint
-	Takedown  bool
+
+	// TakenDown is set to true if the user in question has been taken down.
+	// A user in this state will have all future events related to it dropped
+	// and no data about this user will be served.
+	TakenDown bool
 }
 
 type addTargetBody struct {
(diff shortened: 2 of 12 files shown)


bgs/bgs.go
```diff
@@ -661,3 +663,24 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
 
 	return subj, nil
 }
+
+func (bgs *BGS) TakedownRepo(ctx context.Context, did string) error {
```
bradfitz, line 667: "take down" (with a space) is a phrasal verb. "takedown" is the noun. 

So should this be "TakeDownRepo" instead? (and elsewhere)

whyrusleeping: you have reached the limits of my mastery of english. I *think* weve been calling them 'takedowns' elsewhere? cc @bnewbold

bgs/bgs.go
```diff
@@ -281,6 +282,7 @@ type User struct {
 	Handle    string         `gorm:"uniqueIndex"`
 	Did       string         `gorm:"uniqueIndex"`
 	PDS       uint
+	Takedown  bool
```
bradfitz, line 285: docs on this field? the others are pretty obvious. this one is less so.

bgs/bgs.go
bradfitz: You can call it "takedown" when it's a noun... like `(*BGS).GetRepoTakedowns()` returning nouns (hypothetical `Takedown` objects). But when you're telling something to do something (an imperative verb: "take down") then it'd be `(*BGS).TakeDownRepo`

"Clean up this room!" "Nice cleanup." etc

But maybe it's a losing battle when you're also working with lots of generated code anyway that's not going to be grammatical.

bnewbold: I agree with bradfitz that `TakeDownRepo()` is more correct. In the noun form would have to add a verb like `CreateRepoTakedown()`.

bgs/admin.go
```diff
@@ -3,26 +3,14 @@ package bgs
...
-
-	panic("TODO")
-}
-
 func (bgs *BGS) handleAdminBlockRepoStream(e echo.Context) error {
 	panic("TODO")
 }
 
-func (bgs *BGS) handleAdminDisableNewSlurps(e echo.Context) error {
+func (bgs *BGS) handleAdminSetSubsEnabled(e echo.Context) error {
```
bnewbold, line 13: LOL, :skull: slurps

bgs/bgs.go
bnewbold: alternatively, we often use the pattern of having it be an optional timestamp (`TakedownAt`; weird noun-like, not verb-like) as a bit of a papertrail.

bgs/bgs.go
```diff
@@ -661,3 +663,24 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
...
+		return err
+	}
+
+	if err := bgs.repoman.TakedownRepo(ctx, u.ID); err != nil {
+		return err
+	}
+
+	if err := bgs.events.TakedownRepo(ctx, u.ID); err != nil {
+		return err
+	}
```
bnewbold, line 683: at first I thought all of this should be in a transaction. but thinking about it more, I guess we probably want to update the user table pretty quick with the takedown state, and then proceed with the other steps, which take out locks. I think doing it in that order would prevent any obvious race conditions between the takedown API handler and any event stream processing?

And, admin/client can just send another takedown API request if the first does not complete, and that should be fine.

carstore/bs.go
```diff
@@ -497,14 +500,18 @@ func (cs *CarStore) openNewShardFile(ctx context.Context, user util.Uid, seq int
...
 	if err := os.WriteFile(fname, data, 0664); err != nil {
 		return "", err
 	}
 
 	return fname, nil
 }
 
+func (cs *CarStore) deleteShardFile(ctx context.Context, sh *CarShard) error {
+	return os.Remove(fnameForShard(sh.Usr, sh.Seq))
+}
```
bnewbold, line 513: ooooo golly deleting from disk makes me nervous, instead of just moving to a "limbo" directory.

Remember this is just repo content, not, eg, images, so very unlikely to contain "hard illegal" content. I guess stuffing in `$bytes` is not unimaginable though.

xrpc/xrpc.go
```diff
@@ -121,6 +121,7 @@ func (c *Client) Do(ctx context.Context, kind XRPCRequestType, inpenc string, me
 	if resp.StatusCode != 200 {
 		var i interface{}
 		_ = json.NewDecoder(resp.Body).Decode(&i)
+		//fmt.Println(i)
```
bnewbold, line 124: accidental commit?

events/dbpersist.go
```diff
@@ -255,3 +255,21 @@ func (p *DbPersistence) readCarSlice(ctx context.Context, rer *RepoEventRecord)
...
+			return fmt.Errorf("failed to find repo event record: %w", err)
+		}
+
+		if err := p.db.Where("repo_event_record_id = ?", event.Seq).Delete(&RepoOpRecord{}).Error; err != nil {
+			return fmt.Errorf("failed to delete repo op records: %w", err)
+		}
+
+		if err := p.db.Delete(&RepoEventRecord{}, "seq = ?", event.Seq).Error; err != nil {
+			return err
+		}
```
bnewbold, line 272: I'm a little nervous about a database freeze or deadlock coming out of this. We just had an issue with account deletions causing 10+ minute hangs in PDS prod.

One pattern I have seen to avoid that in the past for this kind of thing is to delete up to, eg, 200 rows per transaction in a loop, starting the next scan from the max seq of the last-deleted batch.

I guess we could put this off until after sandbox mode, and repo takedowns are going to be rare, but it feels like leaving a known footgun lying around. Prod BGS is pretty big from the get-go.

bgs/bgs.go
```diff
@@ -198,7 +198,8 @@ func (bgs *BGS) Start(listen string) error {
 	e.GET("/xrpc/_health", bgs.HandleHealthCheck)
 
 	admin := e.Group("/admin", bgs.checkAdminAuth)
-	admin.POST("/deleteRecord", bgs.handleAdminDeleteRecord)
+	admin.POST("/subs/setEnabled", bgs.handleAdminSetSubsEnabled)
+	admin.POST("/repo/takedown", bgs.handleAdminTakedownRepo)
```
bnewbold, line 202: I think we need "un-takedown", even from the get-go. The lesson from PDS is that even if we only think we will use this in extreme cases, if it is a lever then in an emergency we might use it, and need a way to undo it.

Fortunately, that handler should be quite simple: just flip the boolean in the database table.

bnewbold (approved):
Left a few comments.

Sorry for the slow review, and know we are trying to ship this fast to get to sandbox mode, and can polish more for prod later. I think un-takedown should be pretty easy.

The deadlock concern we could just file a ticket for now.

The last issue is no test coverage. I think this is a pretty critical behavior and we should have at least some automated testing of it.

bgs/bgs.go
whyrusleeping: yeah, the actions are idempotent, so can just keep rerunning it until it sticks if theres any errors.

carstore/bs.go
whyrusleeping: Theres very few reasons we will need to actually hard takedown an account at this level, I think a hard delete of files here is fine, especially since its very easy to re-pull everything back by just un-takedowning the user and triggering an index

events/dbpersist.go
whyrusleeping: Yeah, lets make an issue for this. I think that deleting on a simple condition that is easily looked up by an index should be fine, especially for an admin action that happens *very* rarely.

bgs/bgs.go
whyrusleeping: I dont know about that. We shouldnt ever do a takedown at the bgs level lightly. If we make that decision, we should have zero plans to un-make it.

whyrusleeping: added the route anyways

bnewbold: thanks!

bnewbold (approved):
Looks good! Appreciate you working through all the feedback

PR #164, opened 2023-05-09
some improvements to report listing

whyrusleeping:
This panics at some point due to trying to marshal an empty enum. Need to debug that


cmd/gosky/admin.go
```diff
@@ -361,33 +361,41 @@ var listReportsCmd = &cli.Command{
...
 			}
-			fmt.Println(string(b))
-			for _, act := range rep.ResolvedByActionIds {
-				action, err := comatproto.AdminGetModerationAction(ctx, xrpcc, act)
+
+			tojson := func(i any) string {
+				b, err := json.MarshalIndent(i, "", "  ")
 				if err != nil {
-					return err
+					panic(err)
```
notque, line 382: if err != nil {
	fmt.Printf("Failed to marshal value: %v. Error: %v\n", i, err)

instead of panic to find the error?

PR #167, opened 2023-05-10
go.mod: bump github.com/ipld/go-car to latest

bradfitz:
Updates github.com/ipld/go-car to the latest commit to bring in https://github.com/ipld/go-car/pull/420 which should drop allocated bytes (garbage that triggers the GC to spend CPU) by about 9%.


bradfitz:
cc @bnewbold @whyrusleeping

bnewbold (approved):

PR #171, opened 2023-05-11
chore: cleanup old github.com/ipfs/go-car import

Jorropo:
github.com/ipfs/go-car is an archived repo that have been moved to github.com/ipld/go-car.


whyrusleeping:
old habits die hard

whyrusleeping:
Thanks!

PR #173, opened 2023-05-15
add tooling for invite control

whyrusleeping:


whyrusleeping:
hrm... some missing stuff from the lexgen. need to investigate

whyrusleeping:
There we go, still hate the codegen, but now we handle the case where a struct has a field that is a ref of an array of a union... lexicon is great. I love it.

cmd/gosky/admin.go
```diff
@@ -6,10 +6,12 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/bluesky-social/indigo/api"
+	"github.com/bluesky-social/indigo/api/atproto"
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
```
bnewbold, line 15: duplicate import under a different name?

bnewbold (approved):
Small note about import duplication, but overall looks good.

Would be good to have specific test coverage for new Lexicon lang features (when we add them to lexgen), but there isn't much a framework for that in place and we will probably re-implement all that at some point anyways.

PR #176, opened 2023-05-19
Fewer database operations for event persistence

whyrusleeping:
This is based on the rebase PR since they would conflict otherwise.

### DIFF
--- a/events/dbpersist.go
+++ b/events/dbpersist.go
@@ -34,23 +34,11 @@ type RepoEventRecord struct {
 	Event  string
 	Rebase bool
 
-	Ops []RepoOpRecord
-}
-
-type RepoOpRecord struct {
-	ID                uint `gorm:"primarykey"`
-	RepoEventRecordID uint `gorm:"index"`
-	Path              string
-	Action            string
-	Rec               *util.DbCID
+	Ops []byte
 }
 
 func NewDbPersistence(db *gorm.DB, cs *carstore.CarStore) (*DbPersistence, error) {
 	if err := db.AutoMigrate(&RepoEventRecord{}); err != nil {
-		return nil, err
-	}
-
-	if err := db.AutoMigrate(&RepoOpRecord{}); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


bnewbold:
@whyrusleeping I have some git-level confusion here... https://github.com/bluesky-social/indigo/pull/175, which this PR is basically "on top of" merged a couple days ago, but this PR seems to have a very similar diff. I would have expected the diff to be much smaller against `main` branch now. Maybe i'm just confused? Or maybe there were enough changes to that other branch, this needs to be rebased / cherry-picked?

whyrusleeping:
Oh yeah, this needs to get rebased. Githubs diff computation is funky

events/dbpersist.go
```diff
@@ -34,26 +34,14 @@ type RepoEventRecord struct {
...
-	Ops []RepoOpRecord
-}
-
-type RepoOpRecord struct {
-	ID                uint `gorm:"primarykey"`
-	RepoEventRecordID uint `gorm:"index"`
-	Path              string
-	Action            string
-	Rec               *util.DbCID
+	Ops []byte
```
bnewbold, line 37: I think that stream CBOR size limits (a few MByte?) would prevent the ops list from being so large it results in a DB error. But it might be worth a sanity check anyways, that the JSON encoded bytes aren't more than a MByte or something like that?

IIRC we exclude "huge" events at the block (data) level, but don't remember if we include all the ops in that case. Eg, if you do a huge delete of 10k records in a single repo request (to PDS), does the full ops list get emitted?

bnewbold (approved):
Fewer tables, less work, love it!

Left a comment about a sanity check but otherwise LGTM.

How would we deploy this to prod? Does it need a full database wipe/rebuild? Might want to clarify that before merging to `main`

whyrusleeping:
Yeah, we have to do a full database wipe anyways because of the forks in the PDS output stream. the PDS was emitting forks in repo history which the BGS can't recover from, so those repos are permanently borked in the existing db. We can add a 'reset repo' endpoint, but then we would have to invoke it many thousands of times and would end up with weird discontinuities in *our* output stream (which consumers *should* be able to deal with, but it still sucks).

events/dbpersist.go
whyrusleeping: Yeah, near the top of the persist method we truncate ops to be max 8192 long (can make that more aggressive too)

PR #177, opened 2023-05-19
add sorting and pagination and 'and' to search

whyrusleeping:
Not sure why ive been dragging my feet on this but this should make search suck less

### DIFF
--- a/search/handlers.go
+++ b/search/handlers.go
@@ -2,6 +2,8 @@ package search
 
 import (
 	"context"
+	"fmt"
+	"strconv"
 	"strings"
 
 	api "github.com/bluesky-social/indigo/api"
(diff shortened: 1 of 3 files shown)


whyrusleeping:
While im pretty confident this will work, elasticsearch has bitten me before.
The following curl command i have tested against the prod search instance, but i cant easily hook this service up to the prod deployment because the VPN and firewalling make it very difficult.
```
curl -u '$SEARCHCREDS' -H "Content-Type: application/json" -d '{                                    "query": {
        "match": {
            "text": {
                "query": "cat dog",
                "operator": "and"
            }
        }
    },
    "sort": [
        {
            "createdAt": {
                "order": "desc"
            }
        }
    ],
    "size": 30,
    "from": 0
}
' https://$SEARCHDOMAIN/posts/_search
```

whyrusleeping:
also for reference the mapping schema looks like this:
```json
{"posts":{"mappings":{"properties":{"createdAt":{"type":"long"},"text":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"user":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}}}}}}
```

PR #179, opened 2023-05-21
integrate consumer pool into stream handler

whyrusleeping:
Add an easy way to process the event stream in parallel

### DIFF
--- a/bgs/fedmgr.go
+++ b/bgs/fedmgr.go
@@ -231,7 +231,7 @@ func (s *Slurper) handleConnection(ctx context.Context, host *models.PDS, con *w
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	return events.HandleRepoStream(ctx, con, &events.RepoStreamCallbacks{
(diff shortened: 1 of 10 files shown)


whyrusleeping:
I think I want to deprecate the old sequential interface entirely, or maybe i can just stub it into the new interface with a non-parallel event pool

strideynet:
> I think I want to deprecate the old sequential interface entirely, or maybe i can just stub it into the new interface with a non-parallel event pool

From an external API consumer point of view, I support this. 

As Bluesky grows, the number of events coming over is going to grow, and the callback style method is just going to be unsuitable for most consumers who spend anything more than a non-trivial amount of time on each event. I'm already operating a worker pool for mine, and I think assuming the worker pool implementation shipped in this repo is solid then there's no reason why this wouldn't be a suitable approach for most/all consumers. It also probably keeps things way simpler to have "the one right way" of consuming the websocket, especially this early on.

whyrusleeping:
Yeah, I just switched everything over to use the new interface and added an option to have a 'sequential' scheduler that replicates the old behavior

PR #182, opened 2023-05-26
add did:web support

whyrusleeping:


did/web.go
```diff
@@ -0,0 +1,44 @@
...
+func (wr *WebResolver) GetDocument(ctx context.Context, didstr string) (*Document, error) {
+	pdid, err := did.ParseDID(didstr)
+	if err != nil {
+		return nil, err
+	}
+
+	val := pdid.Value()
+	if strings.Contains(val, ":") {
+		return nil, fmt.Errorf("did:web resolver does not handle ports or documents at sub-paths")
+	}
```
bnewbold, line 26: :+1:

did/web.go
```diff
@@ -0,0 +1,44 @@
...
+
+	resp, err := http.Get("https://" + val + "/.well-known/did.json")
```
bnewbold, line 28: Other hardening checks we should do, either now or soon:

- no IP addresses allowed, needs to be a hostname, including both IPv4 and IPv6, with or without a bracket. I guess the colon check mostly disallows IPv6?
- a bit ambigious about hex-encoded colon and portname; simplest to just disallow hex-encoding and require a complete valid hostname
- don't allow bare hostnames ("blah", must be "blah.com"); don't allow trailing-period syntax (even if usually allowed for DNS/hostnames). possible special-case for localhost in dev/testing, but simplest to just disallow
- disallow TLDs which start with a digit (to prevent IP address confusion)
- disallow a set of meta/reserved TLDs: .example, .invalid, .local, .arpa, .onion, .internal. we allow .test for testing in other contexts

bnewbold (approved):
Left some comments about hardening checks, but otherwise looks good. moving things in to a a local `did` package seems like a good move

bnewbold:
Changes look good!

PR #183, opened 2023-05-29
add command to help debug feed generators

whyrusleeping:
Some basic checks to help find issues with feed generator deployments

### DIFF
--- a/cmd/gosky/debug.go
+++ b/cmd/gosky/debug.go
@@ -312,6 +312,25 @@ var debugFeedGenCmd = &cli.Command{
 			Host: ss.ServiceEndpoint,
 		}
 
+		desc, err := bsky.FeedDescribeFeedGenerator(ctx, fgclient)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("Found %d feeds at discovered endpoint\n", len(desc.Feeds))
+		var found bool
+		for _, f := range desc.Feeds {
+			fmt.Println("Feed: ", f.Uri)
+			if f.Uri == uri {
+				found = true
+				break
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


cmd/gosky/debug.go
```diff
@@ -244,3 +251,76 @@ var debugStreamCmd = &cli.Command{
...
+			return fmt.Errorf("No '#bsky_fg' service entry found in feedgens DID document")
+		}
+
+		fmt.Println("Service endpoint is: ", ss.ServiceEndpoint)
+
+		fgclient := &xrpc.Client{
+			Host: ss.ServiceEndpoint,
+		}
+
+		skel, err := bsky.FeedGetFeedSkeleton(ctx, fgclient, "", uri, 30)
```
dholms, line 315: this might fail if the route is behind auth

may be worth disambiguating in the error handling - but also since this is used by devs, probably not a big deal

cmd/gosky/debug.go
```diff
@@ -244,3 +251,76 @@ var debugStreamCmd = &cli.Command{
...
+		if err != nil {
+			return fmt.Errorf("failed to fetch feed skeleton: %w", err)
+		}
+
+		if len(skel.Feed) > 30 {
```
dholms, line 320: nice great call 👍

dholms:
`app.bsky.feed.describeFeedGenerator` is a required route as well that announces the feed generator's DID & available feeds

I'd add that check as well 👌

cmd/gosky/debug.go
whyrusleeping: Ah, good point. Is there any way for me to generate the correct auth token?

PR #187, opened 2023-06-03
Refactor to have event persisters do writethrough broadcast

whyrusleeping:


events/dbpersist.go
```diff
@@ -5,56 +5,228 @@ import (
...
+	MaxTimeBetweenFlush time.Duration
+	UIDCacheSize        int
+	DIDCacheSize        int
+}
+
+func DefaultOptions() *Options {
+	return &Options{
+		MaxBatchSize:        200,
+		MinBatchSize:        10,
+		MaxTimeBetweenFlush: 500 * time.Millisecond,
```
whyrusleeping, line 39: probably also add a config option for the check interval

events/dbpersist.go
```diff
@@ -5,56 +5,228 @@ import (
...
+	p.batch = []*PersistenceBatchItem{}
+	p.lastFlush = time.Now()
+
+	return nil
+}
+
+func (p *DbPersistence) AddItemToBatch(ctx context.Context, rec *RepoEventRecord, evt *XRPCStreamEvent) error {
+	p.lk.Lock()
+	if p.batch == nil {
+		p.batch = []*PersistenceBatchItem{}
```
whyrusleeping, line 164: this bits actually not necessary, `append(nil, thing)` does the right thing

events/dbpersist.go
```diff
@@ -5,56 +5,228 @@ import (
...
+
+	return nil
 }
 
 func (p *DbPersistence) Persist(ctx context.Context, e *XRPCStreamEvent) error {
-	if e.RepoCommit == nil {
+	var rer *RepoEventRecord
+	var err error
+
+	if e.RepoCommit != nil {
```
whyrusleeping, line 189: cleaner to use a switch here

events/dbpersist.go
```diff
@@ -137,24 +314,51 @@ func (p *DbPersistence) Playback(ctx context.Context, since int64, cb func(*XRPC
...
 	return u.Did, nil
 }
 
-func (p *DbPersistence) hydrateRepoEvent(ctx context.Context, rer *RepoEventRecord) (*comatproto.SyncSubscribeRepos_Commit, error) {
+func (p *DbPersistence) hydrateHandleChange(ctx context.Context, rer *RepoEventRecord) (*XRPCStreamEvent, error) {
+	did, err := p.didForUid(ctx, rer.Repo)
+	if err != nil {
+		return nil, err
+	}
+
```
whyrusleeping, line 351: should check that the handle isnt nil

events/dbpersist.go
```diff
@@ -137,24 +314,51 @@ func (p *DbPersistence) Playback(ctx context.Context, since int64, cb func(*XRPC
...
+	return &XRPCStreamEvent{
+		RepoHandle: &comatproto.SyncSubscribeRepos_Handle{
+			Did:    did,
+			Handle: *rer.NewHandle,
+			Time:   rer.Time.Format(util.ISO8601),
+		},
+	}, nil
+}
+
+func (p *DbPersistence) hydrateCommit(ctx context.Context, rer *RepoEventRecord) (*XRPCStreamEvent, error) {
```
whyrusleeping, line 361: lets sanity check that the rer.Commit isnt nil at some point in this function, now that we've changed it to being a pointer

whyrusleeping (commented):
Thanks for the contributions @ericvolp12 !

PR #188, opened 2023-06-04
Feat/evt playback pipeline

ericvolp12:


ericvolp12:
@whyrusleeping I was able to get a like 2x speedup using a batching pattern for hydration and maintaining serialized emits.

```
BenchmarkPlayback-16    	    1332	    932659 ns/op	  135399 B/op	    2488 allocs/op
BenchmarkPlayback-16    	    2086	    517006 ns/op	  144658 B/op	    2624 allocs/op
```

Double check my before/after runtimes?

ericvolp12:
I tried speeding up the carstore with WAL and the SQLite cache but it didn't have any impact on the benchmark.

ericvolp12:
That was a really not fun rebase lmao, I think I got it tho.

whyrusleeping:
Okay, im doing a rebuild of my dev bgs instance. Currently getting around 850 events per second on playback, once I get the rebuild done i'll give this PR a try and see how we did

whyrusleeping:
Testing against my dev bgs running on prod.
Current performance before this PR is ~200 events per second for 'warmed' events, and ~80-100 events per second for unwarmed (OS level caching seems to play a decently big part here)

With this PR, i'm averaging around 550 events per second, *but* it randomly hangs and outputs no events for seconds at a time. Kinda weird and probably worth looking into, but it *is* faster.

whyrusleeping:
Managed to get it a little bit faster and *much* more consistent by making hydrateBatch call the callback directly instead of having it collect everything and return it as a slice. 

I couldnt manage to rebase this easily so im just going to push the monkeypatched version to a new branch for now.

whyrusleeping:
closing since we merged #209

Issue #165, opened 2023-05-09
Integration tests need randomly assigned ports


whyrusleeping:
Our integration test setup (testing/) currently hardcodes a port for each service it spins up. This is cumbersome and error prone, we should instead listen on port 0 and detect the port the OS chooses for us, then set that field in the struct from there.

PR #193, opened 2023-06-12
support booting PDS on a random port in testing

jmhodges:
This patch makes PDS boot on a random port in the integration tests.
It's a first step towards #165 ("Integration tests need randomly
assigned ports") and request for feedback on its strategy before I apply
it to BDS.

To do this work, we have to teach `pds.Server` how to boot with an
existing net.Listener and teach its Echo instance to do the same. That
required using `Echo#StartServer` instead of `Echo#Start`. Using
`Echo#StartServer` seems to be the accepted pattern for booting from an
existing Listener in Echo and both it and `Start` call the same
`Echo#configureServer` under the hood.

PDS isn't a production service right now, but I believe I've reproduced the 
configuration that echo would have done that there wouldn't be production 
impact. However, if BGS is, we could talk about what feature flagging we'd 
like to prevent problems when this is applied there.

Along the way, we refactor the uses of the `TestPDS.host` field into
more explicit methods on `TestPDS`. I'm not completely sold on those
methods' names, but they seem handy.

Updates #165

### DIFF
--- a/pds/server.go
+++ b/pds/server.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
+	"net/http"
 	"net/mail"
 	"net/url"
 	"strings"
@@ -265,7 +267,22 @@ func (s *Server) readRecordFunc(ctx context.Context, user bsutil.Uid, c cid.Cid)
 	return lexutil.CborDecodeValue(blk.RawData())
 }
 
-func (s *Server) RunAPI(listen string) error {
+func (s *Server) RunAPI(addr string) error {
+	var lc net.ListenConfig
+	// This is an arbitrary timeout that should be safe on any platform, but
+	// there's no great way to weave this timeout without adding another
+	// parameter to the (at time of writing) long signature of NewServer.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	li, err := lc.Listen(ctx, "tcp", addr)
+	if err != nil {
+		return err
+	}
+	return s.RunAPIWithListener(li)
+}
+
+func (s *Server) RunAPIWithListener(listen net.Listener) error {
 	e := echo.New()
 	s.echo = e
 	e.HideBanner = true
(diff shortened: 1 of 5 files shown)


jmhodges:
I also could be easily convinced to add some default socket timeouts to the `http.Server` instantiation. 

But I wasn't sure what your desired deadlining strategy was and thought maybe the server's individual RPC methods were expected to handle them, instead, since echo doesn't add those out of the box. That plus having to add yet more fields to `pds.NewServer` felt bad made we wait for feedback to do it.

pds/server.go
```diff
@@ -265,7 +274,19 @@ func (s *Server) readRecordFunc(ctx context.Context, user bsutil.Uid, c cid.Cid)
 	return lexutil.CborDecodeValue(blk.RawData())
 }
 
-func (s *Server) RunAPI(listen string) error {
+func (s *Server) RunAPI(addr string) error {
+	var lc net.ListenConfig
+	ctx, cancel := context.WithTimeout(context.Background(), serverListenerBootTimeout)
+	defer cancel()
+
+	li, err := lc.Listen(ctx, "tcp", addr)
```
whyrusleeping, line 282: just to confirm, since the docs arent super clear without diving into the implementation, the lifetime of this context only governs the creation of the listener and allocation of the file descriptor, right? Its not going to break the listener once the timeout occurs, is it?

whyrusleeping (approved):
Aside from the question around the behavior of the context with the listen function, this looks good! Thank you for the PR (and apologies for the delay, i'm not used to checking for other peoples PRs here yet)

pds/server.go
jmhodges: Correct! I've used this API successfully in other cases but also I traced through all the code (through [internetSocket](https://github.com/golang/go/blob/dbf9bf2c39116f1330002ebba8f8870b96645d87/src/net/ipsock_posix.go#L137), and [netFD#dial](https://github.com/golang/go/blob/dbf9bf2c39116f1330002ebba8f8870b96645d87/src/net/sock_posix.go#L116) and such) to double-check. And, yep, it's just for the boot time. I might post them a PR to add to the docs.

jmhodges:
Dope! Sorry, did a rebase to latest `main` and now need a maintainer approval to run the workflows on this PR again.

PR #194, opened 2023-06-12
use errors.Is instead of xerrors.Is

jmhodges:
Cleaning up a deprecation warning that popped up in my editor while
working on other things. xerrors has marked itself as deprecated for Go
versions 1.13 and later.

### DIFF
--- a/pds/server.go
+++ b/pds/server.go
@@ -31,7 +31,6 @@ import (
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/whyrusleeping/go-did"
-	"golang.org/x/xerrors"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Ah interesting, is there another package that does the stack traces with the logs now? Or did they somehow merge that into the errors package?

jmhodges:
> Ah interesting, is there another package that does the stack traces with the logs now? Or did they somehow merge that into the errors package?

Oh, I'm not sure which functionality you're referring to (I suspect I'm not sure which logs you're referring to). If you mean the `Unwrap`/`Wrap` stuff, and that showing up in output when you use `%w`, then yep. [Example that I re-purposed from someone's blogpost](https://go.dev/play/p/hYtQx9dSFlV)

whyrusleeping:
So the reason i habitually use xerrors is that it would embed a stack trace with the error, so that if you printed it with %v, you could tell the entire chain of the error wrapping that happened and what lines things were on

jmhodges:
Ah, they were not. The `Is` works the same, though, and is the function that's actually been deprecated. The `xerrors.New` have been left alone.

whyrusleeping:
Ah, great. Thanks for the clarification

PR #196, opened 2023-06-16
update handle resolution to use new well-known method

whyrusleeping:


pds/stubs.go
```diff
@@ -804,7 +804,6 @@ func (s *Server) RegisterHandlersComAtproto(e *echo.Echo) error {
 	e.POST("/xrpc/com.atproto.admin.takeModerationAction", s.HandleComAtprotoAdminTakeModerationAction)
 	e.POST("/xrpc/com.atproto.admin.updateAccountEmail", s.HandleComAtprotoAdminUpdateAccountEmail)
 	e.POST("/xrpc/com.atproto.admin.updateAccountHandle", s.HandleComAtprotoAdminUpdateAccountHandle)
-	e.GET("/xrpc/com.atproto.identity.resolveHandle", s.HandleComAtprotoIdentityResolveHandle)
```
bnewbold, line 807: This is still a real lexicon endpoint, used by clients to resolve handlers without going out to the network themselves.

The PDS isn't "real" so don't need to implement it, but seems like it should at least stay a stub with an HTTP "not implemented" error.

pds/server.go
```diff
@@ -164,15 +165,17 @@ func (s *Server) createExternalUser(ctx context.Context, did string) (*models.Ac
 	c := &xrpc.Client{Host: svc.ServiceEndpoint}
 
 	if peering.ID == 0 {
-		pdsdid, err := comatproto.IdentityResolveHandle(ctx, c, "")
+		cfg, err := atproto.ServerDescribeServer(ctx, c)
```
bnewbold, line 168: :+1:

cmd/gosky/debug.go
```diff
@@ -254,6 +261,172 @@ var debugStreamCmd = &cli.Command{
 	},
 }
 
+var compareStreamsCmd = &cli.Command{
```
bnewbold, line 264: seems like some other stuff got rolled up in this PR; not reviewing these bits

cmd/gosky/admin.go
```diff
@@ -364,6 +366,13 @@ var listReportsCmd = &cli.Command{
 		&cli.BoolFlag{
 			Name: "raw",
 		},
+		&cli.BoolFlag{
+			Name:  "resolved",
+			Value: true,
+		},
+		&cli.BoolFlag{
+			Name: "template-output",
```
bnewbold, line 374: was this added accidentally? don't see it used (at least in the PR diff)

bgs/bgs.go
```diff
@@ -48,6 +49,11 @@ type BGS struct {
 	didr    did.Resolver
 
 	blobs blobs.BlobStore
+	hr    api.HandleResolver
+
+	// TODO: work on doing away with this flag in favor of more pluggable
```
bnewbold, line 54: :+1:

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
 
-	if verdid.Did != udid {
+	if verdid != udid {
 		return "", "", fmt.Errorf("pds server reported different did for claimed handle")
 	}
 
 	return handle, svc.ServiceEndpoint, nil
 }
+
+type HandleResolver interface {
```
bnewbold, line 70: feel like this should go somewhere other than `api/extra.go`... maybe in util, or a new top-level package? but not urgent, can live here for now.

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
+	ResolveHandleToDid(ctx context.Context, handle string) (string, error)
+}
+
+type ProdHandleResolver struct {
+}
+
+func (dr *ProdHandleResolver) ResolveHandleToDid(ctx context.Context, handle string) (string, error) {
```
bnewbold, line 77: It would make sense to me to consolidate identifier checks on `handle` here in this method, even if redundant with previous validation. We directly inject the string in to a URL below, and should ensure we aren't doing weird security things like probing our own internal/private/VPN network, even if "just" doing an HTTP GET should be safe. Eg, a malicious handle string could have username/pass, hostname (including IP addresses and port), path, and query parameters.

I don't remember if we have a "check if valid hostname" function in indigo yet, but a quick version could probably re-use golang URL verification, then ensure that none of the weird/meta TLDs like `.arpa` are used. Full list of those weird TLDs at:

https://atproto.com/specs/handle#additional-non-syntax-restrictions

bnewbold: Doing full handle validation isn't a blocking thing for this PR, but if you do want to knock it out there is typescript code and example test cases here:

https://github.com/bluesky-social/atproto/blob/main/packages/identifier/tests/handle.test.ts
https://github.com/bluesky-social/atproto/blob/main/packages/identifier/src/handle.ts

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
+		parsed, err := did.ParseDID(string(b))
+		if err != nil {
+			return "", err
+		}
+
+		return parsed.String(), nil
+	}
+	log.Infof("failed to resolve handle (%s) through well-known route: %s", handle, wkerr)
+
+	res, err := net.LookupTXT("_atproto." + handle)
```
bnewbold, line 96: In typescript the HTTP and TXT resolution happens in parallel (using async). The consensus when chatting with Dan was to prefer DNS TXT over HTTPS when doing them sequentially, so try the DNS method before the HTTPS method. We consider the HTTPS method a fallback for large hosting providers etc.

This isn't blocking, especially if it causes the test/dev flow to break. Implementations can do whichever prioritization they want.

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
+		}
+
+		parsed, err := did.ParseDID(string(b))
+		if err != nil {
+			return "", err
+		}
+
+		return parsed.String(), nil
+	}
+	log.Infof("failed to resolve handle (%s) through well-known route: %s", handle, wkerr)
```
bnewbold, line 94: "through HTTP well-known route"

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
+
+type HandleResolver interface {
+	ResolveHandleToDid(ctx context.Context, handle string) (string, error)
+}
+
+type ProdHandleResolver struct {
+}
+
+func (dr *ProdHandleResolver) ResolveHandleToDid(ctx context.Context, handle string) (string, error) {
+	c := http.DefaultClient
```
bnewbold, line 78: This HTTP client in particular should have a short timeout, like 5-10 seconds, I think. Not formalized anywhere, but don't want to spend a bunch of time waiting for what should be a very quick response.

api/extra.go
```diff
@@ -49,14 +55,102 @@ func ResolveDidToHandle(ctx context.Context, xrpcc *xrpc.Client, res did.Resolve
...
+
+	resp, wkerr := c.Get(fmt.Sprintf("https://%s/.well-known/atproto-did", handle))
+	if wkerr == nil && resp.StatusCode == 200 {
+		b, err := ioutil.ReadAll(resp.Body)
```
bnewbold, line 82: should probably limit the body size to just a couple KBytes. can leave for a hardening iteration, but a TODO reminder here would be good.

bnewbold (approved):
Overall looks good. Left several comments.

Don't want to block progress, but do expect a bunch of shenanigans around handle string format, so should do basic validation of that identifier before doing a network request. Slapping a tighter timeout on the HTTP validation would also be a quick tweak that could prevent infra from getting overwhelmed.

pds/stubs.go
whyrusleeping: Ah, I missed that, thought we did away with it entirely for some reason

cmd/gosky/admin.go
whyrusleeping: Yeah, some other WIP progress got rolled up here by accident. Was very carefully avoiding pulling it in until i wasnt..

cmd/gosky/debug.go
whyrusleeping: yeah, got pulled in unintentionally, not a big deal but just ignore for this PRs purposes

api/extra.go
whyrusleeping: yeah, it might deserve its own package once we feel it out a bit more. For now though all users were previously using the api package for the old behavior, so this is 'drop in' in terms of deps

api/extra.go
whyrusleeping: Made an issue for it with #200, dont want to block getting this to sandbox but will get it done soon

PR #197, opened 2023-06-19
Add support for OpenTelemetry HTTP collector

Jacob2161:
I had to update the Gorm tracing plugin to get support for current version of OpenTelemetry tracing, which also updated some other deps. Please verify that this is okay!


whyrusleeping (approved):
Seems good to me! Thanks Jake

PR #198, opened 2023-06-19
add ability to ban entire domains, including subdomains

whyrusleeping:
Add the ability to ban domains and subdomains from being scraped by the bgs

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -642,6 +642,19 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
 		return nil, err
 	}
 
+	if peering.Blocked {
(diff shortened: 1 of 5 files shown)


bgs/bgs.go
```diff
@@ -406,6 +407,39 @@ func prometheusHandler() http.Handler {
 	return exporter
 }
 
+func (s *BGS) domainIsBanned(ctx context.Context, host string) (bool, error) {
```
bnewbold, line 410: A doc comment explaining that this works by checking the full domain, then all the parent domains, up to and including the TLD, would be helpful.

bnewbold (approved):
A comment explaining the logic would be nice, but otherwise LGTM.

PR #202, opened 2023-06-21
support booting BGS on a random port in testing

### LINKED ISSUE #165: Integration tests need randomly assigned ports

jmhodges:
This patch makes BGS boot on a random port in the integration tests.

To do this work, we have to teach `bgs.Server` how to boot with an
existing net.Listener and teach its Echo instance to do the same. That
required using `Echo#StartServer` instead of `Echo#Start`. Using
`Echo#StartServer` seems to be the accepted pattern for booting from an
existing Listener in Echo and both it and `Start` call the same
`Echo#configureServer` under the hood.

Along the way, we refactor the uses of the `TestBGS.host` field into a
`TestBGS#Host` method.

Fixes #165

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	_ "net/http/pprof"
 	"net/url"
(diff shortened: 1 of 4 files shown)


whyrusleeping (approved):
Wonderful, thank you so much for this!

PR #203, opened 2023-06-26
optimize block_ref creation

whyrusleeping:
The GORM create helpers dont have any way to avoid 'returning' the ID values of created records, and that was causing a noticeable performance hit in this critical section.

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Hard to notice too much of a difference in the existing benchmarks because we're using an in-memory sqlite instance, but heres the numbers:
Before:
```
why@sirius ~/c/indigo (main)> go test ./carstore -bench=.
7256
goos: linux
goarch: amd64
pkg: github.com/bluesky-social/indigo/carstore
cpu: AMD Ryzen Threadripper 3970X 32-Core Processor
BenchmarkRepoWritesCarstore-64    	    2698	    535345 ns/op
BenchmarkRepoWritesFlatfs-64      	    2444	    540607 ns/op
BenchmarkRepoWritesSqlite-64      	    5752	    339228 ns/op
PASS
ok  	github.com/bluesky-social/indigo/carstore	4.878s
```

After:
```
why@sirius ~/c/indigo (feat/carstore-opti)> go test ./carstore -bench=.
9931
goos: linux
goarch: amd64
pkg: github.com/bluesky-social/indigo/carstore
cpu: AMD Ryzen Threadripper 3970X 32-Core Processor
BenchmarkRepoWritesCarstore-64    	    3488	    483285 ns/op
BenchmarkRepoWritesFlatfs-64      	    2767	    509982 ns/op
BenchmarkRepoWritesSqlite-64      	    6513	    326489 ns/op
PASS
ok  	github.com/bluesky-social/indigo/carstore	5.370s
```

Jacob2161 (approved):
Batching is good :+1:

PR #212, opened 2023-07-04
Supercollider: a loadtesting tool for the BGS

ericvolp12:
We need to characterize the performance of the BGS implementation so I'm building `supercollider` which is a very noisy PDS that serves to emit lots of events of a configurable mix on the `subscribeRepos` socket.

It will track the event throughput on all connections and make it easy to test the performance of BGS implementations.


ericvolp12:
Benching on this latest version using the in-memory persister:
image

Looks kinda weird, throughput on the event manager is pretty low for emitted events when lots of events are being added but during playback (the third peak), the emission rate goes up a ton. Wonder what's going on there.

ericvolp12:
This test had three phases
1. Produce 1M events and then connect 3 sockets with cursor=0 for full playback
2. Produce 10M events with 3 sockets connected (starting from live)
3. Connect 3 sockets to an event manager that's already got 10M events in the queue for playback

image

- Phase 1 shows solid peak throughput at around 60k events per second per socket
- Phase 2 shows sustained throughput of around 16k events per second per socket while persisting at the same time as emitting
- Phase 3 shows higher sustained throughput of 75k events per second per socket

I wonder if this has to do with the event manager buffer size?

ericvolp12:
oops, didn't mean to click that button, guess it's time for bed

ericvolp12:
So I bumped the event manager buffer to `16_384` and managed to get much better emission during event persistence:

image

ericvolp12:
With a rate limiter keeping us at 80k events per second produced, we can keep a dumb socket in sync. This should be a good baseline for per-socket throughput.

image

ericvolp12:
Since the BGS actually checks the validity of events (looking up the repo and making sure the events are in order), we may have to revert back to basically running a PDS here.

cmd/supercollider/main.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	cs, err := carstore.NewCarStore(cardb, cspath)
+	if err != nil {
+		return nil, err
+	}
+
+	hs := repomgr.NewMemHeadStore()
+
+	repoman := repomgr.NewRepoManager(hs, cs, &util.FakeKeyManager{})
```
whyrusleeping, line 358: going to need a real keymanager for this too unfortunately

ericvolp12:
Running with a dev BGS connected to the socket, we can now effectively speedtest a BGS! (at least with posts with the same content coming from a bunch of different users).

```
![CleanShot 2023-07-06 at 14 16 14](https://github.com/bluesky-social/indigo/assets/1617325/77545f9b-6538-4408-b76e-7685b2cc45c9)
```

cmd/bigsky/main.go
```diff
@@ -245,7 +256,9 @@ func run(args []string) {
 
 		var hr api.HandleResolver = &api.ProdHandleResolver{}
 		if cctx.Bool("crawl-insecure-ws") {
-			hr = &api.TestHandleResolver{}
+			hr = &api.TestHandleResolver{
+				TrialHosts: []string{"supercollider.jazco.io"},
```
whyrusleeping, line 260: lets make this a flag/env-var

events/events.go
```diff
@@ -27,7 +27,7 @@ type EventManager struct {
 
 func NewEventManager(persister EventPersistence) *EventManager {
 	em := &EventManager{
-		bufferSize: 1024,
+		bufferSize: 32_768,
```
whyrusleeping, line 30: aka `32 << 10`

events/persist.go
```diff
@@ -88,3 +88,55 @@ func (mp *MemPersister) RebaseRepoEvents(ctx context.Context, usr util.Uid) erro
 func (mp *MemPersister) SetEventBroadcaster(brc func(*XRPCStreamEvent)) {
 	mp.broadcast = brc
 }
+
+// YoloPersister is used for benchmarking, it has no persistence, it just emits events and forgets them
+type YoloPersister struct {
```
whyrusleeping, line 93: toss this in its own file

PR #213, opened 2023-07-05
extract headStore to make it easier to swap out database

whyrusleeping:
The in-memory variant goes about 20% faster in the benchmark. The rest is carstore costs


ericvolp12 (approved):
LGTM

Issue #210, opened 2023-07-01
Add mute and block support to gosky cli


servitor-social:
I'd like to implement mute and block support to gosky cli. I see ```follows``` has ```followsAddCmd``` and ```followsListCmd```

TODO:
- [ ] implement mutes
  - [ ] mutesAddCmd
  - [ ] mutesListCmd
- [ ] Implement blocks
  - [ ] blocksAddCmd
  - [ ] blocksListCmd

PR #214, opened 2023-07-05
Implementeed blocksCmd, blocksAddCmd, and blocksListCmd

servitor-social:
Implementing blocksCmd, blocksAddCmd, and blocksListCmd from #210

### DIFF
--- a/cmd/gosky/main.go
+++ b/cmd/gosky/main.go
@@ -100,6 +100,7 @@ func run(args []string) {
 		createInviteCmd,
 		adminCmd,
 		createFeedGeneratorCmd,
+		blocksCmd,
 	}
 
 	app.RunAndExitOnError()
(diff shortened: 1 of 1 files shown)


cmd/gosky/main.go
```diff
@@ -1469,3 +1470,72 @@ var createFeedGeneratorCmd = &cli.Command{
...
+	Name:      "add",
+	Flags:     []cli.Flag{},
+	ArgsUsage: `<user>`,
+	Action: func(cctx *cli.Context) error {
+		xrpcc, err := cliutil.GetXrpcClient(cctx, true)
+		if err != nil {
+			return err
+		}
+
+		user := cctx.Args().First()
```
whyrusleeping, line 1492: we might want to confirm that this is a DID instead of a handle, or if its a handle resolve it to a DID first

PR #215, opened 2023-07-05
disk based event persistence

whyrusleeping:
Still need to implement some of the trickier parts like takedowns and rebases, but the primary IO flows are here.

### DIFF
--- a/events/diskpersist.go
+++ b/events/diskpersist.go
@@ -23,6 +23,8 @@ type DiskPersistence struct {
 	archiveDir    string
 	eventsPerFile int64
 
+	evts chan *persistJob
+
 	meta *gorm.DB
 
 	broadcast func(*XRPCStreamEvent)
@@ -30,13 +32,23 @@ type DiskPersistence struct {
 	logfi *os.File
 
 	curSeq int64
-	seqLk  sync.Mutex
 
 	uidCache *lru.ARCCache
 	didCache *lru.ARCCache
 
 	buffers *sync.Pool
 	scratch []byte
+}
+
+type persistJob struct {
+	Buf  []byte
+	Evt  *XRPCStreamEvent
+	Done chan jobResult
+}
+
+type jobResult struct {
+	Err error
+	Seq int64
 }
 
 var _ (EventPersistence) = (*DiskPersistence)(nil)
@@ -85,6 +97,7 @@ func NewDiskPersistence(primaryDir, archiveDir string, db *gorm.DB, opts *DiskPe
 		buffers:       bufpool,
 		uidCache:      uidCache,
 		didCache:      didCache,
+		evts:          make(chan *persistJob, 1024),
 		eventsPerFile: opts.EventsPerFile,
 		scratch:       make([]byte, headerSize),
 	}
@@ -92,6 +105,8 @@ func NewDiskPersistence(primaryDir, archiveDir string, db *gorm.DB, opts *DiskPe
 	if err := dp.resumeLog(); err != nil {
 		return nil, err
 	}
+
+	go dp.persistWorker()
 
 	return dp, nil
 }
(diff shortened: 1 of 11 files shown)


events/diskpersist.go
```diff
@@ -1,19 +1,447 @@
...
+	}).Error; err != nil {
+		return err
+	}
+
+	dp.logfi = fi
+	dp.curSeq = 1
+	return nil
+}
+
+func (dp *DiskPersistence) swapLog(ctx context.Context) error {
```
ericvolp12, line 157: Should this rely on the called to lock the seq or should we lock the seq inside this function?

ericvolp12: If the seq needs to be locked by the caller, we should put a comment on this function to remind ourselves later

events/diskpersist.go
```diff
@@ -1,19 +1,447 @@
...
+			var evt atproto.SyncSubscribeRepos_Handle
+			if err := evt.UnmarshalCBOR(io.LimitReader(bufr, h.Len)); err != nil {
+				return err
+			}
+			evt.Seq = h.Seq
+			if err := cb(&XRPCStreamEvent{RepoHandle: &evt}); err != nil {
+				return err
+			}
+		default:
+			log.Warnw("unrecognized event kind coming from log file", "seq", h.Seq, "kind", h.Kind)
```
ericvolp12, line 427: Would be nice to put the byte offset for the log file in here so we can investigate if that's easy

events/diskpersist.go
```diff
@@ -1,19 +1,447 @@
...
+			return fmt.Errorf("halting on unrecognized event kind")
+		}
+	}
+}
+
+func (p *DiskPersistence) TakeDownRepo(ctx context.Context, usr util.Uid) error {
+	panic("no")
```
ericvolp12, line 434: Should this panic or return an error?

whyrusleeping: just not implemented yet, panic to remind myself to finish it before merging

events/diskpersist.go
```diff
@@ -1,19 +1,710 @@
...
+		Evt:    e,
+		Buffer: buffer,
+	})
+}
+
+type evtHeader struct {
+	Flags uint32
+	Kind  uint32
+	Seq   int64
+	Usr   util.Uid
```
ericvolp12, line 413: `util.Uid` is a `uint` which is system dependent, you probably want to change that to be a `uint32` or something fixed-width if we're gonna put it the header size below and depend on it being a fixed width.

events/diskpersist.go
```diff
@@ -1,19 +1,710 @@
...
+	scratch = scratch[:headerSize]
+	_, err := io.ReadFull(r, scratch)
+	if err != nil {
+		return nil, fmt.Errorf("reading header: %w", err)
+	}
+
+	flags := binary.LittleEndian.Uint32(scratch[:4])
+	kind := binary.LittleEndian.Uint32(scratch[4:8])
+	l := binary.LittleEndian.Uint32(scratch[8:12])
+	usr := binary.LittleEndian.Uint64(scratch[12:20])
```
ericvolp12, line 433: Same as above comment, this is fine if we set `usr` to be a fixed-width `uint64` (though in the header above it seems like you want it to be a `uint32`?), also above you have `l` here as a `uint32` but you've declared it in the `evtHeader` struct as an `int64`. Which is it? I think these short variable names don't help when trying to do byte manipulation fwiw.

events/diskpersist.go
```diff
@@ -1,19 +1,710 @@
...
 }
 
-func (p *DiskPersistence) Persist(e *XRPCStreamEvent) error {
-	panic("nyi")
+func (p *DiskPersistence) writeHeader(ctx context.Context, flags uint32, kind uint32, l uint32, usr uint64, seq int64) error {
+	binary.LittleEndian.PutUint32(p.scratch, flags)
+	binary.LittleEndian.PutUint32(p.scratch[4:], kind)
+	binary.LittleEndian.PutUint32(p.scratch[8:], l)
+	binary.LittleEndian.PutUint64(p.scratch[12:], usr)
+	binary.LittleEndian.PutUint64(p.scratch[20:], uint64(seq))
```
ericvolp12, line 450: Revise this once you solidify the schema for the header above.

ericvolp12 (changes requested):

ericvolp12 (approved):
LGTM

PR #217, opened 2023-07-07
Create Sonar, an ATProto RepoSync Stats Generator

ericvolp12:
Sonar is an ATProto SubscribeRepos firehose listener that emits Prometheus metrics to characterize the kind and volume of content moving through the firehose.


ericvolp12:
Here's an example Grafana dashboard powered by Sonar

You can switch between different instances of Sonar to see the view of the firehose from different geographical (and network topological) locations.

I think I want to make a small change before merging that adds a label for the host the instance is connecting to so it's easy to switch your view between different PDSs or BGS firehoses.

image

cmd/sonar/main.go
```diff
@@ -0,0 +1,220 @@
...
+		mux.Handle("/metrics", promhttp.Handler())
+		log.Infof("metrics server listening on port %d", cctx.Int("port"))
+		err = http.ListenAndServe(fmt.Sprintf(":%d", cctx.Int("port")), mux)
+		if err != nil {
+			cancel()
+			log.Fatalf("failed to start metrics server: %+v\n", err)
+		}
+	}()
+
+	s.ProgMux.Lock()
```
whyrusleeping, line 199: no real reason to lock here, nothing is writing to these values yet

sonar/sonar.go
```diff
@@ -0,0 +1,287 @@
...
+func (s *Sonar) HandleStreamEvent(ctx context.Context, xe *events.XRPCStreamEvent) error {
+	ctx, span := otel.Tracer("sonar").Start(ctx, "HandleStreamEvent")
+	defer span.End()
+
+	switch {
+	case xe.RepoCommit != nil:
+		eventsProcessedCounter.WithLabelValues("repo_commit").Inc()
+		return s.HandleRepoCommit(ctx, xe.RepoCommit)
+	case xe.RepoHandle != nil:
+		eventsProcessedCounter.WithLabelValues("repo_handle").Inc()
```
whyrusleeping, line 111: RepoHandle also has a 'seq' field we should update on just like in RepoCommit

whyrusleeping:
build fails because theres a directory named sonar as well as the binary

whyrusleeping:
current test failure fixed in #218

whyrusleeping (approved):
LGTM

PR #222, opened 2023-07-12
fix cleanup of rebases in carstore

whyrusleeping:
This fixes two things, first the test-pds didnt properly implement rebases which caused some issues in tests. 
The second thing was that old blockRefs werent getting deleted after the rebase, despite their associated CarShards getting deleted.

### DIFF
--- a/indexer/crawler.go
+++ b/indexer/crawler.go
@@ -196,7 +196,7 @@ func (c *CrawlDispatcher) fetchWorker() {
 
 func (c *CrawlDispatcher) Crawl(ctx context.Context, ai *models.ActorInfo) error {
 	if ai.PDS == 0 {
-		panic("not today!")
+		panic("must have pds for user in queue")
 	}
 
 	ctx, span := otel.Tracer("crawler").Start(ctx, "addToCrawler")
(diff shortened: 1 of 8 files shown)


indexer/crawler.go
```diff
@@ -214,12 +215,19 @@ func (c *CrawlDispatcher) AddToCatchupQueue(ctx context.Context, host *models.PD
 		panic("not okay")
```
ericvolp12, line 215: Why do we panic here?

indexer/indexer.go
```diff
@@ -838,28 +838,48 @@ func (ix *Indexer) FetchAndIndexRepo(ctx context.Context, job *crawlWork) error
 	}
 
 	var rebase *comatproto.SyncSubscribeRepos_Commit
-	for _, job := range job.catchup {
-		if job.evt.Rebase {
-			rebase = job.evt
+	var rebaseIx int
+	for i, j := range job.catchup {
+		if j.evt.Rebase {
+			fmt.Println("REBASE AT: ", i, len(job.catchup))
```
ericvolp12, line 844: Do we have a logger we can use for this?

ericvolp12 (approved):
This looks okay to me but I've also not got a ton of context on the indexer code yet.

indexer/crawler.go
whyrusleeping: I should make those messages better... but we need to be aware of which PDS the work is coming from for scheduling purposes

indexer/indexer.go
whyrusleeping: oh whoops, yes.

PR #233, opened 2023-07-19
add in smarter timestamp parsing

whyrusleeping:
i'm surely missing some that are *Technically* valid, but i'd prefer we just got more strict on whats allowed.


bnewbold (approved):
there are some missing cases, but this is an improvement and realistically covers the vast majority of real-world vals.

PR #235, opened 2023-07-20
Track websocket consumers, expose them via an admin endpoint, list th…

ericvolp12:
…em in the UI, clean up connections when they stop responding

CleanShot 2023-07-20 at 10 58 20


bgs/bgs.go
```diff
@@ -394,26 +413,86 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
...
+	defer cleanup()
+
+	// Keep track of the consumer for metrics and admin endpoints
+	consumer := SocketConsumer{
+		RemoteAddr:  c.RealIP(),
+		UserAgent:   c.Request().UserAgent(),
+		ConnectedAt: time.Now(),
+	}
+
+	sentCounter := eventsSentCounter.WithLabelValues(consumer.RemoteAddr, consumer.UserAgent)
```
whyrusleeping, line 459: the cardinality on this metric is probably going to be very high

ericvolp12: I'm guessing it'll be under 10k series initially, basically one series per unique consumer (though repeat consumers will end up using the same series). If it starts to get too high we can try reducing it but I like the idea of remote address and/or user-agent tracking in a visible way.

ericvolp12: Also it'll get GC'd on BGS restart so we don't have a ton of inactive series to track (hopefully).

ericvolp12: We could maybe track it internally at this level and via prometheus only at the RemoteAddr level to start

bgs/bgs.go
```diff
@@ -394,26 +413,86 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
...
+	consumer.EventsSent = sentCounter
+
+	bgs.consumersLk.Lock()
```
whyrusleeping, line 462: probably deserves a `bgs.registerConsumer()` method

bgs/bgs.go
```diff
@@ -394,26 +413,86 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
...
+	bgs.nextConsumerID++
+	consumerID := bgs.nextConsumerID
+	bgs.consumers[bgs.nextConsumerID] = &consumer
+	bgs.consumersLk.Unlock()
+
+	// Cleanup the consumer when we're done
+	defer func() {
```
whyrusleeping, line 469: and `bgs.cleanupConsumer()`

ts/bgs-dash/src/components/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,352 @@
+import { ChevronDownIcon, ChevronUpIcon } from "@heroicons/react/24/solid";
```
whyrusleeping, line 1: image

bgs/metrics.go
```diff
@@ -19,3 +19,8 @@ var rebasesCounter = promauto.NewCounterVec(prometheus.CounterOpts{
 	Name: "event_rebases",
 	Help: "The total number of rebase events received",
 }, []string{"pds"})
+
+var eventsSentCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "events_sent_counter",
+	Help: "The total number of events sent to consumers",
+}, []string{"remote_addr", "user_agent"})
```
whyrusleeping, line 26: smells like lots of metrics

bgs/bgs.go
whyrusleeping: yeah i guess that seems fine for now, lets just keep an eye on it

whyrusleeping (approved):
LGTM

PR #241, opened 2023-07-21
use parallel event processor in bgs

whyrusleeping:
There might be other test failures as a result of this, but i ran them 20 or so times and this was the only one that popped up locally.

We might want to make the settings on the consumer pool tuneable, but for now this should get us some much better numbers on our load testing


ericvolp12 (approved):
LGTM

PR #242, opened 2023-07-21
do handle resolution in parallel

whyrusleeping:
This is the proper spec compliant way of doing things.


api/extra.go
```diff
@@ -83,6 +85,41 @@ func (dr *ProdHandleResolver) ResolveHandleToDid(ctx context.Context, handle str
...
+
+	if wkerr == nil {
+		return wkres, nil
+	}
+
+	return "", errors.Join(fmt.Errorf("no did record found for handle %q", handle), dnserr, wkerr)
+}
+
+func (dr *ProdHandleResolver) resolveWellKnown(ctx context.Context, handle string) (string, error) {
 	c := http.DefaultClient
```
ericvolp12, line 123: Can we use an instrumented client here?
```go
client := http.Client{
	Transport: otelhttp.NewTransport(http.DefaultTransport),
}
```

ericvolp12 (approved):
LGTM but we could plug in otel here if we want.

PR #243, opened 2023-07-23
Use record timestamps for record timestamp tracker in sonar

ericvolp12:


sonar/sonar.go
```diff
@@ -183,14 +185,15 @@ func (s *Sonar) HandleRepoCommit(ctx context.Context, evt *comatproto.SyncSubscr
 	}
 
 	// Parse time from the event time string
-	t, err := time.Parse(time.RFC3339, evt.Time)
+	evtCreatedAt, err := time.Parse(time.RFC3339, evt.Time)
```
whyrusleeping, line 188: does RFC3339 work here? I wrote `util.ParseTimestamp` to deal with the large variety of different formats ive seen on records

whyrusleeping: Oh, I see you know about that. carry on

ericvolp12: Oh I started using a better thing like here: https://github.com/ericvolp12/bsky-experiments/blob/main/pkg/consumer/consumer.go#L256

https://github.com/araddon/dateparse/ - Faster than what we're currently doing i.e. "shotgun parsing"

For Sonar I think it makes sense for me to update to use that library so I'll patch that today.

util/time.go
```diff
@@ -9,6 +9,8 @@ const ISO8601 = "2006-01-02T15:04:05.000Z"
 
 const ISO8601_milli = "2006-01-02T15:04:05.000000Z"
 
+const ISO8601_milli_without_z = "2006-01-02T15:04:05.000000"
```
whyrusleeping, line 12: This isnt valid according to the spec, timestamp is required

PR #245, opened 2023-07-24
Batch cursor updates for PDSs

ericvolp12:
Bumping the cursor in the DB for every event has become a bottleneck on event throughput.

To ameliorate this, we can batch cursor updates and flush them to the DB every 10 seconds.

image


ericvolp12:
Shutdown handling ensures cursors flush on the way out (though they could fail to flush)
CleanShot 2023-07-24 at 14 07 12

whyrusleeping (approved):

PR #247, opened 2023-07-25
implement pagination for profile search

whyrusleeping:
Also put did's in the index for posts, so we can implement 'search by user' later. We already have handle there, but it doesnt help if the user changed their handle ever


ericvolp12 (approved):
LGTM

Do we wanna fuzzy search text posts in this patch too?

bnewbold:
The other refactor hit this feature, I believe, so closing this PR.

PR #249, opened 2023-07-25
Remove headstore since new index makes it cheap to query the cardb

ericvolp12:
Updates to the headstore have become a new bottleneck and since we have an index on `usr, seq DESC` in the CarDB, we can get a repo head trivially and don't need the headstore anymore.

CleanShot 2023-07-25 at 15 24 13


ericvolp12:
After removal, 1.35k evt/sec throughput is easy:
CleanShot 2023-07-25 at 16 05 24

whyrusleeping (approved):

PR #251, opened 2023-07-26
Add flag to print invite tree handles

emilyliu7321:
Adds a flag `--print-handles` to print handles in an invite tree


whyrusleeping (approved):
LGTM, thanks @emilyliu7321 !

PR #253, opened 2023-07-27
Refactor Supercollider to have Reload and Fire modes for higher burst emission

ericvolp12:
This change refactors Supercollider to support two different modes: Reload and Fire.

In Reload mode, Supercollider generates a specified number of events as quickly as it can and writes the CBOR to a file. It uses in-memory SQLite for the repo manager and I've benched it at around 2.2k evt/sec for generation.

The static file includes creation of the repos and all the repo appends for lots of Posts. 2 Million events takes around 16 minutes to produce. We call this a loaded magazine.

In Fire mode, Supercollider will fire the magazine of events at any client that connects to the subscribe repos endpoint. Since we're reading from a file, we can go as fast as the socket can handle, though we also accept a flag to limit emission rate if we want to for any reason.

This method allows us to get significantly higher rates (tested at >10k evt/sec) of event emission for benchmarking a repo stream consumer like the BGS.

You need to reload between firing only if the consumer has durable repo state, if you blow away the DB on the consumer between runs, you don't need to reload since we can play the repos over from creation again without issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ bsky.auth
 
 # Sonar cursor file
 sonar_cursor.json
+out/
--- a/Makefile
+++ b/Makefile
@@ -90,3 +90,21 @@ run-dev-search: .env ## Runs search daemon for local dev
 .PHONY: sonar-up
 sonar-up: # Runs sonar docker container
 	docker compose -f cmd/sonar/docker-compose.yml up --build -d || docker-compose -f cmd/sonar/docker-compose.yml up --build -d
+
+.PHONY: sc-reload
+sc-reload: # Reloads supercollider
+	go run cmd/supercollider/main.go \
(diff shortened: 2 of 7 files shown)


whyrusleeping (approved):
LGTM, though you could use a cborgen.Deferred instead of decoding and re-encoding the objects to write out to the stream

ericvolp12:
> LGTM, though you could use a cborgen.Deferred instead of decoding and re-encoding the objects to write out to the stream

How would you recommend I do that? It would be neat if I could still count the number of events I've emitted so we can track the throughput. Looking at the [tests](https://github.com/whyrusleeping/cbor-gen/blob/master/testing/roundtrip_test.go#L93) that use Deferred it isn't super clear to me how it functions or how I'd make use of it here.

Issue #256, opened 2023-07-27
Please update the generated files in `api/`


imax9000:
Please run `make lexgen`. The files weren't updated since May 22, and in particular `langs` field is missing from `FeedPost`

PR #257, opened 2023-07-27
update lexgen, cborgen, for current Lexicons

bnewbold:
This is a simple/naive update, just running the commands and getting things to compile. Had to insert default values in some client request cases. This does not update *server-side* parsing, particularly for the BGS sync requests (new nested object).

I'm also not sure if there are currently lexgen things we are supposed to manually edit/override in the struct defs.

Closes: https://github.com/bluesky-social/indigo/issues/256


testing/utils.go
```diff
@@ -165,7 +165,7 @@ func (tp *TestPDS) RequestScraping(t *testing.T, b *TestBGS) {
 	t.Helper()
 
 	c := &xrpc.Client{Host: "http://" + b.Host()}
-	if err := atproto.SyncRequestCrawl(context.TODO(), c, tp.RawHost()); err != nil {
```
whyrusleeping, line 168: did the requestCrawl endpoint change? or is this just a weird artifact?

ericvolp12 (approved):
LGTM

testing/utils.go
ericvolp12: It's effectively the same but it supports cborgen now as well as json afaict.

ericvolp12: Or at least it uses an input struct instead of a `map[string]interface` for marshalling the request body (and that struct supports JSON hints as well as CBORGEN hints)

PR #258, opened 2023-07-30
Skip JWT middleware for `com.atproto.server.createSession` RPC in PDS

strideynet:
Hey there - I'm trying to build out an E2E test suite based on the PDS implementation included here.

It looks like `com.atproto.server.createSession` has the JWT Auth middleware running on it, which prevents me calling this RPC in order to log in. As far as I can tell, the handler has been implemented for this RPC correctly (`pds/Server.handleComAtprotoServerCreateSession`) - it's just not possible to call it.

### DIFF
--- a/pds/server.go
+++ b/pds/server.go
@@ -33,6 +33,7 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/whyrusleeping/go-did"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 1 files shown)


pds/server.go
```diff
@@ -33,7 +33,6 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/lestrrat-go/jwx/v2/jwt"
-	"github.com/whyrusleeping/go-did"
```
strideynet, line 36: Looks like my editor whipped this import out since it was being used in this file.

ericvolp12 (approved):
Seems reasonable to me, thanks!

PR #259, opened 2023-07-31
Add Querycheck: a service to check the plan and execution time of DB Queries

ericvolp12:
Querycheck provides a continuous, live-configured, query plan checking tool.

Queries can be added/updated/removed at-runtime to allow for use in investigation of DB performance issues in-vivo.

A `/metrics` endpoint exposes a large set of Prometheus metrics for all executed queries which can be scraped and graphed in tailored dashboards based on the needs and behavior of each query.


whyrusleeping (approved):
Nice, this seems really useful

PR #262, opened 2023-08-02
fakermaker: add support for invite code and domain suffix

bnewbold:
This gets the tool working with, eg, sandbox, or staging.

Tested with a small handful of domains.


ericvolp12 (approved):
LGTM

PR #263, opened 2023-08-02
palomar (search) iteration

bnewbold:
Larger refactors in this branch:

- [x] local docker dev env documented
- [x] specify mappings (schemas) for post and profile indices
- [x] transform raw records in to the index schemas
- [x] different doc _id syntax
- [x] skip read+deserialization of records other than profile and post, for efficiency
- [x] don't store records in database; database only used for firehose cursor state
- [x] switch to informal /xrpc/app.bsky.unspecced.search*Skeleton endpoints
- [x] return only skeleton responses (eg, AT-URI or DID lists)
- [x] handle non-success OpenSearch responses as errors
- [x] auto-create indices with schema when in indexing mode (not READONLY) (with `go:embed` schemas)
- [x] switch logging to `log/slog`, including echo integration
- [x] use `atproto/identity` package for identity caching and handling, not `User` database record
- [x] merged in backfill worker code
- [x] use `analysis-icu` plugin for (hopefully) better internationalized search
- [x] special typeahead indexing and query parameter
- [x] basic/simple query string parsing, which should be safe, supports quoted phrases, and `from:` filtering

This branch includes a couple small commits to SDK code, which i've cherry-picked out as separate PRs for easier review.

See also Lexicon PR in atproto repo: https://github.com/bluesky-social/atproto/pull/1594

This is not compatible with the previous version of `palomar` at the HTTP API, opensearch index, or database schema levels. The config vars should be backwards compatible. The operational plan for staging and prod is to deploy this as an entirely new environment (eg, "prod2", "staging2"), get everything backfilled, and then flip over the AppView and then client app to use the lexicons/endpoints instead of the older version.

----

I think this is ready for review, merge, and deploy to staging. Some things to check before prod:

- [ ] compare index size and performance to existing version/schema
- [ ] real-world testing of profile typeahead (eg, do we need fuzzy?)
- [ ] real-world search relevancy checks
- [ ] real-world CJK text analysis checks (https://github.com/bluesky-social/indigo/issues/302)

Out of scope for this PR:

- [ ] deal with `created_at` timestamp not being reliable, by adding a `sort_at` hybrid field, for future "sort by date"
- [ ] instrumentation and metrics (Jaz to implement on top of this branch)
- [ ] better bulk indexing performance, especially during backfill: disable refresh during backfill? longer refresh window? bulk (batch) indexing would be best
- [x] integrate a better identity service/cache; current is probably Ok in context of backfill. or perhaps just bump the cache size to ~50k or ~100k identities in prod?


cmd/palomar/main.go
```diff
@@ -138,12 +137,27 @@ var runCmd = &cli.Command{
...
+		// TODO: replace this with "bingo" resolver
+		base := identity.BaseDirectory{
+			PLCURL: cctx.String("atp-plc-host"),
+			HTTPClient: http.Client{
+				Timeout: time.Second * 15,
+			},
+			TryAuthoritativeDNS:   true,
+			SkipDNSDomainSuffixes: []string{".bsky.social"},
+		}
+		dir := identity.NewCacheDirectory(&base, 50000, time.Hour*24, time.Minute*2)
```
ericvolp12, line 149: We could probably bump this to like 200k or smth considering we've got 1M users now

search/firehose.go
```diff
@@ -50,7 +53,15 @@ func (s *Server) RunIndexer(ctx context.Context) error {
...
 	d := websocket.DefaultDialer
-	con, _, err := d.Dial(fmt.Sprintf("%s/xrpc/com.atproto.sync.subscribeRepos?cursor=%d", s.bgshost, cur), http.Header{})
+	u, err := url.Parse(s.bgshost)
+	if err != nil {
+		return fmt.Errorf("invalid bgshost URI: %w", err)
+	}
+	u.Path = "xrpc/com.atproto.sync.subscribeRepos"
+	if cur != 0 {
+		u.RawPath = fmt.Sprintf("cursor=%d", cur)
+	}
```
ericvolp12, line 63: A `cursor` of 0 asks for the farthest back event that the host is willing to provide. Maybe we want to default the value to `-1` as s sentinel so we can allow for an explicit 0 cursor?

Doesn't need to be blocking but just a thing to think about.

search/handlers.go
```diff
@@ -2,78 +2,100 @@ package search
...
-
-	q := strings.TrimSpace(e.QueryParam("q"))
-	if q == "" {
-		return e.JSON(400, map[string]any{
-			"error": "must pass non-empty search query",
-		})
-	}
-
+func parseCursorLimit(e echo.Context) (int, int, error) {
 	offset := 0
```
ericvolp12, line 29: Is this a cursor or an offset?

search/handlers.go
```diff
@@ -83,10 +105,89 @@ func (s *Server) handleSearchRequestProfiles(e echo.Context) error {
...
+	resp, err := DoSearchPosts(ctx, s.dir, s.escli, s.postIndex, q, offset, size)
+	if err != nil {
+		return nil, err
+	}
+
+	posts := []syntax.ATURI{}
+	for _, r := range resp.Hits.Hits {
+		var doc PostDoc
+		if err := json.Unmarshal(r.Source, &doc); err != nil {
+			return nil, fmt.Errorf("decoding post doc from search response: %w", err)
```
ericvolp12, line 136: Do we want this to fail the entire query or just log it with some context and move to the next result?

search/handlers.go
```diff
@@ -83,10 +105,89 @@ func (s *Server) handleSearchRequestProfiles(e echo.Context) error {
...
+		}
+
+		did, err := syntax.ParseDID(doc.DID)
+		if err != nil {
+			return nil, fmt.Errorf("invalid DID in indexed document: %w", err)
```
ericvolp12, line 141: Same as above

search/handlers.go
```diff
@@ -83,10 +105,89 @@ func (s *Server) handleSearchRequestProfiles(e echo.Context) error {
...
+		}
+
+		posts = append(posts, syntax.ATURI(fmt.Sprintf("at://%s/app.bsky.feed.post/%s", did, doc.RecordRkey)))
+	}
+
+	out := SearchPostsSkeletonResp{Posts: posts}
+	if len(posts) == size && (offset+size) < 10000 {
+		out.Cursor = fmt.Sprintf("%d", offset+size)
+	}
+	fmt.Println(resp.Hits.Total)
```
ericvolp12, line 151: Is this supposed to be here?

search/handlers.go
```diff
@@ -83,10 +105,89 @@ func (s *Server) handleSearchRequestProfiles(e echo.Context) error {
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	actors := []syntax.DID{}
+	for _, r := range resp.Hits.Hits {
+		var doc ProfileDoc
+		if err := json.Unmarshal(r.Source, &doc); err != nil {
+			return nil, fmt.Errorf("decoding profile doc from search response: %w", err)
```
ericvolp12, line 174: Similarly, do we want to error here or log it and keep going?

search/query.go
```diff
@@ -44,60 +47,150 @@ type PostSearchResult struct {
 	Post any        `json:"post"`
 }
 
-func doSearchPosts(ctx context.Context, escli *es.Client, q string, offset int, size int) (*EsSearchResponse, error) {
+func checkParams(offset, size int) error {
+	if offset+size > 5000 || size > 1000 || offset > 1000 || offset < 0 || size < 0 {
```
ericvolp12, line 51: Are these a bit redundant? `offset+size` can't be > `2000` with the explicit size and offset rules, right?

search/query.go
```diff
@@ -44,60 +47,150 @@ type PostSearchResult struct {
...
+
+func DoSearchPosts(ctx context.Context, dir identity.Directory, escli *es.Client, index, q string, offset, size int) (*EsSearchResponse, error) {
+	if err := checkParams(offset, size); err != nil {
+		return nil, err
+	}
+	queryStr, filters := ParseQuery(ctx, dir, q)
+	basic := map[string]interface{}{
+		"simple_query_string": map[string]interface{}{
+			"query":            queryStr,
+			"fields":           []string{"everything"},
```
ericvolp12, line 65: I wonder how this is going to impact searching for words that people have in their handles

search/query.go
```diff
@@ -44,60 +47,150 @@ type PostSearchResult struct {
...
 
 func doSearch(ctx context.Context, escli *es.Client, index string, query interface{}) (*EsSearchResponse, error) {
 	var buf bytes.Buffer
 	if err := json.NewEncoder(&buf).Encode(query); err != nil {
-		log.Fatalf("Error encoding query: %s", err)
+		return nil, fmt.Errorf("error encoding query: %s", err)
+	}
+
+	// re-marshal query as string for logging
+	if true {
```
ericvolp12, line 172: Why is this in an `if true`?

search/server.go
```diff
@@ -87,19 +72,22 @@ func NewServer(db *gorm.DB, escli *es.Client, plcHost, pdsHost, bgsHost string)
...
+		bgshost:      config.BGSHost, // NOTE: the original URL, not 'bgshttp'
+		bgsxrpc:      bgsxrpc,
+		dir:          dir,
+		logger:       logger,
 	}
 
 	bfstore := backfill.NewGormstore(db)
 	opts := backfill.DefaultBackfillOptions()
+	opts.ParallelRecordCreates = 20
+	opts.SyncRequestsPerSecond = 8
```
ericvolp12, line 89: Should these be runtime configurable?

ericvolp12 (approved):
Overall looks solid, have a few comments on how we want to handle different error cases and some dead code bits etc.

cmd/palomar/main.go
bnewbold: A lot fewer accounts will have emitted an event within the TLL window though. Probably should be configurable but i'll set to 200k for now, this is an indexing service.

search/firehose.go
bnewbold: We'll probably want to be able to configure this, the same way Kafka lets you default a consumer group to "start at current offset", "start at earliest available", "start at a configured time delta", etc.

For now it is in the "zero means empty/nil" situation. I'm going to leave as-is, which I think is a sane default with the backfill stuff now handling broader enumeration of backfill work.

search/handlers.go
bnewbold: This is a pretty bad error which should "never" happen, so I think we want to fail-and-bail on this situation.

An example where this would happen is if there was a config mistake where a query worker got pointed at the wrong elasticsearch index (either post instead of profile, or an older schema version). I think in that kind of situation we want to fail loudly.

search/handlers.go
bnewbold: whoops!

search/handlers.go
bnewbold: same as above: this is a bad and very unexpected situation

search/handlers.go
bnewbold: same as above: this is a bad and very unexpected situation

search/handlers.go
```diff
@@ -83,10 +105,89 @@ func (s *Server) handleSearchRequestProfiles(e echo.Context) error {
...
+		}
+
+		did, err := syntax.ParseDID(doc.DID)
+		if err != nil {
+			return nil, fmt.Errorf("invalid DID in indexed document: %w", err)
```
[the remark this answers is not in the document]
bnewbold: same as above

search/query.go
bnewbold: yeah, it is redundant. I guess should decide whether we want this check here or in the other place. I like having it just before query to have confidence it is always applied; but also verifying when HTTP requests are parsed is idiomatic. hrm.

search/query.go
bnewbold: For the post schema, we don't include handles in the document, or in "everything", so should not impact.

search/query.go
ericvolp12: The check is fine you're just checking x+y > 5000 || x > 1000 || y > 1000 which is redundant since x+y will only be > 5000 if one of them is > 1000.

search/query.go
bnewbold: I can't remember why now... this code block is weird/annoying, will clean it up

search/server.go
bnewbold: yup, added

search/query.go
bnewbold: oh, hah, yeah. updated with better and no-longer-redundant sanity check:

`offset+size > 10000 || size > 250 || offset > 10000 || offset < 0 || size < 0`

bnewbold: (well, I guess it is still slightly redundant, but :shrug:)

bnewbold:
Addressed a bunch of these. The main outstanding thing is failing hard on some of the query error cases. I *think* that is the right thing to do, but could be wrong and we could get bitten.

I'll probably merge this later/tomorrow after thinking a bit more.

search/firehose.go
ericvolp12: Just clarifying, this means if you provide no cursor we'll start from as far back as the BGS/PDS will allow us, not from live.

bnewbold: Looking at this again, I guess I had a typo and it should be `RawQuery` not `RawPath`.

But if we don't provide a `cursor` HTTP query param at all, shouldn't the PDS (or BGS) give us the current stream, not start at the beginning? The intent of this snippet of code is to not set the parameter at all, instead of defaulting to `cursor=0` (which would scroll back to oldest available).

ericvolp12: Oh oops yeah you're right, I read that wrong

PR #264, opened 2023-08-02
Only ping a consumer if we haven't written _anything_ in a while

ericvolp12:
Consumer timeout checking is great but if we're playing back a TON of events, the ping ends up deep in a backlog and out 5 second timeout on the ping doesn't make sense so let's keep track of the last time we successfully wrote to the consumer and only try to ping if it's been 30 seconds since our last write.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -474,11 +474,12 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
 			select {
 			case <-ticker.C:
 				lastWriteLk.RLock()
-				if time.Since(lastWrite) < 30*time.Second {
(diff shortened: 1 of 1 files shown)


bgs/bgs.go
```diff
@@ -460,6 +460,9 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
 		return fmt.Errorf("upgrading websocket: %w", err)
 	}
 
+	lastWriteLk := sync.RWMutex{}
```
whyrusleeping, line 463: i'd just use a normal Mutex here, you only go faster with an RWMutex if reads are far more common than writes

whyrusleeping (approved):
LGTM

PR #268, opened 2023-08-10
Remove Consumer timeout stuff since it's causing more problems than it solves

ericvolp12:
Consumer timeout polling only makes sense when there are lots of consumers and no write throughput on the BGS which isn't really a usecase we're going to see outside of contrived test environments atm.


ericvolp12:
This is actually fine, I'm just dumb

PR #270, opened 2023-08-11
atproto/crypto: in-repo SDK for atproto cryptography

bnewbold:
This borrows some code from https://github.com/whyrusleeping/go-did, but has a refactored API, exposes fewer internals, and has a couple differences:

- P-256 "low-S" signatures are enforced (when signing and verifying), in addition to just K-256. This fixes an interop test failure vs the typescript repo
- P-256 binary private key serialization is the normal compact representation (32 bytes), instead of using `crypto/x509` with a wrapping format
- docs and examples

Looking for API-level feedback, and thoughts on the safety/risk of the "low-S" hack code. We could consider asking external folks for API feedback as well, or just iterate on the API later.

If this looks good, I would imagine merging as-is, and then refactoring the code base to use this code instead of `go-did` as a separate PR. One operational concern i'd want to address is the private key serialization format used in our infra: do we use P-256 private keys, or K-256 private keys, or actually no private keys at all yet?

Can preview godoc HTML output here: https://static.bnewbold.net/tmp/atproto_crypto_godoc.html

cc: @whyrusleeping @ericvolp12


atproto/crypto/keys.go
```diff
@@ -0,0 +1,457 @@
...
+	if k.keyType != other.keyType {
+		return false
+	}
+	switch k.keyType {
+	case P256:
+		return k.privP256.Equal(other.privP256)
+	case K256:
+		return k.privK256.Equal(other.privK256)
+	default:
+		panic("unexpected crypto KeyType")
```
ericvolp12, line 112: Do we need to `panic` here? Can we have this function return an error? It might be helpful to return an error for mismatched key types too.

atproto/crypto/keys.go
```diff
@@ -0,0 +1,457 @@
...
+// SHA-256 is the hash algorithm used, as specified by atproto. Signing digests is the norm for ECDSA, and required by some backend implementations. This method does not "double hash", it simply has name which clarifies that hashing is happening.
+//
+// Calling code is responsible for any string encoding of signatures (eg, hex or base64). Both P-256 and K-256 signatures are 64 bytes long.
+//
+// NIST ECDSA signatures can have a "malleability" issue, meaning that there are multiple valid signatures for the same content with the same signing key. This method always returns a "low-S" signature, as required by atproto.
+func (k *PrivateKey) HashAndSign(content []byte) ([]byte, error) {
+	hash := sha256.Sum256(content)
+	switch k.keyType {
+	case P256:
+		r, s, err := ecdsa.Sign(rand.Reader, k.privP256, hash[:])
```
ericvolp12, line 167: What's the function of using a `[:]` here? I'm unfamiliar with what this does when no indexes are specified on either side, is it some kind of shorthand to copy the slice?

atproto/crypto/keys.go
```diff
@@ -0,0 +1,457 @@
...
+	if k.keyType != other.keyType {
+		return false
+	}
+	switch k.keyType {
+	case P256:
+		return k.pubP256.Equal(other.pubP256)
+	case K256:
+		return k.pubK256.Equal(other.pubK256)
+	default:
+		panic("unexpected crypto KeyType")
```
ericvolp12, line 194: Again, do we need to panic here or can we return an error from this function?

I know it's not as "clean" from a equality operator standpoint but it may be more helpful.

ericvolp12 (changes requested):
This interface and the GoDoc look pretty good to me.

I'm not a cryptographer but I have strong opinions about never writing a `panic` in Go code, especially in a library, so I've requested you remove those and make those functions return errors even if it's less ergonomic.

There was a spot where we might wanna bounds check too.

Overall, looks great and I'm happy to have nice ergonomic functions to work with the strange collection of key representations we've gotta deal with.

bnewbold:
@ericvolp12 thanks for the review!

An alternative API, which wouldn't be too hard to update to, would be to define an interface and have separate P256 and K256 implementations of that interface. That would remove a bunch of the `panic()` conditions by pushing things to the type system. Any thoughts on that?

I would probably name them `PublicKey`, `PrivateKey` (though `SecretKey` is also decent, and `PrivateKeyExportable`. The exportable-or-not being whether the key material is actually in memory, or something like a yubikey. Having overlapping interfaces makes it possible to wrap non-exportable keys in the future and use them with most other APIs/implementations, which seems good.

atproto/crypto/keys.go
bnewbold: `sha256.Sum256()` returns an array of fixed size `[32]byte`, while `ecdsa.Sign()` takes a slice `[]byte`. The `[:]` is a shorthand to do that type conversion ("take all the elements of the array as a slice").

It is a little tempting to use fixed size byte arrays in a couple places, like parsing keys from raw bytes, or verifying signatures, because K-256 and P-256 have the same size keys, and separately same size signatures, in bytes. But that locks the function signature in to those sizes, and we want to preserve the ability to add more curves in the future.

atproto/crypto/keys.go
```diff
@@ -0,0 +1,457 @@
...
+	if len(encoded) < 2 || encoded[0] != 'z' {
+		return nil, fmt.Errorf("crypto: not a multibase base58btc string")
+	}
+	data, err := base58.Decode(encoded[1:])
+	if err != nil {
+		return nil, fmt.Errorf("crypto: not a multibase base58btc string")
+	}
+	if data[0] == 0x80 && data[1] == 0x24 {
+		// multicodec p256-pub, code 0x1200, varint-encoded bytes: [0x80, 0x24]
+		return ParsePublicBytes(data[2:], P256)
```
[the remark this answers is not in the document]
bnewbold: good catch! added a bounds check just above

bnewbold:
@ericvolp12 I iterated on the API to have it use interfaces instead of a hacked enum struct.

This got rid of most of the panics. There are two remaining panics in the P-256 implementation. These are relatively safe because I call the same code block in all of construction code paths (which return `err`) to verify that the routine should not actually panic. This is kind of weird, and also resulted in returning `err` for `Public()` which wouldn't otherwise be necessary, so i'm willing to iterate again if we want to truely expunge the panics. I don't have particularly strong opinions and would follow your judgement, it just feels weird that serializing a pubkey to bytes, or any of the dependent string encodings than use the bytes, could return `err`.

A couple methods are probably going to be deprecated. I guess we could could just not include those here so the new interface is clean and stable? But they might also linger longer or be permanent. I feel comfortable yanking them in a couple months as long as they were marked as such from the begining.

bnewbold:
After sleeping on it, i'll probably ease the stuff for working with "legacy" DID document (weird multibase encoding, "uncompressed" byte formats, suite names) out of any interface definitions. Will keep the uncompressed helpers on the current public key types, but move the rest to the (forthcoming) identity package.

TODO to update this PR reflecting those changes.

bnewbold:
@ericvolp12 updated and ready for (final?) review

godoc preview: https://static.bnewbold.net/tmp/atproto_crypto_godoc2.html

ericvolp12:
> @ericvolp12 updated and ready for (final?) review
> 
> godoc preview: https://static.bnewbold.net/tmp/atproto_crypto_godoc2.html

New interface looks sensible to me, good job!

Only question I have is still around the use of `hash[:]` instead of just `hash` in a few functions.

Afaict it's a no-op: https://goplay.tools/snippet/Hvt9pv9jAAX

If that's the case, can we pull the `[:]` out?

ericvolp12 (approved):
Just the one note about using `hash[:]`, otherwise LGTM!

bnewbold:
@ericvolp12 I had a reply to a review comment above about it, the SHA-256 return is typed. If I try removing the `[:]` from any of those call points, I get a compiler error:

```
./k256.go:94:38: cannot use hash (variable of type [32]byte) as []byte value in argument to k.privK256.Sign
```

bnewbold:
Also just because it felt sneaky, to ensure you noticed, I replaced the "should be impossible" `panic` calls with `fmt.Fatal`. I'm not sure if that is actually any better.

atproto/crypto/keys.go
ericvolp12: Ah, okay hrmm, interesting

ericvolp12:
> Also just because it felt sneaky, to ensure you noticed, I replaced the "should be impossible" `panic` calls with `fmt.Fatal`. I'm not sure if that is actually any better.

I don't think it should be possible for this library to make your program exit

ericvolp12:
> @ericvolp12 I had a reply to a review comment above about it, the SHA-256 return is typed. If I try removing the `[:]` from any of those call points, I get a compiler error:
> 
> ```
> ./k256.go:94:38: cannot use hash (variable of type [32]byte) as []byte value in argument to k.privK256.Sign
> ```

Ah, okay yeah I missed that, interesting...

ericvolp12:
> > Also just because it felt sneaky, to ensure you noticed, I replaced the "should be impossible" `panic` calls with `fmt.Fatal`. I'm not sure if that is actually any better.
> 
> I don't think it should be possible for this library to make your program exit

To elaborate on this, afaik the general convention in Go is that "panics should never leave _your_ code" so if there is a panic somewhere, it should be caught by the library and handled appropriately, not bubbled up to the caller.

bnewbold:
Didn't actually *need* the duplicate key representations for PublicKey, so got rid of that duplication and overhead.

There is still the overhead (2x, I assume) for private keys. It seems less likely that folks would have millions of private keys hanging around in memory (compared to public keys).

I am not strongly against just adding error to basically all the return types here to avoid any future issues and leave wiggle room, even though it doesn't seem necessary for now.

ericvolp12:
I think having 2x overhead for private keys is fine. Caching a ton of private keys is a lot less common of a scenario than caching a ton of public keys imo.

ericvolp12 (approved):
LGTM

PR #271, opened 2023-08-11
correct file modes on reopen

whyrusleeping:


events/diskpersist.go
```diff
@@ -145,7 +148,7 @@ func (dp *DiskPersistence) resumeLog() error {
 		return dp.initLogFile()
 	}
 
-	fi, err := os.Open(filepath.Join(dp.primaryDir, lfr.Path))
+	fi, err := os.OpenFile(filepath.Join(dp.primaryDir, lfr.Path), os.O_RDWR, 0)
```
ericvolp12, line 151: LGTM but I'd like a comment above here for the `0` filemode, OpenFile doesn't use the file mode if `O_CREATE` isn't passed but it's kinda scary to see.

ericvolp12 (approved):

events/diskpersist.go
```diff
@@ -284,6 +287,7 @@ func (p *DiskPersistence) flushRoutine() {
 
 	for {
 		select {
+		case <-p.shutdown:
```
ericvolp12, line 290: Are we not flushing on shutdown here?

ericvolp12 (approved):
LGTM

PR #276, opened 2023-08-16
syntax SDK, for identifier string types

bnewbold:
Relatively simple package for working with identifier string syntax.

- [x] NSID
- [x] DID (just the generic syntax, as allowed in atproto)
- [x] Handle
- [ ] TID
- [x] ATURI
- [x] ATIdentifier
- [x] RecordKey
- [ ] ~Datetime~
- [ ] ~URI (generic)~
- [ ] ~CID (string, generic)~
- [ ] ~Language~


bnewbold:
I think this is at a good milestone and ready for review. I would add the other syntax (Datetime, generic URI, Language (BCP-47 code), and CID (as string, not an actual Link, otherwise use `go-cid`).

A key point of the API for these is that they are just strings and don't do any normalization. This is a different from things like the stdlib URL API, which parses out in to segments. The goal here is that it should be safe to round-trip with these types, byte-for-byte.

ericvolp12 (approved):
This looks good to me, one wishlist item I have would be to support TIDs in here somewhere too maybe? Not quite sure if that's a proper protocol spec though.

bnewbold:
Yup, will add TIDs in a follow-up. I'm not sure if i'll do the TID *generation* part in this package, but definitely all the serialization/conversation.

PR #277, opened 2023-08-17
Occasionally garbage collect old event files and references

ericvolp12:
This change sets a retention period for events in the BGS Disk Persister, adding a goroutine that runs once an hour to delete any references and files created before `time.Now() - retention_period`.

We're using the GORM auto-generated `created_at` field to find file creation date right now, we may desire to change that later if it is insufficient. Events younger than the retention period will be deleted using this mechanism, because the reference is tagged with is creation time instead of the timestamp of the last record it has stored, but we won't delete the current file ever.


events/diskpersist.go
```diff
@@ -72,14 +73,16 @@ type DiskPersistOptions struct {
...
 	WriteBufferSize int
+	Retention       time.Duration
 }
 
 func DefaultDiskPersistOptions() *DiskPersistOptions {
 	return &DiskPersistOptions{
-		EventsPerFile:   10000,
-		UIDCacheSize:    100000,
-		DIDCacheSize:    100000,
+		EventsPerFile:   20,
```
whyrusleeping, line 81: 20 seems wrong

ericvolp12: Yeah, good catch we tweaked it while making sure garbage collection worked, bumping it back to 20k

ericvolp12: err, 10k oops

events/diskpersist.go
```diff
@@ -265,71 +270,170 @@ const (
...
+		currentLogfile := dp.logfi.Name()
+		dp.lk.Unlock()
+
+		if filepath.Join(dp.primaryDir, r.Path) == currentLogfile {
+			// Don't delete the current log file
+			log.Info("skipping deletion of current log file")
+			continue
+		}
+
+		// Delete the ref in the database to prevent playback from finding it
```
whyrusleeping, line 403: I would swap the order of deleting the file and the ref. If we fail to delete the file and havent deleted the ref, then we can just try again later once we've rectified whatever issue was there.

though I guess if someone requested the events in the file after we deleted it but before we dropped the ref that would be weird for them.

ericvolp12: I want to delete the ref first because if we delete the file but fail to delete the ref, we could try to open it for playback and fail.

ericvolp12: We can clean up old files with a cron job if we need to

PR #280, opened 2023-08-18
track no longer referenced blocks, delete refs

whyrusleeping:
We won't need this until the sync changes that remove history land.

This adds tracking of which blocks are no longer referenced during processing of a new commit. We then delete the carstore references for those blocks when accepting that commit.
Work still to be done is figuring out how we want to compact old shard files.

It seems relatively straightforward to do the compaction as a post-process step that queries recently changed repos and checks if enough block references have been dropped to make it worth running a repo compaction. (in writing this, im thinking we might want a 'stale' flag on blockRefs that we set to true instead of deleting them).
When compacting, to make the read paths more efficient, we should leave the most recent several car slices alone and only compact the oldest N shards. For efficiency on the write path it is advantageous to not have individual files be too big, so that when reading data for making edits, we can read a small number of shards whole.

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -802,6 +802,10 @@ func BlockDiff(ctx context.Context, bs blockstore.Blockstore, oldroot cid.Cid, n
 	ctx, span := otel.Tracer("repo").Start(ctx, "BlockDiff")
 	defer span.End()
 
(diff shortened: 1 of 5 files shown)


ericvolp12 (approved):
LGTM

PR #288, opened 2023-08-24
update whyrusleeping/go-did dep (DID doc formatting)

bnewbold:


ericvolp12 (approved):
wow i cant believe u broke everything

PR #289, opened 2023-08-24
Auto-refreshing XRPC client

jazware:


lex/gen.go
```diff
@@ -576,6 +576,11 @@ func (s *TypeSchema) WriteRPC(w io.Writer, typename string) error {
 		return fmt.Errorf("can only generate RPC for Query or Procedure (got %s)", s.Type)
 	}
 
+	if fname != "ServerRefreshSession" {
+		pf("\tc.Mux.RLock()\n")
+		pf("\tdefer c.Mux.RUnlock()\n")
+	}
```
bnewbold, line 582: I don't love this. Really feels like any locking should happen in the client, and only when needed (aka, when actually updating auth info). Should be possible to do concurrent requests from the same client.

atproto/syntax/aturi.go
```diff
@@ -103,7 +103,7 @@ func (n ATURI) Normalize() ATURI {
 }
 
 func (a ATURI) MarshalText() ([]byte, error) {
-	return []byte(a.String()), nil
+	return []byte(string(a)), nil
```
bnewbold, line 106: what is the motivation for this diff? does the compiler not optimize this already?

util/xrpcutil/xrpcutil.go
```diff
@@ -0,0 +1,101 @@
+package xrpcutil
```
bnewbold, line 1: can we put this in the `xrpc` package as `AutoClient` or something? Feels weird over here in `util`

atproto/syntax/aturi.go
ericvolp12: I was getting compiler errors from it for some reason?

util/xrpcutil/xrpcutil.go
ericvolp12: It can be a subpackage of the `xrpc` package but it can't be in the main `xrpc` package because of the dependency loop.

lex/gen.go
ericvolp12: This is a readlock, you can do concurrent requests, it just makes sure you're not trying to make a request while the client object is being modified.

atproto/syntax/aturi.go
ericvolp12: CleanShot 2023-08-24 at 17 06 26

ericvolp12: Oh, yeah looks like there isn't a `String()` method on the `ATURI` type rn.

PR #290, opened 2023-08-25
identity SDK

bnewbold:
This is ready for feedback and review. There are a couple TODOs i'd be curious to discuss, and I suspect there are some ptr/non-ptr style polish to the public API.

Includes somewhat polished implementations of the basic DID and handle resolution routines, plus a new Catalog abstraction (interface), with three implementations:

- BasicCatalog: simply calls resolution routines on every single lookup request
- CacheCatalog: wraps another implementation (eg, BasicCatalog), plus an in-memory lookup cache, using `map`, `sync.RWLock`, and per-record TTLs
- MockCatalog: for testing, allowing simple injection

A missing thing we'd want before shipping in prod is better observability, probably via an observability layer that does OTEL metrics and `slog` logging.

We are finding that identity lookups are a pretty performance critical part of our service implementations, and we are likely to iterate on different backends and policies. Hopefully this interface will work well for large and small projects.


atproto/identity/identity.go
```diff
@@ -0,0 +1,187 @@
...
+	return &cached
+}
+
+// Represents an atproto identity. Could be a regular user account, or a service account (eg, feed generator)
+type Identity struct {
+	DID syntax.DID
+
+	// Handle/DID mapping must be bi-directionally verified. If that fails, the Handle should be the special 'handle.invalid' value
+	// TODO: should we make this nullable, instead of 'handle.invalid'?
+	Handle syntax.Handle
```
ericvolp12, line 55: Not all Identities have Handles i.e. lots of FeedGenerators have DID docs at their `https://{feed_host}/.well-known/did.json` and no AKA's defined in them. Do we only want this package to support Identities with handles?

atproto/identity/handle.go
```diff
@@ -0,0 +1,101 @@
...
+	// TODO: could pull a client or transport from context?
+	c := http.DefaultClient
+
+	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s/.well-known/atproto-did", handle), nil)
+	if err != nil {
+		return "", err
+	}
+	req = req.WithContext(ctx)
+
+	resp, err := c.Do(req)
```
ericvolp12, line 56: We probably want some kind of rate limiting logic but I'm not sure if we want it in this function or if the caller should deal with it...

atproto/identity/did.go
```diff
@@ -0,0 +1,108 @@
...
+	return &doc, nil
+}
+
+// plcURL should have URL method, hostname, and optional port; it should not have a path or trailing slash
+func ResolveDIDPLC(ctx context.Context, plcURL string, did syntax.DID) (*DIDDocument, error) {
+	if did.Method() != "plc" {
+		return nil, fmt.Errorf("expected a did:plc, got: %s", did)
+	}
+
+	resp, err := http.Get(plcURL + "/" + did.String())
```
ericvolp12, line 92: Maybe want a rate limiter here too

atproto/identity/cache_catalog.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+type CacheCatalog struct {
+	HitTTL        time.Duration
+	ErrTTL        time.Duration
+	Inner         Catalog
+	mutex         sync.RWMutex
+	handleCache   map[syntax.Handle]HandleEntry
+	identityCache map[syntax.DID]IdentityEntry
+}
```
ericvolp12, line 19: Do we wanna use https://github.com/hashicorp/golang-lru for these caches to limit their sizes and help with eviction?

ericvolp12 (approved):
I've got a few hangups on Handles being "required" for identities given there are cases where it isn't sensible i.e. for a Feed Generator's DID identity.

Also think we should slap some configurable common-sense rate limiters for some of the requests that this package can create to avoid getting WAF'd by services and/or accidentally DDOSing someone. If the caller were required to rate limit, they couldn't make use of cached entries effectively so we should probably use go's `rate.Limit` and stick it right before our web and DNS requests in the reference implementations.

atproto/identity/identity.go
bnewbold: This package should support those just fine; they simply have an "invalid" handle. If you never try to use the handle it isn't a problem.

As the comment says, we could have this field be nullable, which would be a bit more efficient (in memory) for cases with no handle. But it is kind of weird to have it be both nullable *and* there exist the special invalid handle. FWIW, I was against the special invalid handle existing and thought that we should have updated all the lexicons to have nullable handles instead, but we didn't want to break things, and having a specific value fails more gracefully in some situations. I could go either way on this: if there is a performance concern i'm fine having it be nullable.

A tangential corner-case with service accounts vs. regular PDS accounts is that they are not mutually exclusive: you can have a did:web feed generator which also has a repo (PDS location and public signing key).

atproto/identity/cache_catalog.go
bnewbold: I was thinking about that. I kind of want to keep this package more minimal w/r/t deps and have sub-packages or something for fancier implementations. But an LRU is pretty simple/common and we should probably have a polished and really robust solution to caching, as it is pretty load-bearing for identity.

bnewbold:
Thanks for the review! I've got some ideas for a small iteration on this: maybe move all the `Resolve*` routines on to the `BasicCatalog` struct, so that it can have more configuration (authoritative DNS, custom HTTP client/dialer, etc).

atproto/identity/cache_catalog.go
```diff
@@ -0,0 +1,209 @@
...
+func NewCacheCatalog(inner Catalog) CacheCatalog {
+	// TODO: these are kind of arbitrary default values...
+	hitTTL, err := time.ParseDuration("1h")
+	if err != nil {
+		panic(err)
+	}
+	errTTL, err := time.ParseDuration("2m")
+	if err != nil {
+		panic(err)
+	}
```
Jacob2161, line 44: It's safe with constant strings like this but still gives me pause to see a `panic` in a library. Maybe just change it to be`func NewCacheCatalog(inner Catalog, hitTTL, errTTL time.Duration) CacheCatalog` or whatever from the start?

atproto/identity/cache_catalog.go
```diff
@@ -0,0 +1,209 @@
...
+	c.handleCache[ident.Handle] = he
+	c.mutex.Unlock()
+	return &he, nil
+}
+
+func (c *CacheCatalog) ResolveHandle(ctx context.Context, h syntax.Handle) (syntax.DID, error) {
+	var err error
+	var entry *HandleEntry
+	c.mutex.RLock()
+	eObj, ok := c.handleCache[h]
```
Jacob2161, line 110: nit: i dislike "obj" as much as "var" on a variable name. Just tell me what kind of obj I'm looking at! :rofl:

atproto/identity/cache_catalog.go
```diff
@@ -0,0 +1,209 @@
...
+		Err:      err,
+	}
+	var he *HandleEntry
+	if err == nil && !ident.Handle.IsInvalidHandle() {
+		he = &HandleEntry{
+			Updated: time.Now(),
+			DID:     did,
+			Err:     nil,
+		}
+	}
```
Jacob2161, line 144: Slightly confusing/worrying due to the error not being handled directly after the call, seems to make sense but could use a comment.

atproto/identity/handle.go
```diff
@@ -0,0 +1,101 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+)
+
+// Does not cross-verify, just does the handle resolution step.
+func ResolveHandleDNS(ctx context.Context, handle syntax.Handle) (syntax.DID, error) {
+	// TODO: timeout
+	// TODO: mechanism to control NDS better; context? separate method?
```
Jacob2161, line 18: [the mikeg DNS package](https://github.com/miekg/dns) is a godsend here, the `net` DNS stuff is very basic

atproto/identity/handle.go
```diff
@@ -0,0 +1,101 @@
...
+func ResolveHandleWellKnown(ctx context.Context, handle syntax.Handle) (syntax.DID, error) {
+	// TODO: timeout
+	// TODO: could pull a client or transport from context?
+	c := http.DefaultClient
+
+	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s/.well-known/atproto-did", handle), nil)
+	if err != nil {
+		return "", err
+	}
+	req = req.WithContext(ctx)
```
Jacob2161, line 54: nit: there's `http.NewRequestWithContext` which is nice because then you don't think "oh, they forgot to use the context" above

atproto/identity/handle.go
```diff
@@ -0,0 +1,101 @@
...
+			return did, nil
+		}
+	}
+	return "", ErrHandleNotFound
+}
+
+func ResolveHandleWellKnown(ctx context.Context, handle syntax.Handle) (syntax.DID, error) {
+	// TODO: timeout
+	// TODO: could pull a client or transport from context?
+	c := http.DefaultClient
```
Jacob2161, line 48: I (try to) never use DefaultClient because there's no telling who else might be using it

atproto/identity/handle.go
```diff
@@ -0,0 +1,101 @@
...
+		// look for NXDOMAIN
+		var dnsErr *net.DNSError
+		if errors.As(err, &dnsErr) {
+			if dnsErr.IsNotFound {
+				return "", ErrHandleNotFound
+			}
+		}
+		return "", fmt.Errorf("failed to resolve handle (%s) through HTTP well-known route: %s", handle, err)
+	}
+	if resp.StatusCode != 200 {
```
Jacob2161, line 67:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

atproto/identity/did.go
```diff
@@ -0,0 +1,108 @@
...
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		if dnsErr.IsNotFound {
+			return nil, ErrDIDNotFound
+		}
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed HTTP fetch of did:web well-known document: %w", err)
+	}
+	if resp.StatusCode == 404 {
```
Jacob2161, line 72:
```suggestion
	if resp.StatusCode == http.StatusNotFound {
```

atproto/identity/did.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, ErrDIDNotFound
+	}
+	if resp.StatusCode != 200 {
```
Jacob2161, line 75:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

atproto/identity/did.go
```diff
@@ -0,0 +1,108 @@
...
+func ResolveDIDPLC(ctx context.Context, plcURL string, did syntax.DID) (*DIDDocument, error) {
+	if did.Method() != "plc" {
+		return nil, fmt.Errorf("expected a did:plc, got: %s", did)
+	}
+
+	resp, err := http.Get(plcURL + "/" + did.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed did:plc directory resolution: %w", err)
+	}
+	if resp.StatusCode == 404 {
```
Jacob2161, line 96:
```suggestion
	if resp.StatusCode == http.StatusNotFound {
```

Jacob2161 (approved):
Added some minor notes but looks really great :+1: 

Running with a bunch of goroutines with the race detector would be a good idea but the all locking looked tight so wouldn't expect issues.

atproto/identity/cache_catalog.go
bnewbold: strong agree, keep me disciplined!

atproto/identity/cache_catalog.go
bnewbold: good call, this is a footgun. for all the "is the error nil" checks, I flipped to `nil == err` (so it is more visually distinct from the usual `err != nil`), and added a comment, and added a longer comment in this particular "capture err but don't return it directly" situation.

atproto/identity/cache_catalog.go
bnewbold: Cross reference about pain of doing static error handling: https://github.com/golang/go/issues/54297

Replaced with static values in nanoseconds, which I verified were equal to the parsed values.

ericvolp12: Why not just do `time.Hour * 1` and `time.Minute * 2`?

bnewbold: even better!

atproto/identity/identity.go
bnewbold: we discussed and decided current API is fine

atproto/identity/handle.go
bnewbold: we discussed and decided not to do client-side rate limits

bnewbold:
Pushed an iteration which I think is ready to merge, with some possible future work. Changes:

- refactored all the resolve methods on to `BaseDirectory`, where they have access to config
- refactored `BaseDirectory` to be configurable, but to work fine in "zero" status ("uninitialized"), which is distinct from default
- switched cache implementation to use LRU (from the popular hashicorp package)
- added a Purge interface to bust the cache if needed
- added optional authoritative DNS lookup for handles in `BaseDirectory` implementation. if an initial DNS TXT resolution is "not found" (NXDOMAIN), and the config flag is set, it will do a single NS lookup and re-try the resolution under the first nameserver (if any). this is very narrowly scoped to working with new handles, not changing an existing handle
- added some logging (including timing) with `log/slog`

Things discussed but not in this version:

- [ ] handling HTTP 429 with longer backoff, for the default config. I think this needs the hashicorp retry HTTP client, and didn't want to pull in that dependency for now
- [ ] adding some tests with concurrency, to trigger racedetector. not as worried about this now with LRU handling locking
- [ ] adding metrics with promauto. read the docs and it seem like it can panic if config is weird/incorrect, which is not good. I think we should do metrics+tracing in a separate layer, not baked in, unfortunately. the debug-level logs have some timing info which might be sufficiently

atproto/identity/handle.go
bnewbold: I looked at this but didn't want to pull in a dependency in this initial implementation. I did: 

- expose the `net.Resolver`, which can be used to point to a specific DNS server easily, or be swapped out with a not-too-crazy interface implementation wrapping something else
- implemented a basic/partial authoritative lookup with the stdlib net DNS (not as complete as your implementation)

and open to iterating on the API/implementation.

PR #292, opened 2023-08-26
atproto/syntax: hotfix for ATURI String()

bnewbold:
Somehow had the `String()` method defined in the test file instead of `aturi.go` by mistake.


ericvolp12:
I've also got this change in https://github.com/bluesky-social/indigo/pull/289 along with a change to rename the receivers in this file to use `a` consistently instead of a mix of `a` and `n`.

PR #294, opened 2023-08-31
Add backfilling to search backed by GORM

ericvolp12:
Search will need to backfill regularly when rebuilding indexes.

I've adopted a backfill pattern I created a few weeks ago into `indigo` and updated Search's firehose consuming to adopt the backfilling pattern.


backfill/backfill.go
```diff
@@ -0,0 +1,397 @@
...
+	// Number of backfills to process in parallel
+	ParallelBackfills int
+	// Number of records to process in parallel for each backfill
+	ParallelRecordCreates int
+	// Prefix match for records to backfill i.e. app.bsky.feed.app/
+	// If empty, all records will be backfilled
+	NSIDFilter   string
+	CheckoutPath string
+
+	Logger      *zap.SugaredLogger
```
bnewbold, line 64: I don't love proliferating logging libraries. We have go-log currently, and I hoped we could transition to `log/slog`. If we want to use `zap` instead we should probably commit and use it in most places.

bnewbold:
I haven't reviewed in depth, but i'm pretty darn excited about having a generic backfill mechanism like this! Great work.

backfill/backfill.go
ericvolp12: The slog package definitely has a few rough edges i.e.:

```
slog.Info(fmt.Sprintf("processing backfill for %s", repoDid))
```

Instead of 
```
slog.Infof("processing backfill for %s", repoDid))
```

There might be some magic you can do to make this less ugly but I haven't seen it yet.

Refactored to use slog anyhow.

bnewbold: the pattern I would use with structured logging is:

```
slog.Info("processing repo backfill", "did", repoDid)
```

for the common case of errors, I think it is acceptable (?) to do this even without checking if the error is nil or not first:

```
slog.Info("doing a thing", "err", err)
```

ericvolp12: Slog requires you to wrap in a `slog.Attr` but I don't think there's a `slog.Attr` for errors?

i.e.

```
slog.Info("do a thing", slog.String("my_key", "my_value"))
```

ericvolp12: This was one of the errors that prompted me to do what I did there:
CleanShot 2023-08-31 at 15 31 20

bnewbold: I think you can use either API: with attr or without. The with attr is a bit more performant; probably a bunch more in relative terms, but very little impact in absolute terms for infrequent logging.

I used the direct style, and confirmed it worked, here: https://github.com/bluesky-social/indigo/blob/main/atproto/identity/handle.go#L137

ericvolp12: I might have missed something in that case and overcorrected though, it's possible that the key, value pairs are happy without an explicit `log.attr` wrapper but if I wanted to do a format string I needed to explicitly sprintf it.

ericvolp12: Okay I think that's a lot cleaner now

backfill/backfill.go
```diff
@@ -0,0 +1,399 @@
...
+	defer span.End()
+
+	start := time.Now()
+
+	repoDid := job.Repo()
+
+	log := slog.With("source", "backfiller_backfill_repo", "repo", repoDid)
+	log.Info(fmt.Sprintf("processing backfill for %s", repoDid))
+
+	var url = fmt.Sprintf("%s?did=%s", b.CheckoutPath, repoDid)
```
whyrusleeping, line 252: `url := fmt.Sprintf(...` ?

backfill/backfill.go
```diff
@@ -0,0 +1,399 @@
...
+		Transport: otelhttp.NewTransport(http.DefaultTransport),
+		Timeout:   120 * time.Second,
+	}
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		log.Error("failed to create request", "error", err)
+		return
+	}
+
+	req.Header.Set("Accept", "application/vnd.ipld.car")
```
whyrusleeping, line 265: all this feels like it wants to be abstracted out a little. i assume you have a non-lexicon 'fetch repo' endpoint youre hitting?

backfill/backfill.go
```diff
@@ -0,0 +1,399 @@
...
+		})
+	}()
+
+	// Consumer routines
+	for i := 0; i < numRoutines; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for item := range recordQueue {
+				recordCid, rec, err := r.GetRecord(ctx, item.recordPath)
```
whyrusleeping, line 351: TIL: Repo is threadsafe

whyrusleeping: you could also just load the record by CID directly instead of having to do a path traversal

whyrusleeping (approved):
a few comments, the code could be cleaned up a bit, but i wont let that block anything

backfill/backfill.go
ericvolp12: I changed to invoke the blockstore directly and GET the CID but I have to do a little typecast that's kinda annoying, should be cheaper though

PR #295, opened 2023-09-01
Update sync logic to deal with repov3 correctly

whyrusleeping:


indexer/indexer.go
```diff
@@ -194,6 +189,295 @@ func (ix *Indexer) handleRepoOp(ctx context.Context, evt *repomgr.RepoEvent, op
 	return nil
 }
 
+func (ix *Indexer) crawlAtUriRef(ctx context.Context, uri string) error {
```
whyrusleeping, line 192: i just moved all this code to the bottom of the file. I want to delete it, but its used by the PDS still so we should probably just extract it to something pds specific (or just rewrite the pds logic better)

repo/repo.go
```diff
@@ -253,18 +253,18 @@ func (r *Repo) Truncate() {
 }
 
 // creates and writes a new SignedCommit for this repo, with `prev` pointing to old value
-func (r *Repo) Commit(ctx context.Context, signer func(context.Context, string, []byte) ([]byte, error)) (cid.Cid, error) {
+func (r *Repo) Commit(ctx context.Context, signer func(context.Context, string, []byte) ([]byte, error)) (cid.Cid, string, error) {
```
whyrusleeping, line 256: we might be able to get away with removing the cid from the return value here

ericvolp12 (approved):
LGTM

whyrusleeping:
lovely net-negative PR

PR #299, opened 2023-09-07
implement first pass at compaction of carstore shards

whyrusleeping:


carstore/bs.go
```diff
@@ -930,49 +931,320 @@ func (cs *CarStore) Stat(ctx context.Context, usr models.Uid) ([]UserStat, error
...
+	cleanBlocks int
+}
+
+func (cb *compBucket) addShardStat(ss shardStat) {
+	cb.cleanBlocks += (ss.Total - ss.Dirty)
+	cb.shards = append(cb.shards, ss)
+}
+
+func (cb *compBucket) isFull() bool {
+	return cb.cleanBlocks > 50
```
ericvolp12, line 1043: Is this a magic number? If so can it have a comment explaining why it was chosen?

ericvolp12 (approved):
I have absolutely no idea what this does if I'm being honest but the test seems reasonable and it passes so I guess it's okay?

err.. I guess I know what it does in theory but don't know enough about how the carstore is implemented in memory and on disk to definitively say this does what it's supposed to do...

bgs/bgs.go
```diff
@@ -1029,3 +1031,27 @@ func (bgs *BGS) ReverseTakedown(ctx context.Context, did string) error {
...
+		}
+	}
+
+	return nil
+}
+
+func (bgs *BGS) runRepoCompactor(ctx context.Context) {
+	for range time.Tick(time.Hour) {
+	}
+}
```
ericvolp12, line 1057: Do we wanna pull this out for now?

ericvolp12 (approved):
New changes LGTM just wanna get rid of some dead code

ericvolp12 (approved):
LGTM

PR #304, opened 2023-09-13
identity: invalid/missing handle for LookupDID is not an error, just an invalid handle

bnewbold:
This is the documented behavior, but implementation was not working correctly for the base resolver.


ericvolp12 (approved):
lgtm

wondering if we should have a const for `handle.invalid` now tbh

PR #307, opened 2023-09-13
Metrics tracking indexer and crawl stats plus prepping for search metrics

ericvolp12:
Adding some more visibility to BGS crawler state so we can see how aggressively it's crawling.

I also aded a few more "verbose" comments because the state machine in the crawler `mainLoop` is a bit cryptic to reason about without reading the entire thing very thoroughly.

### DIFF
--- a/indexer/crawler.go
+++ b/indexer/crawler.go
@@ -83,26 +83,10 @@ func (c *CrawlDispatcher) mainLoop() {
 		select {
 		case actorToCrawl := <-c.ingest:
 			// TODO: max buffer size
-			c.maplk.Lock()
-
-			_, ok := c.inProgress[actorToCrawl.Uid]
-			if ok {
-				c.maplk.Unlock()
+			crawlJob, enqueued := c.enqueueJobForActor(actorToCrawl)
+			if !enqueued {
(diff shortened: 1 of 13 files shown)


ericvolp12:
Plugged in metrics for Search after rebasing on main with Bryan's changes.

Made a few tweaks as well where we were using the deprecated `ioutil.ReadAll` which just calls `io.ReadAll` under the hood since Go 1.16

indexer/crawler.go
```diff
@@ -64,114 +64,134 @@ type crawlWork struct {
...
+	crawlJob := &crawlWork{
+		act:        ai,
+		initScrape: true,
+	}
+	c.todo[ai.Uid] = crawlJob
+	return crawlJob
+}
+
+// dequeueJob removes a job from the todo list and adds it to the inProgress list
+func (c *CrawlDispatcher) dequeueJob(job *crawlWork) {
```
whyrusleeping, line 168: i wouldnt neccesarily call this dequeue, its more 'beginJob' or something like that

whyrusleeping (approved):
LGTM

PR #309, opened 2023-09-14
identity: support skipping DNS resolution for some domain suffixes

bnewbold:
And specifically, skip it by default for `.bsky.social` (we only do HTTP resolution). This reduces unnecessary DNS traffic, especially with higher concurrency/parallelism.


ericvolp12 (approved):
LGTM

PR #310, opened 2023-09-14
syntax: ATURI ergonomics

bnewbold:
This was something we discussed a couple weeks back: making access to sub-fields of ATURI more ergonomic.

This embraces the reality of "empty" values in golang and will return empty values instead of a `nil` pointer and error when the field doesn't exist, or if it fails to parse. I'm not really sure how I feel about it; feels vaguely unsafe? Maybe we need "IsEmpty()" helpers on all these, or something?

Still open to a `ATURIParts` or similarly named `struct` version of URI, with the various components as accessible fields; that could be in addition to this PR.

### DIFF
--- a/atproto/syntax/atidentifier.go
+++ b/atproto/syntax/atidentifier.go
@@ -24,12 +24,22 @@ func ParseAtIdentifier(raw string) (*AtIdentifier, error) {
 	return &AtIdentifier{Inner: handle}, nil
 }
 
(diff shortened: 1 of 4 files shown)


bnewbold:
(tests pass fine locally, I've retried a couple times and can't figure out why they are failing in CI)

ericvolp12 (approved):
LGTM

PR #318, opened 2023-09-18
Handle conflicts with handles

ericvolp12:
Newer DIDs with valid handles invalidate existing DIDs with the same handle

We check for handle validity when creating an External User which means this handle _should_ be more recently valid than whatever we might have in our database.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -1019,8 +1019,9 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
 			if err := s.db.Create(&u).Error; err != nil {
 				return nil, fmt.Errorf("failed to create user after handle conflict: %w", err)
 			}
+		} else {
+			return nil, fmt.Errorf("failed to create other pds user: %w", err)
 		}
-		return nil, fmt.Errorf("failed to create other pds user: %w", err)
 	}
 
 	// okay cool, its a user on a server we are peered with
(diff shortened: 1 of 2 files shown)


bgs/bgs.go
```diff
@@ -992,7 +993,33 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
...
+				if !errors.Is(err, gorm.ErrRecordNotFound) {
+					return nil, fmt.Errorf("failed to update outdated actorInfo's handle: %w", err)
+				}
+			}
+
+			// Create the new user
+			if err := s.db.Create(&u).Error; err != nil {
+				return nil, fmt.Errorf("failed to create user after handle conflict: %w", err)
+			}
+		}
```
whyrusleeping, line 1022: this will still return this error btw

ericvolp12: It's fine if we return an error here cause that means something went really wrong, right?

ericvolp12: That's attempting to create the user again after we've hopefully resolved the conflict

whyrusleeping: no, what im saying is if we succeed in creating the user here, we still return an error on line 1023

ericvolp12: Ohhh, sorry I was reading the wrong line of code, resolved

whyrusleeping (approved):

PR #321, opened 2023-09-18
do shard deletions in batches

whyrusleeping:
compaction on rapidly changing repos is slow and 99% of the time is spent in deleteShard. This speeds it up a little bit, from there we can look at what else gets in the way.

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -947,39 +947,54 @@ func (cs *CarStore) Stat(ctx context.Context, usr models.Uid) ([]UserStat, error
 }
 
 func (cs *CarStore) TakeDownRepo(ctx context.Context, user models.Uid) error {
-	var shards []CarShard
+	var shards []*CarShard
 	if err := cs.meta.Find(&shards, "usr = ?", user).Error; err != nil {
 		return err
 	}
 
-	for _, sh := range shards {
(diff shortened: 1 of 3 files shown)


carstore/bs.go
```diff
@@ -947,39 +947,54 @@ func (cs *CarStore) Stat(ctx context.Context, usr models.Uid) ([]UserStat, error
...
+	for _, sh := range shs {
+		if err := cs.deleteShardFile(ctx, sh); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+			outErr = err
+		}
+	}
+
+	return outErr
```
ericvolp12, line 997: Is this bit just to return at least one `IsNotExist` if any of the shards end up already having been deleted but we don't wanna let that stop us from deleting the others, however if there's some other error we're interested in stopping immediately and returning?

ericvolp12 (approved):
LGTM

carstore/bs.go
whyrusleeping: exactly, yeah. i want to at least try to delete everything, if we fail because the file is already gone, we continue (but still let our caller know it happened). if we fail for some other error i think it makes sense to bail out immediately

PR #322, opened 2023-09-19
crypto: support for multibase (de)ser of private keys

bnewbold:
See also: https://github.com/bluesky-social/atproto/discussions/1633

and internal discussions


ericvolp12 (approved):
LGTM

PR #323, opened 2023-09-19
processRepo callback needs deltastore bs for gap fill

whyrusleeping:
Im not confident this is right yet, but giving it a try to see if it fixes our issues


whyrusleeping:
added some startup logs because i noticed restarting the bgs took a noticeably long amount of time

PR #324, opened 2023-09-19
split up batches of deletes during compaction

whyrusleeping:
turns out you cant pass infinitely sized arrays to your sql driver.


ericvolp12 (approved):
LGTM

PR #326, opened 2023-09-20
Implement some unimplemented Atproto Sync BGS Handlers

ericvolp12:
This change implements a few more `com.atproto.sync.*` endpoints in the BGS.

After this PR we'll need to figure out Blobs and the Listing endpoints.

For `ListRepos` we should just be able to walk through the DB in some consistent order (maybe by UIDs).

### DIFF
--- a/bgs/handlers.go
+++ b/bgs/handlers.go
@@ -198,6 +198,10 @@ func (s *BGS) handleComAtprotoSyncListRepos(ctx context.Context, cursor string, 
 		return nil, fmt.Errorf("failed to get users: %w", err)
 	}
 
+	if len(users) == 0 {
+		return &comatprototypes.SyncListRepos_Output{}, nil
+	}
+
 	resp := &comatprototypes.SyncListRepos_Output{
 		Repos: []*comatprototypes.SyncListRepos_Repo{},
 	}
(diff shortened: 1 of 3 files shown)


bgs/handlers.go
```diff
@@ -6,36 +6,33 @@ import (
...
 	comatprototypes "github.com/bluesky-social/indigo/api/atproto"
+	"gorm.io/gorm"
+
 	"github.com/bluesky-social/indigo/util"
 	"github.com/bluesky-social/indigo/xrpc"
+	"github.com/ipfs/go-cid"
 	"github.com/labstack/echo/v4"
 )
 
+// Since the removal of repo history, this is the same as GetRepo but without the "since" bit
```
whyrusleeping, line 22: should just remove this endpoint

bgs/handlers.go
```diff
@@ -155,9 +176,74 @@ func (s *BGS) handleComAtprotoSyncListBlobs(ctx context.Context, cursor string,
...
+	c := int64(0)
+	if cursor != "" {
+		c, err = strconv.ParseInt(cursor, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("invalid cursor: %w", err)
+		}
+	}
+
+	users := []User{}
+	if err := s.db.Model(&User{}).Where("uid > ?", c).Limit(limit).Find(&users).Error; err != nil {
```
whyrusleeping, line 194: does limit actually do what we want here without an 'order by'?

whyrusleeping (approved):
a couple comments, otherwise LGTM

bgs/handlers.go
ericvolp12: o yea good catch

bgs/handlers.go
ericvolp12: Yeah I think so, it'll be deprecated soon

PR #328, opened 2023-09-21
improved compaction, minimize shard count while minimizing IO

whyrusleeping:
The main idea here is to have each shard contain twice the number of blocks as the next shard, under the logic that older shards will need to be rewritten less and also accessed less in general.

still to discuss, should there be a max size on shards?


ericvolp12 (approved):
LGTM

PR #330, opened 2023-09-22
Allow bulk dids for takedown

emilyliu7321:
- Modifies`takeDownAccountCmd` to take multiple dids for bulk takedowns 
- Renames cli command `takeDownAccount` to standardize with other commands

Example command:
`./gosky admin take-down --reason="<reason here>" --admin-user="<did here>" --revert-actions exampleuser.bsky.social exampleuser2.bsky.social`


whyrusleeping (approved):
lgtm, i might have pulled some of this out as a subroutine or something but if it works it works. Thanks emily!

PR #332, opened 2023-09-22
in the case of missing blocks, fall back to full repo fetch

whyrusleeping:
If for some reason we end up with missing data, add a multi-stage fallback to first fetch the missing revision range ( in the case of an event slice erroring) and then if data is still missing, fall back to fetching the entire repo.

also added a debug command to fetch and verify data in a repo.


ericvolp12 (approved):
LGTM

PR #334, opened 2023-09-22
break up requests for block refs

whyrusleeping:
turns out some repos have *a lot* of shards and this makes sql unhappy.


ericvolp12 (approved):
YEah ok

PR #338, opened 2023-09-25
Coalesce requests for the same DID or Handle in Directory

ericvolp12:
This change uses a `sync.Map` to coalesce requests for the same DID or Handle during a lookup to prevent lots of requests for the same content from going out before the first one comes back to get cached.


bnewbold (approved):
Nice, love how simple this is.

Would always be nice to have test coverage, but won't block merging if you want to move ahead with palomar indexing ASAP.

PR #346, opened 2023-09-26
if catching up on buffered jobs fails, fallback to getRepo

whyrusleeping:
im not entirely happy with the flow here, feels kinda jank. The main thing we want to do is just keep trying if the event sync fails instead of returning an error (which results in the users repo getting behind until they send another event).

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -420,7 +420,7 @@ func (ix *Indexer) fetchRepo(ctx context.Context, c *xrpc.Client, pds *models.PD
 	repo, err := comatproto.SyncGetRepo(ctx, c, did, rev)
 	if err != nil {
 		reposFetched.WithLabelValues("fail").Inc()
-		return nil, fmt.Errorf("failed to fetch repo: %w", err)
+		return nil, fmt.Errorf("failed to fetch repo (did=%s,rev=%s,host=%s): %w", did, rev, pds.Host, err)
 	}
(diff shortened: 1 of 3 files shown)


whyrusleeping:
this is now a 'clean up FetchAndIndexRepo' PR

indexer/indexer.go
```diff
@@ -412,69 +446,48 @@ func (ix *Indexer) FetchAndIndexRepo(ctx context.Context, job *crawlWork) error
...
+		return err
 	}
-	reposFetched.WithLabelValues("success").Inc()
 
 	// this process will send individual indexing events back to the indexer, doing a 'fast forward' of the users entire history
 	// we probably want alternative ways of doing this for 'very large' or 'very old' repos, but this works for now
 	if err := ix.repomgr.ImportNewRepo(ctx, ai.Uid, ai.Did, bytes.NewReader(repo), &rev); err != nil {
 		span.RecordError(err)
 
 		if ipld.IsNotFound(err) {
```
ericvolp12, line 484: Do we want to log when this happens? Also can we make the error messages a bit more informative i.e. `failed to full-resync repo after catchup failure: %w` or something?

whyrusleeping: yeah, thats a good idea

ericvolp12 (approved):
lgtm, BGS is prod now for search so pls rollout to staging and test first :P

PR #349, opened 2023-09-27
palomar: default q to '*' when there is only filter

bnewbold:
I haven't tested against an actual index, but there is a parsing test, and i'm pretty confident this will resolve the `from:blah` issue.


ericvolp12 (approved):
LGTM

PR #351, opened 2023-09-27
Allow crawling repos that have invalid handles

ericvolp12:
We still want to index repos with invalid handles, it should be the downstream consumer's responsibility to validate a handle, especially cause it can change and become invalid without a new HandleChange event being emitted. These events are just hints that you should go grab the DID doc for this user cause it may have changed.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -2,6 +2,7 @@ package bgs
 
 import (
 	"context"
+	"database/sql"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


bgs/bgs.go
```diff
@@ -1017,9 +1024,12 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
...
 	u := User{
-		Handle: handle,
-		Did:    did,
-		PDS:    peering.ID,
+		Did:         did,
+		PDS:         peering.ID,
+		ValidHandle: validHandle,
+	}
+	if validHandle {
+		u.Handle = handle
```
whyrusleeping, line 1032: i think we have a uniqueness constraint on handle in the db so we will have to explicitly set it to NULL

ericvolp12: Ah oops, yeah that's not a pointer, good call

ericvolp12: How do we do that with a `Create` using GORM?

ericvolp12: Might need to update that model to use a sql.NullString I think

whyrusleeping: I would just manually write some SQL for this case. I hate using orm types for this, makes other things more annoying.

whyrusleeping: ah, i see the change already. Thats not terrible i guess

whyrusleeping (approved):

PR #353, opened 2023-09-28
for now, add duplicate stale refs to the keep set

whyrusleeping:
if a deleted block matches multiple blockRefs in a given repo, we can't safely assume its no longer referenced. The 'correct' way to move forward is to do a full repo traversal to enumerate the 'correct' set of blocks referenced in the MST, but that requires more thought and is actually something that needs to be done at the repo manager level (and involves extra locking and other shenanigans that i dont want to implement just yet).

So for now, we add duplicate blocks to the keep set and just assume they are still being referenced somewhere.

### DIFF
--- a/bgs/admin.go
+++ b/bgs/admin.go
@@ -506,5 +506,7 @@ func (bgs *BGS) handleAdminResetRepo(e echo.Context) error {
 		return err
 	}
 
-	return nil
+	return e.JSON(200, map[string]any{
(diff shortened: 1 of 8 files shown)


whyrusleeping:
something wonky with this, ends up with missing blocks sometimes.. need to investigate

ericvolp12 (approved):
sure

PR #356, opened 2023-09-29
bgs: proposed changes from review

bnewbold:
Had a couple proposed changes from reviewing this code, was easiest to just edit locally and send a PR. feel free to keep or drop anything here.

other review notes:

Do we need to do the blob sync stuff at all in BGS? not sure that is in-scope. maybe it is disabled by not configuring when deployed, but could just rip the code out also.

As a high-level thing, many error cases result in `err` getting returned, but I believe we just log and move on. Maybe event processing should have multiple high-level results:

- success: processed and emitted an event
- skip: intentionally did not process event, and a downstream event was not emitted. eg, takedowns, some categories of invalid event
- dirty: event *should* have been processed, but failed, in a way that leaves repo in partially-processed state. could flag the repo as dirty and needing cleanup. eg, identity resolution timeout
- error: bad overall error, like disk full or database connection lost. instead of rapidly erroring on many sequential events, should probably just crash or halt

kind of an architecture philosophy thing and maybe you want to stick with the current system (pros/cons), but as we firm up all the possible validation error cases a framework like the above might be helpful.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -765,16 +765,19 @@ func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *event
 		}
 
 		if evt.Rebase {
-			return fmt.Errorf("rebase was true in event seq:%d,host:%s", evt.Seq, host.Host)
+			log.Warnw("rebase was true in event", "did", evt.Repo, "seq", evt.Seq, "host", host.Host)
+			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


whyrusleeping (approved):

ericvolp12 (approved):
Seems legit to me

ericvolp12:
Makes me wonder if we want to be forwarding event times produced by the PDS or if we should be recreating the timestamp with when _we_ saw them...

PR #360, opened 2023-10-02
syntax: Datetime, URI, CID, TID

bnewbold:
More of these parse helpers.

`Datetime` is the main thing here, with a bunch of test cases.

TID is also helpful, and a bunch of little interop issues came up when implementing this.

The CID and URI ones are a bit weird. URIs aren't very well spec'd in a general/abstract sense, and for CIDs we already have `go-cid` which is what would normally be used. But I think it is good to have string-only validation that ensures exact pass-through, and can be used as a type signature.

---

Questions for review:

- the TID clock generator really isn't a "syntax" thing. I slipped it in b/c it is only a few lines of code and stdlib (no external deps). but maybe it should live elsewhere?


atproto/syntax/tid.go
```diff
@@ -0,0 +1,132 @@
...
+	s := ""
+	for i := 0; i < 13; i++ {
+		s = string(Base32SortAlphabet[v&0x1F]) + s
+		v = v >> 5
+	}
+	return TID(s)
+}
+
+// Constructs a new TID from a UNIX timestamp (in milliseconds) and clock ID value.
+func NewTID(unixMilis int64, clockId uint) TID {
```
ericvolp12, line 56: I wonder if this should take a `time.Time` and extract the milliseconds for us. Thinking about the calling conventions here and Go doesn't really deal with millisecond time often.

bnewbold: ah, good catch on the millis/micros, I fixed the behavior but didn't remember to update the variable names (or a test call).

I added `NewTIDFromTime`. I'd be open to renaming and have `NewTID` take a time and rename the existing to `NewTIDFromUnixMicros` (eg). I think it is fine either way.

ericvolp12 (approved):
Sure, there may be some iteration as we start using these but looks good!

PR #361, opened 2023-10-02
BGS: allow non-SSL BGS with SSL PDS

snarfed:
seems like this should hopefully be allowed for testing, and the main combination we want to prevent is SSL (ie prod) BGS with non-SSL PDS...?


ericvolp12 (approved):
LGTM
Gonna let build run tho

PR #363, opened 2023-10-03
identity: add fallback DNS servers

bnewbold:
Quick addition to DNS handle resolution to allow configuring a set of fallback DNS servers which will be tried after regular DNS and (optional) authoritative DNS.

Haven't tested super carefully. Does seem kind of aggressive to do all these DNS queries for *every* failed DNS lookup, but we can tweak at larger scale if needed.


ericvolp12 (approved):
Sure

PR #367, opened 2023-10-03
more trace attributes and some tweaks

whyrusleeping:
a small grab-bag of changes, most things are either tweaking a batch size or adding an attribute to a trace.
The one 'real' change here is falling back to an individual block read instead of a prefetch if the shard in question is over a size limit.


ericvolp12 (approved):
Sure

PR #368, opened 2023-10-03
extract all interactions with database in carstore to separate interface

whyrusleeping:
step 1 of operation "every user has a sqlite db in the carstore"


jazware (approved):
LGTM, might need some recombobulating for Sqlite stuff but this looks like a step towards it.

PR #371, opened 2023-10-04
add commands for deleting accounts

whyrusleeping:


cmd/gosky/main.go
```diff
@@ -1566,3 +1567,44 @@ var rebaseRepoCmd = &cli.Command{
...
+	Action: func(cctx *cli.Context) error {
+		xrpcc, err := cliutil.GetXrpcClient(cctx, false)
+		if err != nil {
+			return err
+		}
+
+		token := cctx.Args().First()
+		password := cctx.Args().Get(1)
+
+		err = comatproto.ServerDeleteAccount(context.TODO(), xrpcc, &comatproto.ServerDeleteAccount_Input{
```
ericvolp12, line 1599: Do you just wanna pass the `cctx.Context` through here? Same with above.

PR #373, opened 2023-10-05
Resync with a PDS by comparing revs insted of commit CIDs

ericvolp12:
When resyncing with a PDS, if users are active then the CID we get from `ListRepos` can become out of date by the time we start comparing the CIDs against our local repo heads (i.e. new events come in for some users, we update the repo head in our local Carstore, and our CID is fresher than the one in-memory from `ListRepos`).

Since the Repo Rev is monotonically increasing, we can use the Rev we get back from `ListRepos` as a "low-water-mark" such that, if our local Rev for that repo is > the one we get from the PDS, we know we're current. This should reduce the number of unnecessary repos being recrawled during a resync.


whyrusleeping (approved):
LGTM

Issue #306, opened 2023-09-13
atproto/identity: add a const for invalid handle


bnewbold:
as mentioned in https://github.com/bluesky-social/indigo/pull/304

PR #377, opened 2023-10-06
disallow invalid handle TLDs

bnewbold:
Realized this wasn't being enforced at resolution time like it should be. These are allowed in *syntax* but not at resolution time.

Also adds the `.example` and `.alt` TLDs, and defines invalid handle as a constant (requested earlier). (updating spec doc and typescript impl in parallel)

closes: https://github.com/bluesky-social/indigo/issues/306


ericvolp12 (approved):
LGTM

PR #380, opened 2023-10-06
add flag to skip big shards during compaction

whyrusleeping:


whyrusleeping:
if a shard is 'big enough', dont bother compacting it if we just want to go fast.

This will result in some data not getting cleaned up that should, but we already have that as a problem, and we need to implement a 'full clean compaction' step anyways.

carstore/bs.go
```diff
@@ -1228,6 +1239,27 @@ func (cs *CarStore) CompactUserShards(ctx context.Context, user models.Uid) (*Co
...
+			}
+
+			if size > BigShardThreshold {
+				skip = i + 1
+			} else {
+				break
+			}
+		}
+		shards = shards[skip:]
+	}
```
ericvolp12, line 1261: This skip logic is assuming that shards of increasing Seq are monotonically decreasing in size? This feels a bit too clever to me for some reason, it requires a lot of context to be able to read this code and know it's doing what it's supposed to be doing. Could we instead create a new slice of shards we don't want to skip as we check if each shard is within the BigShardThreshold, and then use that for compaction?

whyrusleeping: no, because we can't skip shards between other shards that might be compacted.
Say we have:
A B C D E

and we skip A and C. If B and D get compacted together, then we have an out-of-order shard set

whyrusleeping: we also want to avoid running a stat on every shard, especially in cases where the user has like 20k shards because we havent compacted them in a while.

ericvolp12: Makes sense, thanks for clarifying. Maybe that deserves a comment somewhere in the code?

ericvolp12 (approved):
LGTM

PR #382, opened 2023-10-09
combine all staleRefs for a given operation into a single record

whyrusleeping:
writing and loading large amounts of individual staleRefs is a bit slower than it should be. Since we only ever access them all at once, storing them as a single row makes things a lot easier, especially on the write path (the indexes get pretty large and inserting into them many times gets expensive)

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -1472,16 +1472,24 @@ func (cs *CarStore) deleteStaleRefs(ctx context.Context, uid models.Uid, brefs [
 		}
 	}
 
-	if err := cs.meta.Delete(&staleRef{}, "usr = ?", uid).Error; err != nil {
+	txn := cs.meta.Begin()
+
+	if err := txn.Delete(&staleRef{}, "usr = ?", uid).Error; err != nil {
 		return err
 	}
 
 	// now create a new staleRef with all the refs we couldnt clear out
-	if err := cs.meta.Create(&staleRef{
-		Usr:  uid,
-		Cids: packCids(staleToKeep),
-	}).Error; err != nil {
-		return err
+	if len(staleToKeep) > 0 {
+		if err := txn.Create(&staleRef{
+			Usr:  uid,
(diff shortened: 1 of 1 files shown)


carstore/bs.go
```diff
@@ -1411,35 +1449,42 @@ func (cs *CarStore) deleteStaleRefs(ctx context.Context, brefs []blockRef, stale
...
 
-		if err := cs.meta.Delete(&staleRef{}, "id in (?)", sl).Error; err != nil {
-			return 0, err
-		}
+	// now create a new staleRef with all the refs we couldnt clear out
+	if err := cs.meta.Create(&staleRef{
+		Usr:  uid,
+		Cids: packCids(staleToKeep),
+	}).Error; err != nil {
+		return err
```
ericvolp12, line 1484: Should we only be doing this if `staleToKeep` is not empty?

carstore/bs.go
```diff
@@ -1411,35 +1449,42 @@ func (cs *CarStore) deleteStaleRefs(ctx context.Context, brefs []blockRef, stale
...
 
-	chunkSize := 10000
-	for i := 0; i < len(staleToDelete); i += chunkSize {
-		sl := staleToDelete[i:]
-		if len(sl) > chunkSize {
-			sl = sl[:chunkSize]
-		}
+	if err := cs.meta.Delete(&staleRef{}, "usr = ?", uid).Error; err != nil {
+		return err
+	}
```
ericvolp12, line 1477: Do we want to wrap this in a transaction with the new staleref creation? Is there a potential race condition here or do we only create `staleRefs` in this function? I can't remember if we also create them when a record gets deleted if any relevant blocks have no more refs.

ericvolp12 (approved):
I think this looks good, I definitely wanna see how it shakes out in staging/sandbox first to make sure the transition from the old unpacked refs to the new packed refs is smooth.

PR #383, opened 2023-10-10
Palomar profile search improvements

bnewbold:
These are tweaks to search quality, mostly typeahead, with no change to index schema (so no reindexing needed). A couple basic fixes, and then some boosting and rescoring hacks.

I tested these by having local palomar query the prod search index over an SSH tunnel. I didn't check query latencies, but nothing was crazy slow, and based on the changes i'm fairly confident there won't be a significant latency hit for typeahead. Profile search may get a bit slower.

There are some remaining quality issues we will probably want to revisit, but would probably require re-indexing. Eg, we are not doing any stemming at all on fulltext fields (my bad!). This will impact profile descriptions, but in particular posts.

### DIFF
--- a/cmd/palomar/main.go
+++ b/cmd/palomar/main.go
@@ -219,12 +219,8 @@ var runCmd = &cli.Command{
 }
 
 var elasticCheckCmd = &cli.Command{
-	Name: "elastic-check",
-	Flags: []cli.Flag{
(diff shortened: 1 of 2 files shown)


ericvolp12 (approved):
Looks good, I can deploy this to staging which is running the prod dataset right now to test it out a bit.

PR #390, opened 2023-10-11
Starting to build out compactor daemon

ericvolp12:
Daemonizing the Compactor

Design goals:
- Always be running compactions
- Enqueue repos eagerly based on activity, more active repos should be compacted more frequently
- Easily re-compact all repos
- Get state of currently running compaction
- Verbose output from compactions with stats and timing etc.


ericvolp12:
Gonna roll this out to the test environment once it's done backfilling repos

whyrusleeping (approved):

PR #392, opened 2023-10-11
palomar: remove query re-scoring

bnewbold:

PR #393, opened 2023-10-11
palomar: revert 'and' in a query

bnewbold:
This is on top of https://github.com/bluesky-social/indigo/pull/392, and removes the "and" in query as well


bnewbold:
this wasn't the performance problem

PR #397, opened 2023-10-13
make it optional to hydrate records in internal events

whyrusleeping:


whyrusleeping:
this causes a whole lot of memory allocations and cpu churn for no reason in the bgs, where its not used (outside of spidering records, which we are leaning towards permanently disabling anyways)

ericvolp12:
Gonna deploy to `test` and poke at it

ericvolp12 (approved):
lgtm

PR #398, opened 2023-10-16
Add error handling for listInviteTreeCmd

emilyliu7321:
Adds error handling so we can iterate through an entire invite tree even if it contains accounts that have been taken down or otherwise error out on getRepo.


whyrusleeping (approved):
LGTM

PR #401, opened 2023-10-23
Use a 4MiB buffer for event stream connections instead of 10KiB

ericvolp12:
This change should improve performance of event playback to higher latency connections which have currently been limited in throughput because of bandwidth-delay-product issues.

https://linuxczar.net/blog/2016/09/18/bandwidth-delay-product/

Using a 4MiB buffer should allow for up to 16MiB/sec of throughput on a connection with a 250ms RTT.

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -517,7 +517,7 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
 	defer cancel()
 
 	// TODO: authhhh
-	conn, err := websocket.Upgrade(c.Response(), c.Request(), c.Response().Header(), 10<<10, 10<<10)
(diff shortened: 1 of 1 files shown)


bnewbold:
I wonder if it is better/necessary to reach down to the kernel socket config, instead of the gorilla (websocket-level) config:

- https://man7.org/linux/man-pages/man2/setsockopt.2.html
- https://iximiuz.com/en/posts/go-net-http-setsockopt-example/#net-http-set-server-socket-options
- https://pkg.go.dev/github.com/gorilla/websocket#hdr-Buffers

bnewbold:
cc: @whyrusleeping

PR #402, opened 2023-10-27
Fix/events playback

whyrusleeping:
This is all the refactors and fixes from #388, but without the new splitter daemon code. should get this merged asap


ericvolp12 (approved):
Yeah OK, let's roll out to Test first and poke at it tho

PR #404, opened 2023-10-30
improve timeouts on event streams a bit

whyrusleeping:


ericvolp12 (approved):
Let's run this in test as well with a slow consumer to see how it behaves, we've got a good number of consumers consuming slower than realtime right now. Might be good to kick them off.

whyrusleeping:
so this shouldnt actually affect consumers, but now im realizing that we probably want to put it in the Pong handler or something, because if we don't receive any events from the pds for a minute, this will kill the connection (even if the ping-pongs still happen)

ericvolp12 (approved):
Definitely want to see this one tested with a slow consumer before prod rollout but hopefully should prevent stuck sockets.

PR #406, opened 2023-10-31
refactor: replace util/version with versioninfo dep

bnewbold:
this should fix version metadata being missing or incorrect for several tools.

the old mechanism required pushing through a linker variable at build time (eg, in dockerfile). this package pulls the info a different way which works more reliably. note that it doesn't work with `go run`, but it does with `go build`. because we don't have tags/versions in this repo, we'll get versions like `rev-42f80c4`, which seems fine.


ericvolp12 (approved):
Sure, LGTM

PR #409, opened 2023-11-01
fix goroutine leak in playback

whyrusleeping:
This fixes the goroutine leak but im not confident it addresses the event leak (the larger memory problem).


ericvolp12 (approved):
LGTM

PR #411, opened 2023-11-02
drop unique constraint on handles

whyrusleeping:
When users change their handles, that change doesnt necessarily propagate through the whole world consistently. The big issue this causes for us is if a user changes their handle from X to Y, and we arent able to validate it, we dont update our accounting of that change. Then if a new user creates an account with handle X, we fail to correctly initialize that user because of the uniqueness constraint.

I think I'll have to manually do these migrations, i doubt gorm will correctly swap the indexes from unique to not-unique.

I read through all the code referencing unique conflicts and handle setting and I believe this should work just fine.


ericvolp12 (approved):
LGTM

PR #414, opened 2023-11-02
test that empty events make it through

whyrusleeping:
quick test to make sure empty events get handled properly


ericvolp12 (approved):
ok

PR #417, opened 2023-11-03
drop consumers if their out buffer fills up

whyrusleeping:
right now if these buffers fill up we just start dropping events, and then when the buffers drain we just resume sending events. this seems quite wrong.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -78,7 +78,7 @@ func (em *EventManager) broadcastEvent(evt *XRPCStreamEvent) {
 				log.Warnw("dropping slow consumer due to event overflow", "bufferSize", len(s.outgoing), "ident", s.ident)
 				go func(torem *Subscriber) {
 					select {
-					case s.outgoing <- &XRPCStreamEvent{
+					case torem.outgoing <- &XRPCStreamEvent{
 						Error: &ErrorFrame{
 							Error: "ConsumerTooSlow",
 						},
--- a/events/events.go
+++ b/events/events.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
 	label "github.com/bluesky-social/indigo/api/label"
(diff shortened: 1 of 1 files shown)


whyrusleeping:
hrm.... this isnt quite right, it doesnt close out the connections and will result in a hanging subscriber

whyrusleeping:
This should work now

ericvolp12 (approved):
LGTM

ericvolp12:
Actually we should probably emit a `ConsumerTooSlow` Error Event as well before closing

events/events.go
```diff
@@ -73,11 +73,11 @@ func (em *EventManager) broadcastEvent(evt *XRPCStreamEvent) {
 			select {
 			case s.outgoing <- evt:
 			case <-s.done:
+			default:
+				log.Warnw("event overflow", "bufferSize", len(s.outgoing), "ident", s.ident)
```
ericvolp12, line 77: Might be helpful to add like `dropping slow consumer due to event buffer overflow` here to give the log message a bit more context.

events/events.go
```diff
@@ -73,11 +73,11 @@ func (em *EventManager) broadcastEvent(evt *XRPCStreamEvent) {
...
 				go func(torem *Subscriber) {
-					em.rmSubscriber(torem)
+					torem.cleanup()
```
ericvolp12, line 79: Do we want to emit a `ConsumerTooSlow` error frame here before closing?

events/events.go
```diff
@@ -73,11 +74,20 @@ func (em *EventManager) broadcastEvent(evt *XRPCStreamEvent) {
 			select {
 			case s.outgoing <- evt:
 			case <-s.done:
+			default:
+				log.Warnw("dropping slow consumer due to event overflow", "bufferSize", len(s.outgoing), "ident", s.ident)
 				go func(torem *Subscriber) {
-					em.rmSubscriber(torem)
+					select {
+					case s.outgoing <- &XRPCStreamEvent{
```
ericvolp12, line 81: should this be `torem.outgoing`?

whyrusleeping: good catch reviewer.

events/events.go
```diff
@@ -73,11 +74,20 @@ func (em *EventManager) broadcastEvent(evt *XRPCStreamEvent) {
...
 				go func(torem *Subscriber) {
-					em.rmSubscriber(torem)
+					select {
+					case torem.outgoing <- &XRPCStreamEvent{
+						Error: &ErrorFrame{
+							Error: "ConsumerTooSlow",
+						},
+					}:
+					case <-time.After(time.Second * 5):
+						log.Warnw("failed to send error frame to backed up consumer", "ident", s.ident)
```
ericvolp12, line 87: this should be `torem.ident`

whyrusleeping: ah dammit

ericvolp12 (approved):
LGTM again

PR #421, opened 2023-11-07
Create Literiver, a mass-sqlite backup tool

jazware:
For more context and history check out https://github.com/ericvolp12/litestream/pull/1


ericvolp12:
Some notes on restoring as I work it out:
- You can grab the whole bucket down via `rclone` pretty quickly by setting up an S3 datastore and then running:
```
$ rclone copy {rclone_source}:{bucket_name} {destination_folder} --include "*" --progress --checkers 32 --transfers 32
```
Still working on turning them from the raw files back into a ton of DBs but this is a good starting point cause now we know which DBs were backed up to that bucket (something we would otherwise have to track somehow in a metadata file or something), and should be able to perform _local_ restores to our live DB folder.

PR #425, opened 2023-11-09
only traverse cbor blocks when running mst diff

whyrusleeping:
I spaced the fact that we include proper dag-cbor cid links to images which are not actually included in the car slices. In the case where someone drops a reference to an image (deletes an image post, or updates their profile picture) then we would hit this 'ipld not found' error on the image cid (since we never got it in the first place).

Here, we skip any CID that isnt CBOR, This is a reasonable assumption to make for now, unless we start dangling random cbor graphs off of the MST, at which point ill probably need to get more clever (I'd really rather not).


ericvolp12 (approved):
Sure

PR #429, opened 2023-11-10
make testing slow consumers easy

whyrusleeping:
used this for testing the bugfix PR earlier

### DIFF
--- a/cmd/gosky/main.go
+++ b/cmd/gosky/main.go
@@ -143,6 +143,10 @@ var readRepoStreamCmd = &cli.Command{
 		&cli.BoolFlag{
 			Name: "resolve-handles",
 		},
+		&cli.DurationFlag{
+			Name:  "read-delay",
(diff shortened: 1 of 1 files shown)


cmd/gosky/main.go
```diff
@@ -204,8 +208,14 @@ var readRepoStreamCmd = &cli.Command{
...
 		}
 
+		rr := cctx.Duration("read-delay")
+
 		rsc := &events.RepoStreamCallbacks{
 			RepoCommit: func(evt *comatproto.SyncSubscribeRepos_Commit) error {
+				if rr != 0 {
+					time.Sleep(rr)
+				}
+
```
ericvolp12, line 218: Could we use a `rate.limiter` here instead and then allow a user to configure a maximum throughput? That might be a bit easier to use.

ericvolp12 (approved):
lgtm

PR #433, opened 2023-11-11
handle resolution does not depend on any xrpc routes

whyrusleeping:
handle resolution is currently failing on things using this method because the PDS check here is just wrong.


ericvolp12 (approved):
sure

bnewbold (approved):
looks tidy

PR #434, opened 2023-11-14
automod: rules engine framework

bnewbold:
`automod` is a package to help write moderation automation "rules" to catch things like spam and flag accounts or content for human review (or potentially even auto-action).

`hepa` is a simple service which consumes from a BGS firehose and pushes events through `automod`.

This is an early draft of the rules engine, some notes:

- rules are written in golang here. we might end up using a DSL or scripting language of some sort eventually
- the framework and some example rules are here in a public repo. we'll obviously keep some rules and configuration private in the long run
- intention is to use redis for caching and state persistence. will likely end up with additional datastores


automod/engine.go
```diff
@@ -0,0 +1,171 @@
...
+	ident, err := e.Directory.LookupDID(ctx, did)
+	if err != nil {
+		return fmt.Errorf("resolving identity: %w", err)
+	}
+	if ident == nil {
+		return fmt.Errorf("identity not found for did: %s", did.String())
+	}
+	collection := strings.SplitN(path, "/", 2)[0]
+
+	switch collection {
```
whyrusleeping, line 83: fyi: You can do a type switch in Go:
```go
switch rec := rec.(type) {
case *appbsky.FeedPost:
  // inside this case, rec is the correct type
}
```

automod/event.go
```diff
@@ -0,0 +1,273 @@
...
+
+// information about a repo/account/identity, always pre-populated and relevant to many rules
+type AccountMeta struct {
+	Identity *identity.Identity
+	// TODO: createdAt / age
+}
+
+// base type for events. events are both containers for data about the event itself (similar to an HTTP request type); aggregate results and state (counters, mod actions) to be persisted after all rules are run; and act as an API for additional network reads and operations.
+type Event struct {
+	Engine            *Engine
```
whyrusleeping, line 26: it feels a bit odd to me that the event would carry the engine reference, seems like the 'persist' methods should just be on the engine itself and we should pass the event through that.

automod/engine.go
bnewbold: yeah, I had it that way first, but need to verify collection/type mapping somehow (aka, type needs to be FeedPost *and* collection is "app.bsky.feed.post") and this felt more legible than switch on type and verify collection string.

automod/event.go
bnewbold: yeah, agree this feels a bit weird.

the rules reach back to the engine other ways as well though: helpers on the event can fetch counters and things. it might be more golang-idiomatic to pass both the engine and the event to all rules.

part of this might be poor naming. "events" here are kind of more like `echo.Context`, maybe "EventContext" would be better.

might still be a bad design pattern.

bnewbold: the more obvious/idiomatic thing would be to have the rules be methods on the engine, or methods on some context object in general. then you would access things like counters, logger, or other external resources through the parent struct. but i'm trying hard to make rules free-floating rules so they can be implemented in arbitrary packages and collected at configuration time.

cmd/hepa/consumer.go
```diff
@@ -0,0 +1,130 @@
...
+
+	logger := s.logger.With("event", "commit", "did", evt.Repo, "rev", evt.Rev, "seq", evt.Seq)
+	logger.Debug("received commit event")
+
+	if evt.TooBig {
+		logger.Warn("skipping tooBig events for now")
+		return nil
+	}
+
+	did, err := syntax.ParseDID(evt.Repo)
```
whyrusleeping, line 90: not now, but we should probably make lexgen use the syntax.DID type for fields that have format:did

whyrusleeping (approved):
This seems like a good scaffold to start from. lets get it running and start poking at things.

cmd/hepa/consumer.go
bnewbold: yup! I can't remember if I have tested yet, but the syntax.DID type implements the text marshaling interface, so this should Just Work

PR #435, opened 2023-11-14
auto detect correct pds for 'home' server in compare repos

whyrusleeping:
calling getRepo against bsky.social for an account thats been migrated no longer works, so this makes the command "do what i mean" more


ericvolp12 (approved):
lgtm

bnewbold:
lmk if you have feedback on the directory and syntax APIs.

PR #439, opened 2023-11-17
Tell RobustHTTPClient to treat rate-limiting as non-retryable error

imax9000:
This is a simplest possible approach to pass through hitting a rate limit to the application that uses the default XRPC client.

It's a design question if this is what you want to have. Alternatives include:
* Gate this behaviour change behind a flag in `xrpc.Client` struct
* Allowing to specify a threshold for the duration until rate limit reset, below which the client would keep the current behaviour

FWIW, JS client seems to not have any retry logic in place.

### DIFF
--- a/util/http.go
+++ b/util/http.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"context"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ericvolp12 (approved):
This looks okay to me, I generally use a new HTTP client with OTEL telemetry hooked up when using the XRPC package myself so I haven't dealt with some of this retry logic personally. I think this is reasonable behavior though.

PR #440, opened 2023-11-17
temporarily remove the did doc field from auth response

whyrusleeping:
this value is typed as "unknown" in lexicon, which we don't really know how to map into a go type here


ericvolp12 (approved):
Yeah sure, this is annoying, why don't they use a language with types?

whyrusleeping:
:shrug:

imax9000:
createSession is affected too

PR #442, opened 2023-11-20
a couple supercollider tweaks

whyrusleeping:
some fixes that helped when i was doing supercollider tests


cmd/supercollider/main.go
```diff
@@ -762,6 +777,10 @@ func (s *Server) HandleSubscribeRepos(c echo.Context) error {
 		limiter.Wait(ctx)
 
 		if err := header.UnmarshalCBOR(f); err != nil {
+			if err == io.EOF {
+				log.Warnf("reached end of playback")
```
ericvolp12, line 781: `log.warnf` doesn't exist in the standard logging package

PR #448, opened 2023-11-22
lexgen: slightly more robust 'unkown' handling (hack)

bnewbold:
This is a bit more robust than the current situation where we manually comment things out.

Just looks for the specific field names `didDoc` or `plcOp` and plops in an `*interface{]` type instead for those.

Could also have done `*map[string]interface{}` but I was worried not initializing that datatype; maybe unwarranted. 

My assumption is that `json.RawMessage` would be best here for JSON stuff, but that causes problems for CBOR stuff.

### DIFF
--- a/api/atproto/repodescribeRepo.go
+++ b/api/atproto/repodescribeRepo.go
@@ -7,17 +7,16 @@ package atproto
 import (
 	"context"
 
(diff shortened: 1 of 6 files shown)


whyrusleeping (approved):
Ah nice, thank you!

PR #451, opened 2023-11-24
Fix TID generator to avoid duplicates

imax9000:
Third call on the same microsecond would not match the `if` condition and would return the same value as the first call.


whyrusleeping (approved):
Ah, good catch. That seems right to me

PR #452, opened 2023-11-27
update generic deps

bnewbold:
run `go get -u <pkg>` manually for a bunch of our direct deps.

did not bite off the large upgrade to IPFS `boxo`. I tried doing that partially and it worked ok, but we should upgrade from `ipfs/go-car` to `ipfs/go-car/v2` everywhere first to really take advantage of it, otherwise we are left with a dual dependency tree with stuff like ipns, bitswap, and unixfs still sticking around.

AFAIK `ipfs/go-car/v2` works fine (we use it in a few places), but the API isn't directly compatible.


ericvolp12 (approved):
This looks okay, the OTEL versions really like to be in sync so I hope they're happy. I ran into some issues a few weeks ago in my own projects when my OTEL versions didn't match the ones in indigo for some reason but figured out a fix. Let's roll this out in staging first just to make sure our traces still show up.

bnewbold:
I just now deployed palomar staging with this PR's commit, and was able to see some new traces in Grafana indexing posts ~10 minutes since after the deploy, so I think it's working.

PR #453, opened 2023-11-27
quick command to hit the list labels endpoint

whyrusleeping:


cmd/gosky/main.go
```diff
@@ -707,3 +709,53 @@ var parseRkey = &cli.Command{
...
+	},
+	Action: func(cctx *cli.Context) error {
+
+		ctx := context.TODO()
+
+		delta := cctx.Duration("since")
+		since := time.Now().Add(-1 * delta).UnixMilli()
+
+		xrpcc := &xrpc.Client{
+			Host: "https://api.bsky.app",
```
bnewbold, line 730: I think this should be configurable via `ATP_BSKY_HOST` or `ATP_APPVIEW_HOST` or similar.

Looking forward it may be an external moderation service, in which case `ATP_MOD_HOST`, but this endpoint is somewhat temporary and also we'll need to refactor a few appview connections anyways.

bnewbold (approved):
Left a comment about making the host configurable but otherwise lgtm, nice addition

PR #454, opened 2023-11-27
Make the backfiller aware of revs and fill gaps

whyrusleeping:


whyrusleeping:
Hrn... jobs don't recover from error states. Need to figure that one out

backfill/gormstore.go
```diff
@@ -118,46 +119,108 @@ func (s *Gormstore) EnqueueJob(repo string) error {
...
 	defer j.lk.Unlock()
 
 	switch j.state {
 	case StateComplete:
-		return false, ErrJobComplete
-	case StateInProgress:
-	// keep going and buffer the op
-	default:
 		return false, nil
+	case StateInProgress, StateEnqueued:
```
ericvolp12, line 129: Why are we buffering ops on enqueued backfill jobs here? They won't have been checked out yet if they're still enqueued so any events they get will be included in the checkout, right?

backfill/gormstore.go
```diff
@@ -195,19 +273,39 @@ func (j *Gormjob) SetState(ctx context.Context, state string) error {
 	return j.db.Save(j.dbj).Error
 }
 
-func (j *Gormjob) FlushBufferedOps(ctx context.Context, fn func(kind, path string, rec *typegen.CBORMarshaler, cid *cid.Cid) error) error {
+var ErrEventGap = fmt.Errorf("buffered event revs did not line up")
+
+func (j *Gormjob) FlushBufferedOps(ctx context.Context, fn func(kind, path string, rec typegen.CBORMarshaler, cid *cid.Cid) error) error {
```
ericvolp12, line 278: How does this work with deletes? IIRC the `rec` being a pointer worked with deletes because it'd be `nil` when deleting a record, does it just end up being an empty `typegen.CBORMarshaler` in this case?

backfill/gormstore.go
whyrusleeping: ah, thats true

whyrusleeping:
need to figure something out there, maybe just keep a counter of how many times its failed and do automatic retries up to a point?

ericvolp12:
> need to figure something out there, maybe just keep a counter of how many times its failed and do automatic retries up to a point?

would we do retries when we see more events from the repo or somehow enqueue it again for a retry when it fails? would probably also want some kind of backoff in there.

whyrusleeping:
i think we should automatically re-enqueue it, after some duration

whyrusleeping:
oh the other thing i havent implemented et is the gap fill handling logic, it spits out an error when we detect a gap, and resets the job state, but when we actually go to process that we arent filtering to just the gap records

PR #456, opened 2023-11-29
automod polish

bnewbold:
More rules, helpers, etc.


bnewbold:
Going to self-merge, will be reviewing and refactoring as we go, but want to start reviewing PRs on-top of main, not on top of this branch.

PR #458, opened 2023-11-29
identity: small improvements

bnewbold:
I'm not sure why the error wasn't getting returned; might have been a typo? Probably worth re-reading that function again in review. I hit a nil de-reference pointer in automod caused by that function returning `nil, nil`.

### DIFF
--- a/atproto/identity/base_directory.go
+++ b/atproto/identity/base_directory.go
@@ -33,6 +33,7 @@ type BaseDirectory struct {
 var _ Directory = (*BaseDirectory)(nil)
 
 func (d *BaseDirectory) LookupHandle(ctx context.Context, h syntax.Handle) (*Identity, error) {
+	h = h.Normalize()
 	did, err := d.ResolveHandle(ctx, h)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


ericvolp12 (approved):
LGTM

ericvolp12:
> I'm not sure why the error wasn't getting returned; might have been a typo? Probably worth re-reading that function again in review. I hit a nil de-reference pointer in automod caused by that function returning `nil, nil`.

I don't really get how that function can ever return `nil, nil` though, `entry` is always set as something and we always return a pointer to a real `entry` object.

bnewbold:
Ah, sorry, to clarify it was the call of `updateDID()` in `LookupDID` (https://github.com/bluesky-social/indigo/pull/458/files#diff-b91976d7b63c04bc9f338df9dba703c4441b915b0535ea598821efa1350d25f9R216-R219) which was unpacking the `Identity` as `doc` (nil) and returned err (also nil) and `return doc, err`.

We could have `LookupDID` unpack the internal `Err` from the entry doc, but that feels a bit convoluted to me. I think of the `Err` in the entry doc as only being used for cache persistence.

PR #460, opened 2023-12-04
trust (for now) ops in events

whyrusleeping:


whyrusleeping:
This does less work (and also avoids a little PDS bug) when processing events.

whyrusleeping:
@ericvolp12 i'd still love to get your thoughts on this one, but im gonna merge it now to unblock some other things im working on

PR #461, opened 2023-12-04
be more permissive in lexicon union type strings

whyrusleeping:


bnewbold:
disposition is to *not* merge this branch, and to be strict about using "just the collection" when referencing `#main` defs in a `$type` field. expect some dev comms, spec update, and test coverage around this.

whyrusleeping:
Agreed.

PR #462, opened 2023-12-06
`make cborgen`

bnewbold:
Seems like this needed to be run after recent cbor-gen dep update.


whyrusleeping:
I think this got superceded by one of my other PRs, right?

bnewbold:
yup! no worries

PR #463, opened 2023-12-06
CLI usage improvements, and crypto private key gen helper

bnewbold:
I only recently discovered the `ArgsUsage` field in urfave/cli command config. It helps a lot!


whyrusleeping (approved):
Thank you!

PR #464, opened 2023-12-06
automod: extracting some packages, and concurrency safety

warpfork:
A few minor hopefully-advancements to the automod/hepa packages:

- `countstore` is now extracted as a package.
- Synchronization primitives are added to MemCountStore.  This fixes crashes when running `hepa` without the use of Redis.
- Additional testing, to make sure that MemCountStore gets enough exercise that the race detector has something to work on in tests.
- Attempted to sneak in some docs along the way!

The diff looks a little larger than it is, because of git and file moves... but individual commits are broken down as much as possible, so those should also be fine to review in isolation if preferred.

### DIFF
--- a/automod/countstore/countstore.go
+++ b/automod/countstore/countstore.go
@@ -2,6 +2,9 @@ package countstore
 
 import (
 	"context"
+	"fmt"
+	"log/slog"
+	"time"
 )
 
 const (
@@ -16,4 +19,20 @@ type CountStore interface {
 	// TODO: batch increment method
 	GetCountDistinct(ctx context.Context, name, bucket, period string) (int, error)
 	IncrementDistinct(ctx context.Context, name, bucket, val string) error
+}
+
+func periodBucket(name, val, period string) string {
+	switch period {
+	case PeriodTotal:
+		return fmt.Sprintf("%s/%s", name, val)
+	case PeriodDay:
+		t := time.Now().UTC().Format(time.DateOnly)
+		return fmt.Sprintf("%s/%s/%s", name, val, t)
+	case PeriodHour:
+		t := time.Now().UTC().Format(time.RFC3339)[0:13]
+		return fmt.Sprintf("%s/%s/%s", name, val, t)
+	default:
+		slog.Warn("unhandled counter period", "period", period)
+		return fmt.Sprintf("%s/%s", name, val)
+	}
 }
--- a/automod/countstore/countstore.go
+++ b/automod/countstore/countstore.go
@@ -20,15 +20,20 @@ const (
 // See the `Period*` consts for the available period types.
 //
 // The "GetCount" and "Increment" methods perform actual counting.
-// The "*Distinct" methods store booleans for values, rather than incrementing;
-// fetching "GetCountDistinct" asks how many values have been seen at least once,
-// rather than how many times they've been seen.
+// The "*Distinct" methods have a different behavior:
+// "IncrementDistinct" marks a value as seen at least once,
+// and "GetCountDistinct" asks how _many_ values have been seen at least once.
+//
+// The exact implementation and precision of the "*Distinct" methods may vary:
+// in the MemCountStore implementation, it is precise (it's based on large maps);
+// in the RedisCountStore implementation, it uses the Redis "pfcount" feature,
+// which is based on a HyperLogLog datastructure which has probablistic properties
+// (see https://redis.io/commands/pfcount/ ).
 type CountStore interface {
 	GetCount(ctx context.Context, name, val, period string) (int, error)
 	Increment(ctx context.Context, name, val string) error
 	// TODO: batch increment method
 	GetCountDistinct(ctx context.Context, name, bucket, period string) (int, error)
-	// REVIEW: s/IncrementDistinct/MarkDistinct/ ?
 	IncrementDistinct(ctx context.Context, name, bucket, val string) error
 }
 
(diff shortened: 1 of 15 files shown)


automod/countstore/countstore_mem.go
```diff
@@ -0,0 +1,64 @@
...
+	if !ok {
+		return 0, nil
+	}
+	return v.Size(), nil
+}
+
+func (s MemCountStore) IncrementDistinct(ctx context.Context, name, bucket, val string) error {
+	for _, p := range []string{PeriodTotal, PeriodDay, PeriodHour} {
+		k := periodBucket(name, bucket, p)
+		s.DistinctCounts.Compute(k,func(nested *xsync.MapOf[string, bool], _ bool) (*xsync.MapOf[string, bool], bool)  {
```
bnewbold, line 55: does the nested map need to be a sync primitive as well? my intuition is that the sync control of the outer map should protect the inner map.

this isn't a review-blocking issue, just curious

automod/countstore/countstore_test.go
```diff
@@ -0,0 +1,100 @@
...
+	assert := assert.New(t)
+	ctx := context.Background()
+
+	cs := NewMemCountStore()
+
+	c, err := cs.GetCount(ctx, "test1", "val1", PeriodTotal)
+	assert.NoError(err)
+	assert.Equal(0, c)
+
+	// Increment two different values from four different goroutines.
```
bnewbold, line 54: :gem:

automod/countstore/countstore.go
```diff
@@ -0,0 +1,54 @@
...
+// The "GetCount" and "Increment" methods perform actual counting.
+// The "*Distinct" methods have a different behavior:
+// "IncrementDistinct" marks a value as seen at least once,
+// and "GetCountDistinct" asks how _many_ values have been seen at least once.
+//
+// The exact implementation and precision of the "*Distinct" methods may vary:
+// in the MemCountStore implementation, it is precise (it's based on large maps);
+// in the RedisCountStore implementation, it uses the Redis "pfcount" feature,
+// which is based on a HyperLogLog datastructure which has probablistic properties
+// (see https://redis.io/commands/pfcount/ ).
```
bnewbold, line 31: this is great, thanks!

we could also note here that `MemCountStore` will grow unbounded in memory over time. this will always be the case for `PeriodTotal` counts, but in particular for distinct maps, and that the hour/day entries are never cleared. which is an intentional simplification for now.

I can take this as a task for me to add

bnewbold (approved):
This looks great!

Left some comments, but nothing merge-blocking from my perspective. If you can add a note about memory usage for MemCountStore that would be great, but I can also add that just before merging.

As a process note, we (bluesky) often leave "approve" review status on PRs if there is only a small amount of cleanup to finish, indicating that anybody could merge if urgent, or that the original author can self-merge after small changes. We tend to keep "request changes" to indicate a blocking/veto issue ("revise and resubmit for review"), and comment for drive-by comments (if review wasn't requested).

automod/countstore/countstore_mem.go
warpfork: ...Okay, that's a good question.  I've been nerdsnipped and I'm actually still a bit unsure.

My theoretical understanding: I don't think it would be safe.  (And not just like "aw that's a minor inaccuracy", like, I think the race detector _should_ flag it.)

- I read the source of this `syncx` package:
  - This map _does_ use some mutexing during writes.
  - And that mutexing _does_ cover the compute function callback body.
  - It does _not_ use mutexting during reads.  (Reads are done purely with atomic load operations.)
  - (And obviously, once you've done a read, you've... got the value; even if there was mutexing during the read (and there's not), it couldn't still be in effect after you hold the value.)
- So do we do anything on the value after a read?  Yes, we do have a read where we extract the inner map... and then call `.Size()` on it (or would call `len()` if it was a native map).
- Is that operation safe?
  - `xsync`'s `.Size()`: yes, defintely.  It's another atomic read.
  - native `len()`:  No. `len()` of native maps is considered a read for race detector purposes.  Performing one outside of a mutex (even if all writes are mutexed) _will_ trigger the race detector.  (I even wrote a little side test to be sure -- separate from our nested maps situation here.  Yep, easy to trigger.)

Practically verifying the whole thing: I wrote a quick variant with a plain native map on the inner, and... I'm unable to get the race detector to flag it, nor get the test to fail.

:exploding_head: 

Overall: I'm suspicious my testing is somehow not aggressive enough, rather than confident that a plain map on the inner is fully safe.

automod/countstore/countstore_mem.go
```diff
@@ -0,0 +1,64 @@
+package countstore
+
+import (
+	"context"
+
+	"github.com/puzpuzpuz/xsync/v3"
```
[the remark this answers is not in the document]
warpfork: Right, just so -- we couldn't get rid of the synchronization on the map itself, and once we've got that done, our time computing the increment, and `xsync.Map`'s time getting the new value written to the slot in the map internals (which `Compute` has already got a cursor to)... is approximately free.

(I think if we really wanted to put the pedal to the metal, a hybrid datastructure that's a map that knows its values are ints could be further optimized.  But I don't think there's any strong nudges to think diving any deeper on that is warranted yet.)

automod/countstore/countstore.go
warpfork: added docs on this, good idea

PR #465, opened 2023-12-07
automod: action limits; create reports for interaction churn

bnewbold:
The motivation here is to start auto-reporting in production based on specific rules. Specifically, this PR would start reporting on interactions churn (follow/unfollow), letting human mods confirm before taking action on an account.

Before we do that, need to de-duplicate reports. For example, if an account creates thousands of spammy posts, only want to report the account once. Generally want to prevent run-away rules from creating millions of reports, or doing thousands of automated account takedowns (for example).

This PR prevents re-reporting based on daily counters (fast checks), as well as double-checking against the mod service API just before filing a report (slower, but reliable).

This PR also adds "quotas" for mod actions, implemented using the counter system. This isn't perfect (the counter system itself might be buggy or broken (causing duplicate actions), but seems like a good start. If quotas are exceeded, automod will log and skip taking additional actions until the next day.

### DIFF
--- a/automod/circuit_breaker_test.go
+++ b/automod/circuit_breaker_test.go
@@ -1,0 +1,94 @@
+package automod
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	appbsky "github.com/bluesky-social/indigo/api/bsky"
+	"github.com/bluesky-social/indigo/atproto/identity"
+	"github.com/bluesky-social/indigo/atproto/syntax"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func alwaysTakedownRecordRule(evt *RecordEvent) error {
+	evt.TakedownRecord()
+	return nil
+}
+
+func alwaysReportRecordRule(evt *RecordEvent) error {
+	evt.ReportRecord(ReportReasonOther, "test report")
+	return nil
+}
+
+func TestTakedownCircuitBreaker(t *testing.T) {
+	assert := assert.New(t)
+	ctx := context.Background()
+	engine := engineFixture()
+	dir := identity.NewMockDirectory()
+	engine.Directory = &dir
+	// note that this is a record-level action, not account-level
+	engine.Rules = RuleSet{
+		RecordRules: []RecordRuleFunc{
+			alwaysTakedownRecordRule,
+		},
+	}
+
+	path := "app.bsky.feed.post/abc123"
+	cid1 := "cid123"
+	p1 := appbsky.FeedPost{Text: "some post blah"}
+
+	// generate double the quote of events; expect to only count the quote worth of actions
+	for i := 0; i < 2*QuotaModTakedownDay; i++ {
+		ident := identity.Identity{
+			DID:    syntax.DID(fmt.Sprintf("did:plc:abc%d", i)),
+			Handle: syntax.Handle("handle.example.com"),
+		}
+		dir.Insert(ident)
+		assert.NoError(engine.ProcessRecord(ctx, ident.DID, path, cid1, &p1))
+	}
+
+	takedowns, err := engine.GetCount("automod-quota", "takedown", PeriodDay)
+	assert.NoError(err)
+	assert.Equal(QuotaModTakedownDay, takedowns)
+
+	reports, err := engine.GetCount("automod-quota", "report", PeriodDay)
+	assert.NoError(err)
+	assert.Equal(0, reports)
+}
+
+func TestReportCircuitBreaker(t *testing.T) {
+	assert := assert.New(t)
+	ctx := context.Background()
+	engine := engineFixture()
+	dir := identity.NewMockDirectory()
+	engine.Directory = &dir
+	engine.Rules = RuleSet{
+		RecordRules: []RecordRuleFunc{
+			alwaysReportRecordRule,
+		},
+	}
+
+	path := "app.bsky.feed.post/abc123"
+	cid1 := "cid123"
+	p1 := appbsky.FeedPost{Text: "some post blah"}
+
+	// generate double the quote of events; expect to only count the quote worth of actions
+	for i := 0; i < 2*QuotaModReportDay; i++ {
+		ident := identity.Identity{
+			DID:    syntax.DID(fmt.Sprintf("did:plc:abc%d", i)),
+			Handle: syntax.Handle("handle.example.com"),
+		}
+		dir.Insert(ident)
+		assert.NoError(engine.ProcessRecord(ctx, ident.DID, path, cid1, &p1))
+	}
+
+	takedowns, err := engine.GetCount("automod-quota", "takedown", PeriodDay)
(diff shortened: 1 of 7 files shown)


automod/event.go
```diff
@@ -149,9 +160,38 @@ func (e *RepoEvent) PersistAccountActions(ctx context.Context) error {
 			newFlags = append(newFlags, val)
 		}
 	}
-	newReports := e.AccountReports
```
warpfork, line 152: I think this works, but this function as a whole took me quite a while to read.  I'm not immediately sure how I'd suggest simplifying it, but in tasting notes...

- this function is getting a tad lengthy now, at 183 lines,
- has many different variables controlling different parts of control flow
  - some booleans;
  - some being lengths of variables
    - which in some cases are variables that are initialized at one point, then reinitialized to zero again later to affect the control flow...
- and is mutating the main object as it goes.

I'm not quite sure how I'd go about breaking it up.  But I think pulling some functions out just to create boundaries to how far one has to mentally trace scopes around (even if the functions aren't actually reused anywhere!) might be useful soon.

automod/event.go
```diff
@@ -201,7 +242,38 @@ func (e *RepoEvent) PersistAccountActions(ctx context.Context) error {
 		needsPurge = true
 	}
 	for _, mr := range newReports {
-		_, err := comatproto.ModerationCreateReport(ctx, xrpcc, &comatproto.ModerationCreateReport_Input{
+		// before creating a report, query to see if automod has already reported this account in the past week for the same reason
```
warpfork, line 245: Suggest pulling this whole block out to a function.  Looks like it should have a pretty simple `(bool, error)` return type.

automod/event.go
```diff
@@ -149,9 +160,38 @@ func (e *RepoEvent) PersistAccountActions(ctx context.Context) error {
...
+	if newTakedown {
+		if e.GetCount("automod-quota", "takedown", PeriodDay) >= QuotaModTakedownDay {
+			e.Logger.Warn("CIRCUIT BREAKER: automod takedowns")
+			newTakedown = false
+		} else {
+			e.Increment("automod-quota", "takedown")
+		}
+	}
+
 	if newTakedown || len(newLabels) > 0 || len(newFlags) > 0 || len(newReports) > 0 {
```
warpfork, line 195: (again in the "i'm not quite sure how I'd factor these things out, but, thinking out loud" vein...)

These four condition clauses seem like a good focal point for guiding a (?future) refactor.

When I was reviewing this, I noticed myself checking this line against all the other branches below to see if a slack notification would be accurately sent for exactly every event that would have at least one other persisted effect.

I think the answer is still "yes", as intended.  It's just getting a bit harder to see as all the other instances of the same condition are drifting further apart.  Maybe roughly a function per section below that's gated on each of these clauses?   Or perhaps if we bundle all these values in a small struct, so their collective role is easier to see (and maybe add some logging methods to that, as a bonus)?

Also: I believe all the `needsPurge = true` assignments and eventual check of that bool can be replaced by... just checking 3 of these 4 again, directly, too.

warpfork (commented):
Looks mostly good to me.

I'm getting a little twitchy at the sheer size of the `PersistAccountActions` func, though.  I think I read it, and I think I understand it, and together with the tests I think it's probably correct... but it's getting harder to be confident, and taking longer to read.  It might be good to break it up soon.  I tried to offer a few inline thoughts about where, but they're very preliminary and it's up to you if you can make any use of them :)

automod/event.go
```diff
@@ -90,6 +96,11 @@ func (e *RepoEvent) AddAccountFlag(val string) {
 }
 
 func (e *RepoEvent) ReportAccount(reason, comment string) {
+	if comment == "" {
+		comment = "(automod)"
+	} else {
+		comment = "automod: " + comment
+	}
```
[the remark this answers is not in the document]
bnewbold: nice

automod/event.go
bnewbold: :+1:

automod/event.go
bnewbold: there is a bit of a corner case with reports, where it might not be known that they were not created until the actual attempt to do so (because there is an existing report). in that case a slack message gets sent out, but no report is created, and the account metadata cache is not purged.

I guess we could treat the skip-record-creation-at-last-minute thing as a real corner-case and always purge the cache?

bnewbold:
I took at pass at breaking this function out and refactoring it. The control flow and behavior should be the same (still pretty complex), but hopefully clearer and more readable.

automod/circuit_breaker_test.go
```diff
@@ -0,0 +1,94 @@
...
+		RecordRules: []RecordRuleFunc{
+			alwaysReportRecordRule,
+		},
+	}
+
+	path := "app.bsky.feed.post/abc123"
+	cid1 := "cid123"
+	p1 := appbsky.FeedPost{Text: "some post blah"}
+
+	// generate double the quote of events; expect to only count the quote worth of actions
```
warpfork, line 77:
```suggestion
	// generate double the quota of events; expect to only count the quota worth of actions
```

automod/event.go
```diff
@@ -183,32 +198,124 @@ func (e *RepoEvent) PersistAccountActions(ctx context.Context) error {
...
+
+	// de-dupe actions
+	newLabels := dedupeLabelActions(e.AccountLabels, e.Account.AccountLabels, e.Account.AccountNegatedLabels)
+	newFlags := dedupeFlagActions(e.AccountFlags, e.Account.AccountFlags)
+
+	// don't report the same account multiple times on the same day for the same reason. this is a quick check; we also query the mod service API just before creating the report.
+	newReports := circuitBreakReports(e, dedupeReportActions(e, e.AccountReports))
+	newTakedown := circuitBreakTakedown(e, e.AccountTakedown && !e.Account.Takendown)
+
+	anyModActions := newTakedown || len(newLabels) > 0 || len(newFlags) > 0 || len(newReports) > 0
```
warpfork, line 302: super better :star_struck:

automod/event.go
warpfork: Yeah, I was wondering about that too.

I was wondering if maybe the heavier query for prior reports should actually move earlier?  Because it's also potentially mildly confusing if the slack notification gets sent out saying a report to happen but it's actually not.

warpfork (approved):
Looking pretty darn good to me!

automod/event.go
bnewbold: :relieved:

automod/event.go
bnewbold: It isn't super great how it is, but it does enable running with pushing to slack without actual admin-token access to the mod service, which I think is helpful as a possible way to deploy/operate, both for internal folks and third parties. The idea there is folks can develop rules and just manually independently report them in-app, based on a slack channel. Maybe that is too jank to care about, but I was running that way for a bit.

PR #466, opened 2023-12-07
automod: identical reply rule

bnewbold:
Two enabling features:

- cheap consistent non-cryptographic-strength hashing of strings for use in counter keys (went with uint64 murmur3, which was already in dependency tree)
- ability to increment a counter for a single time period, to control counter key space growth (for redis)

This initial version of the rule counts replies to any other user in the same bucket, not distinct-accounts-with-same-reply-text. I'm a little worried about redis memory growth if we have a HyperLogLog for each author+text combination (as opposed to simple counter int). Maybe the redis implementation is clever and efficient for the small-distinct case? Or maybe RAM is cheap enough?

This branch will conflict with https://github.com/bluesky-social/indigo/pull/464. Plan to merge that one first, then i'll rebase this one.


bnewbold:
can trigger this against prod data with:

```
go run ./cmd/hepa/ process-recent hackerdarkweb.bsky.social
```

automod/rules/replies.go
```diff
@@ -27,3 +27,24 @@ func ReplyCountPostRule(evt *automod.RecordEvent, post *appbsky.FeedPost) error
...
+// There can be legitimate situations that trigger this rule, so in most situations should be a "report" not "label" action.
+func IdenticalReplyPostRule(evt *automod.RecordEvent, post *appbsky.FeedPost) error {
+	if post.Reply == nil || IsSelfThread(evt, post) {
+		return nil
+	}
+
+	// use a specific period (IncrementPeriod()) to reduce the number of counters (one per unique post text)
+	period := automod.PeriodDay
+	bucket := evt.Account.Identity.DID.String() + "/" + HashOfString(post.Text)
+	if evt.GetCount("reply-text", bucket, period) >= identicalReplyLimit {
```
warpfork, line 44: Right now, this is proceeding to act immediately on a hash collision.  Do you think it would be reasonable to do a more expensive check to see if things are *actually* identical?

I don't know how bad a false positive is here.  Maybe if the threshhold for identical in sheer count is moderate, it's unlikely to trigger on reasonable real human behavior?

automod/rules/replies.go
```diff
@@ -27,3 +27,24 @@ func ReplyCountPostRule(evt *automod.RecordEvent, post *appbsky.FeedPost) error
...
+		evt.AddAccountFlag("multi-identical-reply")
```
warpfork, line 45: Not new to this PR, but I think some docs on the semantic distinctions between `AccountLabels`, `AccountFlags`, and `AccountReports` are needed somewhere (and then perhaps most methods touching them should have a quick pointer to that doc).  The last one of the group is close to self-explanatory, but the first two are both just `[]string` in the code and don't autounpack their meaning very readily.

warpfork (approved):
I think this is an overall LGTM, given that it can be iterated on further if the collision rate were to turn out problematically high.

automod/rules/replies.go
bnewbold: I think the naive false positive rate (for the 64bit variant of murmur3) is low enough to not worry about it and not do secondary network requests to check for exact matches.

This isn't a cryptographic hash, so attacks could be a concern for some rules.

Generally, I feel like all the counters we are using here should be treated as a bit fuzzy, at least for record-level counts. It is totally possible for events to get partially-processed (and partially persisted) and then re-processed again after a crash. I think the semantics and kinds of rules and actions we write are generally resilient to this: doing things like reporting for human review, or having large margins before taking fully-automated action.

automod/rules/replies.go
bnewbold: mmm! added doc comments to all the fields for most of the RepoEvent variants

bnewbold:
Merged with main (copied additions to countstore over to that package), and added a bunch of doc comments.

I'm not too concerned about distinct counter collisions, but can revisit.

PR #467, opened 2023-12-08
automod: remove mistleading url+mention rules from default list

bnewbold:
These are just too noisy right now, turning up client bugs much more often than intentionally misleading content.

May come back to these later; definitely leaving code (and tests) in place.


warpfork (approved):
Agreed

PR #468, opened 2023-12-09
update lexicons to latest

whyrusleeping:


lex/gen.go
```diff
@@ -1194,10 +1194,12 @@ func (ts *TypeSchema) writeTypeDefinition(name string, w io.Writer) error {
...
-		if len(ts.Properties) == 0 {
-			pf("type %s interface{}\n", name)
-			return nil
-		}
+		/*
+			if len(ts.Properties) == 0 {
+				pf("type %s interface{}\n", name)
+				return nil
+			}
+		*/
```
bnewbold, line 1202: was this intentional? if we'll come back to this later, a comment describing why it is commented out for now

bnewbold (approved):
There is a confusing (to me) commented-out bit. Otherwise seems fine!

bnewbold:
also, while you are touching lexgen, might want to merge this small PR which reduces the number of manual tweaks required: https://github.com/bluesky-social/indigo/pull/448

lex/gen.go
whyrusleeping: yeah, this was actually not used for a long while, and now that it is again, we actually want empty structs defined (the threadgate rules are all empty structs with a type field)

PR #471, opened 2023-12-11
reset backoff after successful connection in fedmgr

whyrusleeping:


whyrusleeping:
the problem this has is that if the connection works but the stream is bad, we will still loop here endlessly.
I wonder if I should check that we've actually processed additional events (by checking that the cursor has increased) before resetting the backoff counter

PR #473, opened 2023-12-13
relay: return explicit empty array, not nil, for sync.listRepos

bnewbold:
This is a work-around for the specific situation in: https://github.com/bluesky-social/atproto/issues/1949

We may need to have lexgen stubs ensure that any required array types in Output types are set to an explicit empty array, instead of nil? I haven't dug in to the details of JSON marshaling and CBOR marshaling to understand the scope of this issue yet, but the root of the issue is that the typescript code is being more strict about "empty" types than golang is. The typescript strictness seems both correct and reasonable to me, FWIW.

### DIFF
--- a/bgs/handlers.go
+++ b/bgs/handlers.go
@@ -202,6 +202,10 @@ func (s *BGS) handleComAtprotoSyncListRepos(ctx context.Context, cursor string, 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ericvolp12 (approved):
LGTM

PR #478, opened 2023-12-15
identity: refactor error hierarchy and error handling

bnewbold:
Root motivation here was that a bunch of remote errors were trickling up as full errors when doing identity resolution (via DID), while in reality they should result in "invalid.handle".

I didn't want to just force everything in to ErrHandleNotFound, so I created some new error types (and renamed a couple existing types), and ensured that we were using errors.Is() for matching. This way downstream code can distinguish between "not found" and "error during resolution" when calling the lower-level methods, if they care (eg, a debugging tool or interface would care). Also wrapped the errors instead of just returning, so additional context couple be tacked on the message.

There are still some situations where the root cause might be a local error (like local wifi) as opposed to a remote error (server is broken), and we can't distinguish those, at least for now.

Note: I checked and recent versions of golang does explicitly allow multiple `%w` in a single `fmt.Errorf()`

### DIFF
--- a/atproto/identity/handle.go
+++ b/atproto/identity/handle.go
@@ -20,7 +20,7 @@ func parseTXTResp(res []string) (syntax.DID, error) {
 			parts := strings.SplitN(s, "=", 2)
 			did, err := syntax.ParseDID(parts[1])
 			if err != nil {
-				return "", fmt.Errorf("invalid DID in handle DNS record: %w", err)
+				return "", fmt.Errorf("%w: invalid DID in handle DNS record: %w", ErrHandleResolutionFailed, err)
 			}
 			return did, nil
 		}
@@ -39,7 +39,7 @@ func (d *BaseDirectory) ResolveHandleDNS(ctx context.Context, handle syntax.Hand
 		}
 	}
 	if err != nil {
-		return "", fmt.Errorf("handle DNS resolution failed: %w", err)
+		return "", fmt.Errorf("%w: %w", ErrHandleResolutionFailed, err)
 	}
 	return parseTXTResp(res)
 }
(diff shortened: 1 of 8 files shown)


atproto/identity/handle.go
```diff
@@ -138,22 +138,29 @@ func (d *BaseDirectory) ResolveHandleWellKnown(ctx context.Context, handle synta
...
-		return "", fmt.Errorf("failed to resolve handle (%s) through HTTP well-known route: %s", handle, err)
+		slog.Info("HTTP request failed for well-known handle resolution", "handle", handle, "err", err)
+		return "", fmt.Errorf("%w: HTTP request error: %w", ErrHandleResolutionFailed, err)
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		return "", fmt.Errorf("%w: HTTP 404 for %s", ErrHandleNotFound, handle)
 	}
 	if resp.StatusCode != http.StatusOK {
-		return "", fmt.Errorf("failed to resolve handle (%s) through HTTP well-known route: status=%d", handle, resp.StatusCode)
+		slog.Info("non-success HTTP response for well-known handle resolution", "handle", handle, "status_code", resp.StatusCode)
```
ericvolp12, line 151: We might want this to be debug level maybe? I could see this triggering a ton since we race resolutions.

atproto/identity/handle.go
```diff
@@ -138,22 +138,29 @@ func (d *BaseDirectory) ResolveHandleWellKnown(ctx context.Context, handle synta
 		var dnsErr *net.DNSError
 		if errors.As(err, &dnsErr) {
 			if dnsErr.IsNotFound {
-				return "", ErrHandleNotFound
+				return "", fmt.Errorf("%w: DNS NXDOMAIN for %s", ErrHandleNotFound, handle)
 			}
 		}
-		return "", fmt.Errorf("failed to resolve handle (%s) through HTTP well-known route: %s", handle, err)
+		slog.Info("HTTP request failed for well-known handle resolution", "handle", handle, "err", err)
```
ericvolp12, line 144: This should also be a debug instead of info

ericvolp12 (approved):
These look good, I'm not sure how I feel about the library logging stuff though. The errors are returned with lots of context and the caller should be able to log errors returned to get context on what's going on rather than having the library invoke a global logger at a high noise level.

bnewbold:
Oh, good call! I just pulled out all those "new" log lines. I'd added them in an earlier iteration on this branch when the error messages were fixed, and forgot to pull them out.

It might be a good idea to rip out all the other logging in this package, or at lease review the log levels, but I don't want to scope creep too much.

PR #479, opened 2023-12-15
identity: refactor cache directory to prevent nil/nil condition

bnewbold:
This is closely related to https://github.com/bluesky-social/indigo/pull/458, which tried to resolve an issue with DIDs. I was seeing very confusing behavior with handles; was able to confirm `"", nil` getting returned internally on real-world lookups.

I took a more aggressive approach and changed the return signatures of both update helper methods to just return an "Entry", not as a pointer, and no err. This makes the code flow much simpler.

I tacked on a special control-flow error message as defensive programming after getting bit twice on this (darn self!). Probably isn't necessary and smells a bit but :shrug:.

### DIFF
--- a/atproto/identity/cache_directory.go
+++ b/atproto/identity/cache_directory.go
@@ -90,7 +90,7 @@ func (d *CacheDirectory) IsIdentityStale(e *IdentityEntry) bool {
 	return false
 }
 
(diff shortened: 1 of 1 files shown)


ericvolp12 (approved):
sure

PR #480, opened 2023-12-15
util: remove ParseTimestamp

bnewbold:
In particular from search/palomar codebase where it is causing issues, but also in general the `syntax.ParseDatetime` method is more tested and compliant with the spec.


ericvolp12 (approved):
sure

PR #481, opened 2023-12-15
palomar: prod errors burndown

bnewbold:
I took a peek at prod error logs and found some issues:

- deleting search documents is failing, because the "document ID" (docID) was formatted wrong, because the `rkey` parameter is actually a full path, not just the record key part
- many actually-valid datetimes were not getting parsed correctly, resulting in posts not having correct createdAt, likely resulting in them not appearing with expected ranking in search results
- many HTTP 404 errors on fetches of base path; not really a big deal but might as well return healthcheck

I haven't tested these much yet, we should probably deploy to staging (which consumes the prod firehose, IIRC) and observe that for a bit.

### DIFF
--- a/search/indexing.go
+++ b/search/indexing.go
@@ -17,14 +17,26 @@ import (
 	esapi "github.com/opensearch-project/opensearch-go/v2/opensearchapi"
 )
 
-func (s *Server) deletePost(ctx context.Context, ident *identity.Identity, rkey string) error {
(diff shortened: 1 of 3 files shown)


bnewbold:
This is unblocked now, and has real fixes (checked that these errors are still happening in prod logs).

ericvolp12 (approved):

ericvolp12:
These look fine in staging from what I can see

PR #482, opened 2023-12-15
automod: mentions spike rule

warpfork:
Introduce a simple new rule for detecting and flagging spikes of mentions within a period.

### DIFF
--- a/automod/rules/spammentions.go
+++ b/automod/rules/spammentions.go
@@ -1,0 +1,38 @@
+package rules
+
+import (
+	appbsky "github.com/bluesky-social/indigo/api/bsky"
+	"github.com/bluesky-social/indigo/automod"
+)
+
+var _ automod.PostRuleFunc = SpamMentionsRule
+
+var mentionHourlyThreshold = 20
+
+// SpamMentionsRule looks for accounts which mention an unusually large number of distinct accounts per period.
+func SpamMentionsRule(evt *automod.RecordEvent, post *appbsky.FeedPost) error {
+	did := evt.Account.Identity.DID.String()
+
+	// Increment counters for all new mentions in this post.
+	var newMentions bool
+	for _, facet := range post.Facets {
+		for _, feature := range facet.Features {
+			mention := feature.RichtextFacet_Mention
+			if mention == nil {
+				continue
+			}
+			evt.IncrementDistinct("mentions", did, mention.Did)
+			newMentions = true
+		}
+	}
+
+	// If there were any new mentions, check if it's gotten spammy.
+	if !newMentions {
+		return nil
+	}
+	if mentionHourlyThreshold <= evt.GetCountDistinct("mentions", did, automod.PeriodHour) {
+		evt.AddAccountFlag("high-distinct-mentions")
+	}
+
+	return nil
+}
--- a/automod/rules/all.go
+++ b/automod/rules/all.go
@@ -19,7 +19,7 @@ func DefaultRules() automod.RuleSet {
 			ReplySingleKeywordPostRule,
 			AggressivePromotionRule,
 			IdenticalReplyPostRule,
-			SpamMentionsRule,
+			DistinctMentionsRule,
 		},
 		ProfileRules: []automod.ProfileRuleFunc{
 			GtubeProfileRule,
--- a/automod/rules/all.go
+++ b/automod/rules/all.go
@@ -19,6 +19,7 @@ func DefaultRules() automod.RuleSet {
 			ReplySingleKeywordPostRule,
 			AggressivePromotionRule,
 			IdenticalReplyPostRule,
+			SpamMentionsRule,
 		},
 		ProfileRules: []automod.ProfileRuleFunc{
 			GtubeProfileRule,
--- a/automod/rules/mentions.go
+++ b/automod/rules/mentions.go
@@ -5,12 +5,12 @@ import (
 	"github.com/bluesky-social/indigo/automod"
 )
 
-var _ automod.PostRuleFunc = SpamMentionsRule
+var _ automod.PostRuleFunc = DistinctMentionsRule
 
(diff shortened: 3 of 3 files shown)


automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
...
+			evt.IncrementDistinct("mentions", did, mention.Did)
+			newMentions = true
+		}
+	}
+
+	// If there were any new mentions, check if it's gotten spammy.
+	if !newMentions {
+		return nil
+	}
+	if mentionHourlyThreshold <= evt.GetCountDistinct("mentions", did, automod.PeriodHour) {
```
warpfork, line 33: Interestingly, I think it's basically impossible for something to trigger on a single message, due to the order of events: counter increments get buffered until persist time, so if we're reading the counts on the same topic like this during the same rule function, we're getting the data from before we started.

That's maybe fine, but seems like maybe we should spawn a todo to write something up about that in a doc block somewhere about things to be expecting as a rule author?

automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
+package rules
+
+import (
+	appbsky "github.com/bluesky-social/indigo/api/bsky"
+	"github.com/bluesky-social/indigo/automod"
+)
+
+var _ automod.PostRuleFunc = SpamMentionsRule
```
warpfork, line 8: This wasn't in other rules that I looked at, but I'd like to suggest a "type assertion" like this as a convention.

I find they're good for self-documenting code in general.  They're a good hint at the top of a file about what's coming.  And the closer it is to the type's definition, the more immediately useful any compiler is at bringing you to the origin of the problem if the invariant fails.

Right now, the only other place this gets checked is in the `DefaultRules` constructor _if your rule is not commented out_.  So that's a tad... not fragile per se, but not as sturdy as it could be, either.

bnewbold: Yeah! I'm a fan of those and have a couple in `indigo/atproto/crypto`. I would support having this be a norm in every rule.

For some reason I thought I had read a popular external style guide *against* this pattern, which was confusing, but maybe I was just confused. Uber likes it (for whatever Uber's opinion is worth):
https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

automod/rules/spammentions.go
bnewbold: Yup! And in particular with limits you'll hit most rules on "limit plus one" not the limit number directly. Agree this should be included in rule-writing guide.

automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
...
+			}
+			evt.IncrementDistinct("mentions", did, mention.Did)
+			newMentions = true
+		}
+	}
+
+	// If there were any new mentions, check if it's gotten spammy.
+	if !newMentions {
+		return nil
+	}
```
bnewbold, line 32: this skip-a-network-fetch-if-nothing-new is an interesting little idiom. I think my rules are mostly the other way around (read first, increment later), but this is potentially more efficient.

automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
+package rules
+
+import (
+	appbsky "github.com/bluesky-social/indigo/api/bsky"
+	"github.com/bluesky-social/indigo/automod"
+)
+
+var _ automod.PostRuleFunc = SpamMentionsRule
+
+var mentionHourlyThreshold = 20
```
bnewbold, line 10: FWIW, my intuition is that we'll need to bump this a lot. one example is follow-friday type threads where people list out a bunch of recommended follows, easily a few dozen.

in theory we could do some data science, but I think what probably works best is what you're doing here implicitly: start with just a "flag" which will result in a slack notification, and we can see how legit the hits are.

one workflow thing that will need improvement is flag naming. right now once a flag gets set, it sticks around forever. if we change the theshold, or re-write the rule entirely, it would make sense to purge all the old flags. we could use manually process to work around this (prefix new flags with "dev-" or "beta-"), but there are probably better options.

automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
...
+	did := evt.Account.Identity.DID.String()
+
+	// Increment counters for all new mentions in this post.
+	var newMentions bool
+	for _, facet := range post.Facets {
+		for _, feature := range facet.Features {
+			mention := feature.RichtextFacet_Mention
+			if mention == nil {
+				continue
+			}
```
bnewbold, line 23: there is an `ExtractFacets` helper function, though in this case the code looks pretty tight without it:
https://pkg.go.dev/github.com/bluesky-social/indigo/automod/rules#ExtractFacets

automod/rules/spammentions.go
```diff
@@ -0,0 +1,38 @@
...
+			evt.IncrementDistinct("mentions", did, mention.Did)
```
bnewbold, line 24: we use the `IncrementDistinctPeriod` variant to reduce the number of hyperloglog data counters in memory. on the other hand, distinct mentions *do* seem like the kind of thing we might want to persist long-term, or change our mind amount, so seems good to keep as-is.

bnewbold (approved):
This looks great! It is basically how I would write it.

Before i'd deploy this, i'd probably want to test it a bit, either with a capture of real-world behavior, or by reducing the threshold and testing against local or staging.

automod/rules/spammentions.go
warpfork: Hm, daily numbers might be useful too then.  Follow-friday behaviors could tip the hourly count is that number is relatively low, but if we also do a daily number that's considerably higher, and it should be easy for follow-friday to stay below that while real spam still goes above it.

A mechanism for counting hours-today-that-crossed-the-line would also buff that out.  Follow-friday shouldn't go above 1 or 2.

PR #483, opened 2023-12-15
palomar: handle bogus future createdAt better

bnewbold:
This tries to mitigate issues with bogus createdAt timestamps in post records. In particular this might (?) be causing a bunch of network traffic in-app when you search posts for "hello".

We have a batch of these in the current index which mess up ranking, so there is a query-time filter. A better mitigation would be to run a cleanup "update by query" which might look like this (untested):

```
POST palomar_post/_update_by_query
{
  "script": {
    "source": "ctx._source.created_at = null",
    "lang": "painless"
  },
  "query": {
    "range": {
      "created_at": {
        "gte": "2024-01-01"
      }
    }
  }
}
```

Our PDS currently enforces a reasonable createdAt, but it might stop this Lexicon-specific behavior some day (speculative), or third-party PDS implementations might not take this step. So this PR adds checks at index-time.

There is no test coverage here, and haven't checked against an actual index, but wanted to share partial work if helpful.

### DIFF
--- a/search/query.go
+++ b/search/query.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log/slog"
-	"time"
 
 	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/atproto/syntax"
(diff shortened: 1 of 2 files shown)


bnewbold:
This is an old PR, and probably lower-priority, but unblocked now with palomar deploys possible again (:tada:)

ericvolp12 (approved):
LGTM

PR #484, opened 2023-12-16
dont include did doc in create session for now

whyrusleeping:
The did doc is defined as an `any` type in the lexicon right now, which in all other cases implies a record (with a $type field). For now, lets just ignore this in the go code (its not necessary for anything) until we figure out a better way forward.

### DIFF
--- a/api/atproto/servercreateSession.go
+++ b/api/atproto/servercreateSession.go
@@ -7,7 +7,6 @@ package atproto
 import (
 	"context"
 
-	"github.com/bluesky-social/indigo/lex/util"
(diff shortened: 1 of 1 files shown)


bnewbold:
hmmm, I had just done a patch to lexgen which made this field, and a couple other non-record `unknown` types, an `interface{}`. but looks like that patch had gotten clobbered. the other field like this is `plcOp`.

my main concern is with having to manually comment out these fields after every run of lexgen, that is a footgun and somebody will forget or not know they need to do that.

imax9000:
Please always remove it from other places too. getSession and refreshSession are still broken

PR #485, opened 2023-12-17
Allow LexiconTypeDecoder to handle unknown types (at least from JSON)

imax9000:
Store the values as json.RawMessage and preserve them for marshaling.

CBOR counterpart not implemented due to lack of required functionality in cbor-gen, but can be added later.

Conversion between JSON and CBOR might be possible too, if Lexicon doesn't have any features that violate 1:1 mapping between JSON and CBOR data types.


lex/util/decoder.go
```diff
@@ -76,6 +76,10 @@ func CborDecodeValue(b []byte) (CBOR, error) {
 
 	t, ok := lexTypesMap[tstr]
 	if !ok {
+		// It is possible to gracefully handle this case, like in JsonDecodeValue,
```
whyrusleeping, line 79: `ScanForLinks` should properly advance the reader to the next object, ill mess around with this and see if we can get it working correctly

imax9000: FWIW, we don't necessarily have to do both in one go. Skipping CBOR for now means that firehose consumers will error out on unrecognized types. But having JSON calls not fail alone would be already an improvement.

My largest concern however is that this will trigger new code paths, shifting errors down the road. And it will uncover any bugs that are currently hidden behind a failure to unmarshal the whole message.

bnewbold:
To cross-reference, in case it is helpful, the `indigo/atproto/data` sub-package (PR waiting for review: https://github.com/bluesky-social/indigo/pull/407) can handle generic JSON and CBOR data. It does some light validation and will reject data which is, eg, valid JSON but could never be an atproto record.

Not as efficient as `json.RawMessage` (which is just bytes), and requires a re-marshal/unmarshal cycle to get to a golang struct using the current lexgen codebase.

PR #486, opened 2023-12-18
quick tweaks to rules at scale

bnewbold:
- skip short post text for identical replies
- start reporting bad word records for review

The "short post text" instead of skipping just empty also handles a common case of "handful of emoji".


warpfork (approved):
LGTM

PR #487, opened 2023-12-18
automod context plumbing

warpfork:
Plumb context through rules.

This is mostly not yet used, but prepares for future needs.  It does also remove one TODO -- there was one existing rule that uses a directory lookup, and was doing so without a fully connected context.

I took the liberty of adding type assertion "variables" to the header of each rule file, while at it.

(Currently also includes a few commits from another branch, but will rebase as soon as that's merged.)

### DIFF
--- a/automod/action_dedupe_test.go
+++ b/automod/action_dedupe_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func alwaysReportAccountRule(evt *RecordEvent) error {
+func alwaysReportAccountRule(ctx context.Context, evt *RecordEvent) error {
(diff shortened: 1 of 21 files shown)


bnewbold:
What do you think about plumbing the `context.Context` through the event itself, to avoid having an extra argument in the function signature? Eg, similar to `net/http` `HandleFunc` not taking a ctx argument, but having access to it via `Request.Context()`.

warpfork:
We can discard this.  Even as I was making these edits, I was coming around to thinking... yeah, the sheer args list length feels high and noisy.  I agree making some kind of "business context" one-stop-shop object feels like it will make a lot of sense.  We might as well start that in a new PR.

PR #494, opened 2023-12-22
automod orga

warpfork:
Several more package extractions.

These are all the easy ones -- flagstore, cachestore, setstore, directory -- all get yanked out in the same way as countstore already was.

These didn't cause any major changes to package graph overall, it just makes the main automod package smaller (yay).  Near future targets are to try to tug engines and events a bit further apart, and that might be more complicated, so I thought I'd put up the easy-to-review diffs first.


bnewbold (approved):
nice!

PR #495, opened 2023-12-23
lexgen update

bnewbold:
As a temporary hack, just skip the `com.atproto.temp.importRepo` endpoint which uses Lexicon features that lexgen does not support.

There are a small number of actual lexicon updates, but the main motivation here is to handle *all* of the `didDoc` and `plcOp` fields. That is also a hack for now, but this resolves a number of regressions, eg with refreshSession.


ericvolp12 (approved):
LGTM

Issue #378, opened 2023-10-06
tracking SDK progress (`indigo/atproto/*`)


bnewbold:
Sketching out what a more stable "golang atproto SDK" might look like.

- [x] `atproto/syntax`: syntax types for Handle, DID, NSID, AT-URI, TID
- [x] `atproto/crypto`: basically go-did pulled in to this monorepo, and DID-specific stuff removed. smaller API surface specifically for atproto, abstracting away most implementation details and config knobs
- [x] `atproto/identity`: combined handle and DID resolution. interface for caching/persistent services
- [ ] `atproto/data`: runtime helpers for data model, possibly using `go-ipld-prime`. eg, for working with records and JSON when you don't know the schema. DAG-CBOR to JSON conversion.
- [ ] `atproto/lexicon`: generic Lexicon helpers, and types for dealing with Lexicon schema JSON itself. likely runtime schema verification. the sort of utils which get referenced by codegen'd code
- [ ] `atproto/repo`: MST and repository implementation, including CAR import/export, diff processing, verification, interfaces
- [ ] `atproto/client`: generic HTTP client ("XRPC"), with ergonomic auth help. possibly generic stream (WebSocket) client helpers also. no codegen stuff for specific Lexicons.

Codegen tooling and codegen output (record types, client APIs, server handlers) would probably live elsewhere, at least for a while.

PR #496, opened 2023-12-23
hepa: periodically refresh auth session

bnewbold:
This works around a current issue in prod where actions are failing to persist because the session expires.

The real fix for the underlying issue (IMHO) is to re-implement `xrpc.Client` to handle this in a more elegant and robust way. That is sort-of tracked in https://github.com/bluesky-social/indigo/issues/378

I was initially confused why this was only coming up when taking actions, because the client is used for a number of reads (GET) as well. Those GETs mostly use *admin* auth (a token in the HTTP Authorization header), not login/session auth (which requires refresh), which explains why it mostly works, then fails for actions.

### DIFF
--- a/cmd/hepa/main.go
+++ b/cmd/hepa/main.go
@@ -192,6 +192,14 @@ var runCmd = &cli.Command{
 			}
 		}()
 
+		if srv.engine.AdminClient != nil {
+			go func() {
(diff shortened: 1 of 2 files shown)


bnewbold:
I'm going to merge this as well, to get out a deploy on top of the context refactor.

This PR is really just a hack around `xrpc.Client` not (yet) supporting automatic session refresh.

PR #497, opened 2023-12-23
remove deprecated `api/label` code

bnewbold:
This is a cleanup to unwind some temporary types which were used for labelmaker development before the label Lexicons were merged upstream.

This is just confusing cruft now, so ripping it out.


ericvolp12 (approved):
Sure

PR #499, opened 2023-12-23
redis-cached identity directory

bnewbold:
This exact code is already in the repo, being used for automod. I'm doing this PR for review from @whyrusleeping and/or @ericvolp12, and any code cleanup or API changes. After this merges, I will update automod to use this package in a separate PR.

Thee are no tests yet. Could probably use https://github.com/go-redis/redismock plus the exiting directory mock for testing.


atproto/identity/redisdir/promethus.go
```diff
@@ -0,0 +1,36 @@
+package redisdir
```
ericvolp12, line 1: My convention has been to name this file `metrics.go`

ericvolp12 (approved):
LGTM, haven't had the chance to use that redis cache library before but it looks neat

atproto/identity/redisdir/promethus.go
bnewbold: you didn't like "promethus" :sweat_smile:

PR #500, opened 2023-12-24
automod context refactor

bnewbold:
Builds on warpfork's progress, and our co-working a couple days back.

Changelog:

- all state mutations in a single dedicated `Effects` struct
- consolidated record event metadata into a single RecordOp struct. intended to be immutable (kind of like AccountMeta, but with less fetched context). aligns functionally with repoOp from the atproto firehose schema, but has somewhat different syntax. encompasses any of create, update, or delete ops on a single record
- removed "*Event" and effectively replaced with "*Context". Didn't have immutable "Event" as sub-fields on Context, but the AccountMeta and RecordOp are basically that. engine and effects are (intentionally) private fields. `*Context` is the API exposed to rule functions
- cherry-picked an unrelated API/schema hotfix patch in to this branch
- merged effects package into engine
- split out capture (and "fetch") code to a package
- moved engine-specific tests into the engine package
- re-export some types from the top-level `automod` package
- nuked the `automod/util` package, for now (just copied around dedupeStrings)
- cleaned up some code duplication in tests (just use ProcessRecordOp instead of duplicating code)

Questions for review:

- should canonical-log-line be a method on Engine, or Context, or neither? feels like rules should not have access to this method. I suspect the logic around notifications and dev-observability will get more sophisticated soon, like only logging or notifying in slack for specific effects. the effects split-out should help a lot with this, and simplifying the existing "persist" methods
- should "effects" field be exposed ("Effects") on Context, and methods called directly on that? I decided not for now, and put thin wrapper methods directly on Context. I think we've achieved the state/mutation split in implementation/internals, and rule-writers don't need to know or think about this distinction
- we *could* hide more fields on the Context structs and make them accessible only via methods (`c.Account().Identity` instead of `c.Account.Identity`). I think for the "core" metadata it is fine as fields, even though this locks us in API-wise. maybe things like the admin "Private" metadata (which may or may not be present) should be behind a method call.
- probably should do a review of expected behaviors around partial-processing (eg, if an error is encountered mid-persist) and ensure that is sane and documented

Current state: all tests updated and passing. README updated. all the docs/comments likely need a review and update, but I think that can be a separate PR. This feels pretty good and my disposition is to merge this pretty soon, as the branch has gotten large and will cause merge conflicts (we knew this when we started). the last thing I want to do is testing in staging and against prod firehose (from my laptop).


warpfork:
Awesome.

Very much to review here (although I also suspect it will get less daunting once I defocus my eyes enough to see through all the moves that look like changes).

I like lots and lots of things here :D

I will also say just from the description: on "split out capture (and "fetch") code to a package" -- very happy for that too!

bnewbold:
I'm going to merge this now, to unblock progress on other features/branches. Would still appreciate comments, and very open to iterating on all these changes in the near future.

automod/engine/engine.go
```diff
@@ -0,0 +1,169 @@
...
+	return e.Sets.InSet(context.TODO(), name, val)
+}
+
+// purge caches of any exiting metadata
+func (e *Engine) PurgeAccountCaches(ctx context.Context, did syntax.DID) error {
+	e.Directory.Purge(ctx, did.AtIdentifier())
+	return e.Cache.Purge(ctx, "acct", did.String())
+}
+
+func (e *Engine) CanonicalLogLineAccount(c *AccountContext) {
```
warpfork, line 141: Agree that `CanonicalLogLineAccount` makes sense to have as an engine method since there's no reason to let other code control (and thus be able to forget to) call it.  The other argument would be that it mostly sees fields of effects, but I think log methods being closer to their use than their subject makes sense.

I think it can even be private?  There appears to be one caller, and it is also on the engine.

automod/engine/context.go
```diff
@@ -0,0 +1,184 @@
...
+	effects Effects
+}
+
+type AccountContext struct {
+	BaseContext
+
+	Account AccountMeta
+}
+
+type RecordContext struct {
```
warpfork, line 31: Okay so I'm definitely not complaining if there's these two types specializing the different kinds of business context.  But, I am curious.

The effects type contains "extra" fields that only apply to record event processing and are inapplicable to account event processing... and there, we just say those stay empty/nil.  Here, we make more types.

I don't think there's anything bad in this asymmetry, but is there a reason behind it worth documenting or is it just incidental?  Do we think it'll continue to make sense?

warpfork: The symmetric more-types alternative is probably `AccountEffects` + `RecordEffects` (embeds AccountEffects), and then `AccountContext` embeds AccountEffects and `RecordContext` embeds RecordEffects.  (I can't especially think of a reason for a distinct "base" type for effects.)

The symmetric less-types alternative is just having one business `Context` type here and having RecordOp and AccountMeta fields that are documented to have a possibility of being nil depending on the, uh, context.

automod/engine/context.go
```diff
@@ -0,0 +1,184 @@
...
+func NewRecordContext(ctx context.Context, eng *Engine, meta AccountMeta, op RecordOp) RecordContext {
+	ac := NewAccountContext(ctx, eng, meta)
+	ac.BaseContext.Logger = ac.BaseContext.Logger.With("collection", op.Collection, "rkey", op.RecordKey)
+	return RecordContext{
+		AccountContext: ac,
+		RecordOp:       op,
+	}
+}
+
+// update effects (indirect)
```
warpfork, line 141: I'd vote weakly against these proxy methods and prefer exposing the Effects struct (either as `Effects` a field, or `Effects()` a getter; either works).   Mainly because I'm a fan of the guidance/focus that autocomplete can give if I start with `ctx.Effects.` whenever I'm looking for a mutation method and I can't remember the name.

Field or method doesn't feel like it matters much.  The main thing exposing it as a field directly changes is it makes it zeroable by out-of-package code.  If we were in a language with a stronger typesystem, I'd say that's worse, but in golang, I'd say it's par.

automod/README.md
```diff
@@ -30,27 +30,29 @@ Here is a simple example rule, which handles creation of new events:
...
+		c.AddRecordLabel("spam")
 	}
 	return nil
 }
 ```
 
 Every new post record will be inspected to see if it contains a static test string. If it does, the label `spam` will be applied to the record itself.
 
-The `evt` parameter provides access to relevant pre-fetched metadata; methods to fetch additional metadata from the network; a `slog` logging interface; and methods to store output decisions. The runtime will catch and recover from unexpected panics, and will log returned errors, but rules are generally expected to run robustly and efficiently, and not have complex control flow needs.
+The `c` parameter provides access to relevant pre-fetched metadata; methods to fetch additional metadata from the network; a `slog` logging interface; and methods to store output decisions. The runtime will catch and recover from unexpected panics, and will log returned errors, but rules are generally expected to run robustly and efficiently, and not have complex control flow needs.
```
warpfork, line 43: I'll understand if this doesn't feel worth changing, but I probably would advocate for `ctx` here over just `c`.  I'm not really a fan of single-letter vars (even if it's acclaimedly gopheric).  I think it's also fairly common to reuse the "ctx" identifier among things that have "business context" patterns -- for ex, [urfave CLI](https://pkg.go.dev/github.com/urfave/cli#Command.Run) did this in their docs conventions.

(I see [echo context](https://echo.labstack.com/docs/context) went the other way and kept to a single letter, so it's certainly a matter of preference.)

warpfork:
I got time now to go over in a little higher resolution and left some comments, but I'm very happy with this being merged :)

automod/engine/context.go
bnewbold: My thinking on this is that the Context types are a public API, supporting mutation, and it is nice to use types make only sensical operations possible. The complexity/burden of extra types is worth it.

Effects on the other hand are mostly internal, and at least downstream of rule execution are mostly read-only. Code can just read the fields it cares about. The complexity/burden of additional types not worth it. In particular i'd like to refactor a bit more to have things like de-dupe, circuit-break, and persist mod actions be simple functions working on Effects, not split account/record versions (might want to split that big function up in other ways, like labels vs reports).

Effects is currently a public API, mostly just to allow tests to inspect it. Maybe would be less concerning if it wasn't public?

automod/engine/context.go
bnewbold: One thing I don't like about direct Effects access is splitting things like `GetCount` and `Increment` across two types. Feels like that splits documentation/awareness, though it does (maybe) communicate what methods have effects and which don't.

automod/README.md
bnewbold: Hmm! We use `cctx` for urfave CLI context, fwiw. (I didn't make that decision, just copied existing code). I think I might find it a bit confusing and would lean towards sticking with `c`? But not a tightly-held opinion if you have stronger feelings.

automod/engine/engine.go
bnewbold: :+1: to having it be private, didn't think of that.

A possible refactor will be to have an "alerts" or "notifications" component/interface on engine, which gets called after rule processing to handle things like sending Slack messages or logging more loudly for some cases. I *think* that we'll always want basic logging ("canonical log line") to happen regardless though, and can be done by the engine.

PR #505, opened 2023-12-27
remove three unnecessary direct dependencies

mvdan:
(see commit messages - please do not squash)


util/cliutil/util.go
```diff
@@ -54,7 +53,8 @@ type CliConfig struct {
 }
 
 func readGoskyConfig() (*CliConfig, error) {
-	d, err := homedir.Dir()
+	// TODO: use os.UserConfigDir()/gosky, falling back to os.UserHomeDir()/.gosky for backwards compatibility.
```
bnewbold, line 56: thanks for noting this, and not including a behavior change in this PR

bnewbold (approved):
This is wonderful! Thanks for contributing this kind of maintenance.

PR #506, opened 2023-12-27
gofmt -s and some staticcheck work

mvdan:
(see commit messages - please do not squash)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ coverage-html: ## Generate test coverage report and open in browser
 .PHONY: lint
 lint: ## Verify code style and run static checks
 	go vet -asmdecl -assign -atomic -bools -buildtag -cgocall -copylocks -httpresponse -loopclosure -lostcancel -nilfunc -printf -shift -stdmethods -structtag -tests -unmarshal -unreachable -unsafeptr -unusedresult ./...
-	test -z $(gofmt -l ./...)
+	test -z $(gofmt -s -l ./...)
 
 .PHONY: fmt
 fmt: ## Run syntax re-formatting (modify in place)
--- a/atproto/crypto/cmd/atp-crypto/main.go
+++ b/atproto/crypto/cmd/atp-crypto/main.go
@@ -16,7 +16,7 @@ func main() {
 		Usage: "informal debugging CLI tool for atproto key and cryptography",
 	}
 	app.Commands = []*cli.Command{
-		&cli.Command{
+		{
 			Name:  "generate",
 			Usage: "create a new private key",
 			Flags: []cli.Flag{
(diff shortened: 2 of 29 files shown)


mvdan:
FWIW I would definitely recommend following `go vet ./...` and `staticcheck ./...`. There's quite a bit of work to appease staticcheck, but from the sample I chose above, you can hopefully see it's worth it.

bnewbold:
This one is probably a bit harder to merge as-is. Do really appreciate the contribution.

Some notes:
- 

- `go vet ./...` (vanilla instead of the current tweaked settings): I support this enthusiastically. @whyrusleeping wanted to skip some lints early on. I think there are really just a handful of structs where this comes in to play. maybe we can just cut over now?
- `staticcheck ./...`: as a generally thing i'm in favor of more static analysis. getting towards `golangci-lint` would be great.
- `comatproto` vs `atproto` package name: I'm not-so-secretly hoping to change the package names to `api/comatproto` and `api/appbsky` at some point, to align with the full lexicon names, and it will be an easier refactor if we keep the import aliases as they are. `comatprototypes` should go though.
- removing all the explicit struct types, I assume via `gofmt -s`: seems like a style decision? I'm not sure.
- some of the other "simplify" things, like shorter appends, are fine but lower-priority right now

There seem to be some real fixes in here w/r/t unhandled `err` and things like that, it would be nice to have bugfix commits (with behavior changes) in a separate PR from style and lint commits (no behavior changes).

bnewbold:
By commit:

- `all: gofmt -s `: I would probably pass on this for now
- `all: fix a number of staticcheck warnings`: mixed bag. avoiding for-loops is good, most other things good; renaming imports I would mostly pass on for now
- `all: fix a number of unused values`: this looks very valuable. I'd want another set of eyes on review for the bgs and carstore stuff, but would almost certainly merge

mvdan:
`gofmt -s` removes redundant syntax, and is a pretty standard thing from upstream Go. Some projects only enforce `gofmt`, but many enforce `gofmt -s` as well, and I honestly don't see a drawback. Up to you of course.

Renaming imports up to you - it sounds like you're in the middle of rethinking import paths. Removing that part.

Note that staticcheck and golangci-lint are pretty different tools :) The former is like an extension of vet, with high quality checks that aim for very low false positives. golangci-lint is a collection of linters with a big range of quality, speed, and purpose. I'd personally stick to gofmt+vet+staticcheck in general.

mvdan:
@bnewbold any suggestion on next steps, before this gains merge conflicts?

mvdan:
Already gained some conflicts :) friendly bump, happy to solve them

mvdan:
A large amount of conflicts at this point. I'll leave it to you to figure out whether you want to obey these tools - I very much think you should.

PR #513, opened 2023-12-31
pds: remove `app.bsky.*` endpoints

bnewbold:
This code is unused. These routes are not served from the PDS, they get served from the AppView.

In theory the implementation might be useful for a future golang appview; we can grab the code from git history if needed.


ericvolp12 (approved):
lgtm

PR #520, opened 2024-01-13
automod visual auto-labeling, and abusive image scanning

bnewbold:
This is just porting over existing auto-labeling stuff from appview in to the automod framework.

New engine-level supporting features:

- new "blob" rule function type
- ruleset fetches blobs from upstream PDS, if necessary
- `Effects` can track blob-level takedowns (CDN purges) as part of record takedowns

### DIFF
--- a/automod/engine/blobs.go
+++ b/automod/engine/blobs.go
@@ -1,0 +1,115 @@
+package engine
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+
+	appbsky "github.com/bluesky-social/indigo/api/bsky"
+	lexutil "github.com/bluesky-social/indigo/lex/util"
+)
+
+// Parses out any blobs from the enclosed record.
+//
+// TODO: currently this function uses schema-specific logic, and won't work with generic lexicon records. A future version could use the indigo/atproto/data package and the raw record CBOR to extract blobs from arbitrary records
+//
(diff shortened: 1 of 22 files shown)


bnewbold:
This is ready for review cc: @warpfork 

We are looking to get this deployed relatively soon, and I did at least partial testing on a VM against full firehose, so I may self-merge soon, but retroactive review still requested.

Implementation-wise, I think the most important thing to look at is the nested WaitGroup concurrency pattern. The specific abyss and hive rules are directly based on pre-existing rules in prod, and don't need as much review. We intentionally should not change the policy/behavior of the rules for now (while doing this transition from typescript code to golang code), but will probably iterate on that in the future.

Design-wise, a couple things come up that we could potentially refactor:

- this arch fetches all the blobs pre-emptively (as long as there are any blob rules), and pushes the blob bytes in to the rule. there is no way to select a subset of blobs or anything like that. alternatively, rules could have requested blob bytes from, eg, RecordContext, which would download-and-cache on demand, using the Engine. feels like a toss-up.
- thought about splitting out ImageRuleType from BlobRuleType, but want to keep a lid on rule type proliferation
- hacked in a basic form of ruleset selection in hepa. I think we'll probably end up with more sophisticated ruleset control, eg dynamic runtime re-configuration of which rules are enabled (feature-flag style or an HTTP API)

bnewbold:
I'm going to merge this now, to prevent merge conflicts, but not enable the new rules when deployed.

It would be good to get review, and there are some details like making Effects thread-safe that probably need to land before this is run for real.

PR #527, opened 2024-01-22
automod: notification rules/processing

bnewbold:
This is prep work to get automod inserted in to our push notification service as a library.

Relatively small PR, which is great.

Open questions:
- what should response to the notification service look like? here I proposed just a boolean (false for no change, true to reject the message)
- what other context and info is available? working with the original notification types might be good, though they only really exist in "view" form (with extra hydrated metadata), not as a record or something like that. probably at least the "Reason" and an AT-URI reference?

### DIFF
--- a/automod/engine/context.go
+++ b/automod/engine/context.go
@@ -35,6 +35,13 @@ type RecordContext struct {
 	// TODO: could consider adding commit-level metadata here. probably nullable if so, commit-level metadata isn't always available. might be best to do a separate event/context type for that
 }
 
+type NotificationContext struct {
+	AccountContext
+
+	Recipient AccountMeta
+	// TODO: more context about the notification? "Reason", "Subject"
+}
+
 var (
 	CreateOp = "create"
 	UpdateOp = "update"
@@ -138,6 +145,14 @@ func NewRecordContext(ctx context.Context, eng *Engine, meta AccountMeta, op Rec
 	}
 }
 
+func NewNotificationContext(ctx context.Context, eng *Engine, sender, recipient AccountMeta) NotificationContext {
+	ac := NewAccountContext(ctx, eng, sender)
+	return NotificationContext{
+		AccountContext: ac,
+		Recipient:      recipient,
+	}
+}
+
 // update effects (indirect)
 func (c *BaseContext) Increment(name, val string) {
 	c.effects.Increment(name, val)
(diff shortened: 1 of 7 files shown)


bnewbold:
Discussed with other folks and answers are:

- we'll get "reason" (string) and "subject" (AT-URI) in addition to sender/recipient DIDs
- boolean is fine for now

automod/engine/context.go
```diff
@@ -54,6 +56,19 @@ type RecordOp struct {
...
+// Originally intended for push notifications, but can also work for any inter-account notification.
+type NotificationContext struct {
+	AccountContext
+
+	Recipient    AccountMeta
+	Notification NotificationMeta
+}
+
+type NotificationMeta struct {
+	Reason  string
```
warpfork, line 68: You're on a roll with docs, don't drop it now! :D  What does this string typically look like, who provides it, and what sort of things should it affect?

(Or document that on the `NewNotificationContext` constructor and point at that from the struct comment.)

automod/engine/effects.go
```diff
@@ -51,6 +51,8 @@ type Effects struct {
 	RecordTakedown bool
 	// Set of Blob CIDs to takedown (eg, purge from CDN) when doing a record takedown
 	BlobTakedowns []string
+	// If "true", indicates that a rule indicates that the action causing the event should be blocked or prevented
+	RejectEvent bool
```
warpfork, line 55: I'll comment on vibe, but I don't have a change request: This makes Effects feel a little grab-bag'y, because I don't think the rest of the labels and reports stuff is valid where this bool is relevant, and vice versa.

But I don't think I have a serious complaint about that because I don't know how else I'd write it.  (Introducing a smorgasboard of types for different XEffects per XContext might be an option, but I don't think it would make our lives significantly easier.  And it would also directly align our shins with the familiar pain point of "it would be nice if golang could represent sum types".)

warpfork (approved):
LGTM so far!

automod/engine/effects.go
bnewbold: yeah, that checks out and aligns with my feelings on this struct. It is a bit big and amorphous in scope.

On the other hand it is partially obscured/internal, at least from rules. and all the fields are documented. and I *really* want to avoid an explosion of variant types.

PR #528, opened 2024-01-23
automod: a bit more docs/comments

bnewbold:
Mostly old commits, re-based.

### DIFF
--- a/automod/engine/ruleset.go
+++ b/automod/engine/ruleset.go
@@ -8,6 +8,7 @@ import (
 	lexutil "github.com/bluesky-social/indigo/lex/util"
 )
 
+// Holds configuration of which rules of various types should be run, and helps dispatch events to those rules.
 type RuleSet struct {
 	PostRules         []PostRuleFunc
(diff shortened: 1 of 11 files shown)


automod/engine/ruleset.go
```diff
@@ -17,6 +18,7 @@ type RuleSet struct {
 	BlobRules         []BlobRuleFunc
 }
 
+// Executes all the various record-related rules. Only dispatches execution, does no other de-dupe or pre/post processing.
```
warpfork, line 21: fwiw, some (relatively popular) golang linters will complain if the docs don't start with literally the name of the method (per the golang convention that docs should read as sentences and thus begin with "Foobar is a [...]").

I don't particularly care about this, myself, but it is a common linter.

warpfork (approved):
Yay for docs :+1:

automod/engine/ruleset.go
bnewbold: good to know! I don't love that, but open to it in the future

PR #529, opened 2024-01-23
engine: remove wrapper counter and set methods

bnewbold:
These were from earlier API design, and are just confusing now.


warpfork (approved):
Generally thumbs up to more dots in the path to things :+1:

PR #531, opened 2024-01-24
automod: iterate on error handling

bnewbold:
Saw a bunch of errors in prod, stemming from some federation tests and failure to fetch account metadata (eg, profile).

We should support partial-success and partial-processing in most cases, except for internal errors. This updates a lot of error throws to log-and-continue, and updates a bunch of remaining error throws to have better descriptions.


bnewbold:
confirming that the underlying issue I was seeing had to do with looking up private metadata from a PDS instance in prod which is one of our testing ones, and automod didn't have permissions to that metadata. the problem was that automod didn't fail gracefully in that situation; with this branch, it is more graceful and core rule execution continues to work.

PR #534, opened 2024-01-25
identity/redisdir: various fixes

bnewbold:
A bunch of cleanups and bug fixes.

The biggest fix is to apply a bug fix from the cache directory resulting in "nil/nil" responses. These were fixed in https://github.com/bluesky-social/indigo/pull/479, but not copied to the redis directory implementation, which has the same coalescing code. This recently impacted @whyrusleeping's use in discover. This is a backwards-compatible fix.

Another fix handles cached error messages. An empty string was being stored for the DID. Because redis cache serializes, and the de-serialized DID is re-parsed, the empty strings fail to parse, resulting in an error. I guess we could consider having text parsing of DIDs not actually parse the string? but that would break auto-validation when it is expected. The fix was to make that field nullable. this might make the change not-backwards-compatible with existing cached metadata? this only impacts cached errors, which have a shorter TTL, so might not be that big of a deal in practice?

Other smaller cleanups and fixes, which rolled in here to make the code align better with the in-memory caching directory, so that merge/review is easier:

- adds invalid handle TTL to redisdir
- adds "WithCacheState" variants of methods, for metrics/perf measurement
- fix deletion/purging
- special-case handling of resolving invalid handle

One remaining issue is that the cached error types come back as generic errors (with error kind/type striped). It would be nice if that wasn't the case, so that code could detect specific types of errors? Considering doing a manual enum/integer hack to encode the type and convert the errors to the actual type on read, for a small set of known identity errors.

### DIFF
--- a/atproto/identity/redisdir/redis_directory.go
+++ b/atproto/identity/redisdir/redis_directory.go
@@ -116,6 +116,7 @@ func (d *RedisDirectory) updateHandle(ctx context.Context, h syntax.Handle) hand
 			he.Err = err
 			return he
 		}
+		return he
 	}
 
 	entry := identityEntry{
--- a/atproto/identity/redisdir/redis_directory.go
+++ b/atproto/identity/redisdir/redis_directory.go
@@ -162,9 +162,15 @@ func (d *RedisDirectory) ResolveHandle(ctx context.Context, h syntax.Handle) (sy
 	if err != nil && err != cache.ErrCacheMiss {
 		return "", fmt.Errorf("identity cache read: %w", err)
 	}
-	if err != cache.ErrCacheMiss && !d.isHandleStale(&entry) {
+	if nil == err && !d.isHandleStale(&entry) {
 		handleCacheHits.Inc()
-		return entry.DID, entry.Err
+		if entry.Err != nil {
+			return "", entry.Err
+		} else if entry.DID != nil {
+			return *entry.DID, nil
+		} else {
+			return "", fmt.Errorf("code flow error in redis identity directory")
+		}
 	}
 	handleCacheMisses.Inc()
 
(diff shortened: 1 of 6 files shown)


brianolson (approved):
I think I get it, mostly improving error handling. same quibble that I prefer `errors.New("")` when the string doesn't need formatting params

atproto/identity/redisdir/redis_directory.go
```diff
@@ -139,20 +149,31 @@ func (d *RedisDirectory) updateHandle(ctx context.Context, h syntax.Handle) (*ha
...
+		return "", fmt.Errorf("invalid handle")
+	}
 	var entry handleEntry
 	err := d.handleCache.Get(ctx, redisDirPrefix+h.String(), &entry)
 	if err != nil && err != cache.ErrCacheMiss {
-		return "", err
+		return "", fmt.Errorf("identity cache read: %w", err)
 	}
-	if err != cache.ErrCacheMiss && !d.isHandleStale(&entry) {
+	if nil == err && !d.isHandleStale(&entry) {
```
whyrusleeping, line 168: `err == nil`

atproto/identity/redisdir/redis_directory.go
```diff
@@ -256,49 +295,58 @@ func (d *RedisDirectory) LookupDID(ctx context.Context, did syntax.DID) (*identi
 			// The result should now be in the cache
 			err = d.identityCache.Get(ctx, redisDirPrefix+did.String(), &entry)
 			if err != nil && err != cache.ErrCacheMiss {
-				return nil, err
+				return nil, false, fmt.Errorf("identity cache read: %v", err)
 			}
-			if err != cache.ErrCacheMiss && !d.isIdentityStale(&entry) {
-				return entry.Identity, entry.Err
+			if nil == err && !d.isIdentityStale(&entry) {
```
whyrusleeping, line 300: `err == nil`

atproto/identity/redisdir/redis_directory.go
```diff
@@ -317,11 +365,19 @@ func (d *RedisDirectory) Purge(ctx context.Context, a syntax.AtIdentifier) error
 	handle, err := a.AsHandle()
 	if nil == err { // if not an error, is a handle
```
whyrusleeping, line 366: why is this backwards everywhere? is this a rust thing?

ericvolp12: I think Bryan does this intentionally to make it clear it's a `==` and not a `err != nil`

bnewbold:
Resolved the review things. My disposition is to merge this, but I want to monitor when this gets deployed (in various places), so i'll probably wait until i'm ready to do a controlled deploy and merge it then.

PR #542, opened 2024-01-26
HACK to do abyss-only blob processing temporarily

bnewbold:
DO NOT MERGE

This is to exercise automod more in prod with real rules.


bnewbold:
this is redundant now, handled via an extra ruleset option

PR #546, opened 2024-01-29
Specify key multiple times in values when providing a string slice

ericvolp12:


bnewbold:
This does seem correct to me (the idiomatic way to do multi-value URL query params), but i'm a bit worried the old behavior might have been relied on in some places? Eg, relay admin endpoints or something?

cc: @whyrusleeping

PR #553, opened 2024-02-04
lexgen update (2024-02-03)

bnewbold:
These are pretty simple lexicons, and it seems like lexgen output is clean right now: I don't think this clobbered any manual edits.

Motivation is to use `getRelationships` in automod.


ericvolp12 (approved):
lgtm

PR #559, opened 2024-02-06
automod: ratelimit bypass header

bnewbold:


bnewbold:
self-merging this (working fine)

PR #560, opened 2024-02-06
hepa: use atomics for lastSeq.

warpfork:
Fixes data race.  Was probably harmless in practice, as these values are best-effort anyway, but keeping the race detector clean and happy when engaged is highly valuable for maintainability in the long run.

Some comments attached about the field's concurrency.


bnewbold (approved):
Nice! This looks good. I might hold off on merging for a couple days.

PR #570, opened 2024-02-12
Update `blueskyweb.xyz` links in README.md to `bsky.social`

surfdude29:
This PR updates the two links in README.md that point to `blueskyweb.xyz` to the equivalent `bsky.social` URLs.

Also, I noticed that Line 13 still refers to 'Big Graph Service', should this be updated to 'Relay'?

https://github.com/bluesky-social/indigo/blob/c72ce2340db3cc99fc886c2284daf19e438c7b91/README.md?plain=1#L13


bnewbold:
thanks!

yup, probably should be "Relay"

PR #577, opened 2024-02-14
:skull:  remove labelmaker and labeler code

bnewbold:
pour one out for this big pile of code, mostly from last april.

this still does something a bit unique (basically combined automod plus mod service), but it is only half-complete, not maintained, and just dead code causing refactor friction. best to just jetison and we can pull bits out of git history if needed.


ericvolp12 (approved):
die bart die

Issue #578, opened 2024-02-14
Search: `ParseQuery` usage of shlex causes inability to search for "#"


kisaragi-hiu:
The `ParseQuery` function in search uses [shlex](https://github.com/google/shlex/) to split the search query:

https://github.com/bluesky-social/indigo/blob/a30de45cd80176c16d826108914e8700d9ece8fd/search/parse_query.go#L16-L18

shlex splits input as shell syntax, including comments. `shlex.Split("abc foo #def ghi")` produces `["abc", "foo"]`, with everything after (and including) the `#` thrown out, even though the expectation here is probably `["abc", "foo", "#def", "ghi"]`.

This causes https://github.com/bluesky-social/social-app/issues/2491. It makes the query "#a_tag" equivalent to "" (empty string), and `from:username.bsky.app #art` equivalent to `from:username.bsky.app`.

Zero3K:
This can be closed now since searching for text with # at the front works.

PR #579, opened 2024-02-15
Parse queries better for search

ericvolp12:
Addresses https://github.com/bluesky-social/indigo/issues/578


bnewbold:
thank you for cleaning up my junk code @ericvolp12!

Issue #581, opened 2024-02-16
Searching for symbol and word should only show results that have them and in them, not that and just the word


Zero3K:
*subject*

That way it will be like how Twitter does it.

Zero3K:
This should be easy to fix.

RDxR10:
Adding to it, its not just "#" but other symbols as well.

Clarification:

![search_@](https://github.com/bluesky-social/indigo/assets/43957261/84cc7b7a-9226-4214-9e29-ed8b793b9291)

Here, since my query is "@youtube" so I am expecting to get the symbol including the word "youtube" (case insensitive).

Also I don't see the new posts containing morphemes of a word/non-word. 

For example: I searched for aQi. But did not get the latest post which had "aQi"

 
![aQi](https://github.com/bluesky-social/indigo/assets/43957261/7ec6ea85-256f-4aa4-8fbd-2975cf6e8df6)

![aQi__](https://github.com/bluesky-social/indigo/assets/43957261/265a7d8d-c8a8-42b7-8573-6f84784c0f2f)

Similar search: for "xkc"

![xkc](https://github.com/bluesky-social/indigo/assets/43957261/8f7498e4-7bfc-47a5-a277-11f9b818ed53)

![xkc__](https://github.com/bluesky-social/indigo/assets/43957261/3b103263-703f-4d94-9d6b-ff1d7b62653a)

Zero3K:
Will it be fixed in a commit to https://github.com/bluesky-social/social-app/tree/eric/tags?

Zero3K:
Well?

Zero3K:
Searching for # and word now only shows the results that have both of them. @RDxR10 You will need to make issues in https://github.com/bluesky-social/social-app/issues.

PR #582, opened 2024-02-16
Support facet tags for search

ericvolp12:


Zero3K:
Does this fix https://github.com/bluesky-social/indigo/issues/581?

ericvolp12:
This is the start of getting hashtag searching for the tag field within a post (the actual hashtag implementation).

Not gonna merge anything here for a bit but just poking around with the query constructor for now.

Zero3K:
Tag field? Like Tumblr?

ericvolp12:
The post record type supports tags (like hastags) in Facets as well as in an additional field in the post record, these aren't fully implemented in our client yet but will be eventually - https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/feed/post.json#L50

Zero3K:
Or will it just be like how Twitter does it?

PR #585, opened 2024-02-21
automod: defer parsing

bnewbold:
This shifts the external API for automod from taking a pre-parsed "any" struct of a record, to taking CBOR bytes. the motivation is to make it easier to work with content for which the lexicon isn't know at compile time, which will soon become much more common as we allow folks to write arbitrary record types to their repositories, and this content shows up on the relay firehose.

The `atproto/data` package knows how to parse and extract some information from generic atproto record CBOR, so it is able to pull all the blobs (`$type: blob` objects nested within records), and we can switch over to that, which is more generic (and hopefully robust) than doing schema-by-schema extraction.

For firehose content, this is all pretty straight-forward, because the records are already coming in as CBOR. in the capture system, we now do schema-specific parsing, then re-encode as CBOR from the parsed struct, which is kind of weird. and in tests records need to be marshaled to CBOR, which is a bit unfortunate. maybe some helpers would make this more idiomatic?

tested a small amount against prod firehose.

### DIFF
--- a/automod/engine/context.go
+++ b/automod/engine/context.go
@@ -52,7 +52,7 @@ type RecordOp struct {
 	Collection syntax.NSID
 	RecordKey  syntax.RecordKey
 	CID        *syntax.CID
-	RecordCBOR *[]byte
+	RecordCBOR []byte
 }
 
 // Originally intended for push notifications, but can also work for any inter-account notification.
--- a/automod/engine/context.go
+++ b/automod/engine/context.go
@@ -52,8 +52,7 @@ type RecordOp struct {
 	Collection syntax.NSID
 	RecordKey  syntax.RecordKey
 	CID        *syntax.CID
-	// NOTE: usually a *pointer*, not the value itself
-	Value any
+	RecordCBOR *[]byte
 }
 
 // Originally intended for push notifications, but can also work for any inter-account notification.
@@ -76,11 +75,11 @@ type NotificationMeta struct {
 func (op *RecordOp) Validate() error {
 	switch op.Action {
 	case CreateOp, UpdateOp:
-		if op.Value == nil || op.CID == nil {
+		if op.RecordCBOR == nil || op.CID == nil {
 			return fmt.Errorf("expected record create/update op to contain both value and CID")
 		}
 	case DeleteOp:
-		if op.Value != nil || op.CID != nil {
+		if op.RecordCBOR != nil || op.CID != nil {
 			return fmt.Errorf("expected record delete op to be empty")
 		}
 	default:
--- a/automod/capture/fetch.go
+++ b/automod/capture/fetch.go
@@ -47,7 +47,7 @@ func FetchAndProcessRecord(ctx context.Context, eng *automod.Engine, aturi synta
 		Collection: aturi.Collection(),
 		RecordKey:  aturi.RecordKey(),
 		CID:        &recCID,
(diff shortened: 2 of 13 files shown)


automod/engine/context.go
```diff
@@ -52,8 +52,7 @@ type RecordOp struct {
 	Collection syntax.NSID
 	RecordKey  syntax.RecordKey
 	CID        *syntax.CID
-	// NOTE: usually a *pointer*, not the value itself
-	Value any
+	RecordCBOR *[]byte
```
warpfork, line 55: This can just be `[]byte` rather than `*[]byte`.

I'm guessing the extra pointer appeared out of a habit of using pointers as a stand-in for `Optional<T>`, but... It's not gopheric.  :)

(Okay, admittedly, making distinctions between `nil` and `[]T{}` is also considered non-gopheric.  But in cases where one really wants to use `nil` as a distinction, I think having fewer unnecessary pointers is the lesser of evils.)

(The only time one needs a `*[]T` is if one needs to be able to replace the variable's content with a whole new slice allocation, and _has no other handle_ to it.  Sometimes, fairly rarely, one ends up with a function that takes a `*[]T` parameter because it wants to be able to grow that slice (without having an API involving returning a new one and leaving the caller to handle the assignment).  But if the thing is in a struct, like this... you _already_ have another updatable handle to it, so that situation doesn't apply.  (And since this is supposed to be "immutable", it doubly doesn't apply!))

warpfork (approved):
Yeah, this broadly makes sense to me!

---

To do some principles recap as I see it: there's roughly two major ways to go about deferred parsing:

- Parse things into a full (but relatively generic) AST -- no structure enforcement, so you keep everything, but also involves a decent amount of allocations to make the AST a, well, "T".
- Keep the original blob and build a lex offset tree into it.  (Much fewer allocations, but also several steps further shy from directly helpful accessor APIs.)

This set of diffs is moving towards route 2.

And I like route 2, yeah.

---

Keeping a `[]byte` slice instead of a `bytes.Buffer` or other interface -- yeah, that's also the right way to go, IME.  Good choice.

I've done parsers before based on full interface fanciness, and it works, but there's costs.  Things end up noticeably slower.  If you _can_ have the full thing in-memory continuously, and do offset-based work on it fearlessly, that's gonna win in benchmarks pretty much every time.  (I realize this code might not be hottest path, but this is a lesson that was somewhat of a hard-won/surprise to me from parser work, so it's one I like to talk about.)  And it's just easier to work with!  Especially if one goes for an offset-heavy lexer API.

`bytes.Buffer` is also just plain more likely to be a footgun because so many methods mutate it.  If you want to hold one that's "immutable", really the only thing you ever call on it ends up being... `.Bytes()`.

And doing a `bytes.NewReader` every time you need one... yep.  Does the trick.  :check: 

---

Some other remarks near the lines that they're relevant to.

automod/engine/context.go
warpfork: In naming nits: I'd maybe put "raw" or "original" somewhere in the var name, so the reader doesn't have to do the mental walk through "ah, CBOR, since that's a serial format this probably means this is unparsed original raw bytes".  But that's just a (very mild) opinion.

bnewbold: cool, I switched to `[]byte`. I don't love any "raw" or "original" variants so i'm going to stick with `RecordCBOR` for now.

PR #586, opened 2024-02-21
Bump lexgen and support #identity events on the firehose

ericvolp12:


events/events.go
```diff
@@ -137,6 +137,7 @@ type XRPCStreamEvent struct {
 	Error         *ErrorFrame
 	RepoCommit    *comatproto.SyncSubscribeRepos_Commit
 	RepoHandle    *comatproto.SyncSubscribeRepos_Handle
+	Identity      *comatproto.SyncSubscribeRepos_Identity
```
bnewbold, line 140: It's true that the semantics of "RepoIdentity" would be a bit weird b/c identity kind of transcends a specific repo or account (?), but feels like we should stay consistent here and keep the `Repo` prefix?

ericvolp12: I guess we can, it just seemed a bit wonky cause it's related to the PLC identity not the repo necessarily.

I'm fin with it either way

whyrusleeping (approved):

PR #596, opened 2024-02-26
handle bsky.app urls in get-record

whyrusleeping:
got tired of translating bsky.app links into at:// links


ericvolp12 (approved):
LGTM, you might wanna add `list` in there too

PR #607, opened 2024-03-01
Support searching for quotes of a post when you see a bsky.app post U…

ericvolp12:
…RL in a search term


bnewbold (commented):
I'm a soft-block on this, don't think we should add quote-post search/enumeration lightly. I think this was an intentional decision before, not just "haven't gotten around to it".

If we *did* want to make it possible, we'd probably want to add direct lexicons and appview support, both for quote-posts and re-posts. Thought if we made the decision that we wanted them enumerable I wouldn't block this merge (search functionality).

bnewbold:
going to close this, per discussion and other search refactors happening in parallel

PR #608, opened 2024-03-03
search: allow at-prefixed handle with "from:"

bnewbold:
has a bit of test coverage which gives me confidence, but haven't done any other testings. blast radius seems small.

closes: https://github.com/bluesky-social/atproto/issues/2028


ericvolp12:
FYI I have some changes in https://github.com/bluesky-social/indigo/pull/607 that restructure how we parse queries a bit that might change how we want to do this if/when we merge it.

bnewbold:
will refactor this a bit soon, but can merge for now

PR #613, opened 2024-03-07
identity: remove ParsedPublicKey caching; tweak get/parsing method names and doc-comments

bnewbold:
The main thing here is removing the cached ("pre-parsed") key in the identity doc. I think this was non-trivial overhead when doing lots of lookups, and required manual override when serializing (actually caching!) the objects. I don't think we relied on the key parse caching for performance anywhere; if actual profiling shows we need it, should add as a private field, or cache externally.

`PublicKeyFor` read a bit weird and non-idiomatic to my eyes so I updated to `GetPublicKey(id)`, and added a service endpoint equivalent. If this is too nit-picky or the method is being used a bunch elsewhere (branches?) can revert that bit.

### DIFF
--- a/atproto/identity/base_directory.go
+++ b/atproto/identity/base_directory.go
@@ -53,11 +53,6 @@ func (d *BaseDirectory) LookupHandle(ctx context.Context, h syntax.Handle) (*Ide
 	}
 	ident.Handle = declared
 
-	// optimistic pre-parsing of public key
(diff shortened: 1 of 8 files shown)


ericvolp12 (approved):
LGTM

Jacob2161 (approved):
Like that name better as well :+1:

PR #617, opened 2024-03-12
lexgen: incorporate new tools.ozone namespace; bunch of old lexicons deprecated

bnewbold:
DO NOT MERGE YET

This is based off the atproto PR: https://github.com/bluesky-social/atproto/pull/2288

Does mostly expected things, adding `api/ozone` for tools.ozone.* namespace. Easier than expected because no records in that namespace, so no cborgen needed.

There are some cborgen changes rolled up in this PR. I think I didn't commit changes to at least the bsky codegen, something about string and pointer-to-string. I assume those are unrelated to the namespace stuff and should be resolved in a separate clean PR.

~I haven't self-reviewed or tested this at all, should do that before merging or deploying. I *think* I caught all the string constants with `com.atproto.admin.defs#whatever` in code, but there might be more lurking (these don't result in compile-time errors).~


ericvolp12:
@bnewbold the pointer and string changes are manual fixes that we need for now while a cborgen bug gets worked out. Iirc this has to do with pointers to strings and there's only one spot in the generated code that causes this error. If you manually replace the `string(<var>)` with `&<var>` in that line you can get it to build.

ericvolp12 (approved):
LGTM

ericvolp12 (approved):

PR #631, opened 2024-04-03
fix handle resolution in some cli commands

whyrusleeping:
probably more of this to come, the ProdHandleResolver type is basically broken now

### DIFF
--- a/cmd/gosky/admin.go
+++ b/cmd/gosky/admin.go
@@ -29,6 +29,10 @@ var adminCmd = &cli.Command{
 			EnvVars:  []string{"ATP_AUTH_ADMIN_PASSWORD"},
 			Required: true,
 		},
+		&cli.StringFlag{
+			Name:  "admin-endpoint",
+			Value: "https://mod.bsky.app",
+		},
 	},
 	Subcommands: []*cli.Command{
 		buildInviteTreeCmd,
@@ -78,7 +82,7 @@ var checkUserCmd = &cli.Command{
 
 		adminKey := cctx.String("admin-password")
 		xrpcc.AdminToken = &adminKey
-		xrpcc.Host = id.PDSEndpoint()
+		xrpcc.Host = cctx.String("admin-endpoint")
 
 		rep, err := toolsozone.ModerationGetRepo(ctx, xrpcc, did)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/gosky/admin.go
```diff
@@ -78,7 +82,7 @@ var checkUserCmd = &cli.Command{
 
 		adminKey := cctx.String("admin-password")
 		xrpcc.AdminToken = &adminKey
-		xrpcc.Host = id.PDSEndpoint()
+		xrpcc.Host = cctx.String("admin-endpoint")
```
bnewbold, line 85: makes sense. at some point we should probably have a better name than "admin" but i'm not sure what that should be right now

cmd/gosky/main.go
```diff
@@ -765,7 +765,7 @@ var listLabelsCmd = &cli.Command{
 		since := time.Now().Add(-1 * delta).UnixMilli()
 
 		xrpcc := &xrpc.Client{
-			Host: "https://api.bsky.app",
+			Host: "https://mod.bsky.app",
```
bnewbold, line 768: would be good to have this be a config (env var) as well, but not a blocking issue

cmd/gosky/admin.go
```diff
@@ -623,25 +627,25 @@ var takeDownAccountCmd = &cli.Command{
 
 		for _, did := range cctx.Args().Slice() {
 			if !strings.HasPrefix(did, "did:") {
-				phr := &api.ProdHandleResolver{}
-				resp, err := phr.ResolveHandleToDid(ctx, did)
+				dir := identity.DefaultDirectory()
+				resp, err := dir.LookupHandle(ctx, syntax.Handle(did))
```
bnewbold, line 631: feeling happy that these re-writes are basically one-to-one

bnewbold (approved):
happy to get this stuff over to the identity package! happy to help with any tweaks to that package if needed, and doing other similar code changes

cmd/gosky/admin.go
whyrusleeping: yeah, i always lean towards more verbose when i'm unsure here

cmd/gosky/admin.go
whyrusleeping: yeah, the only weird bit is i'm not using the parse handle code and am just casting direct to the handle type. Feels like unnecessary overhead to me

PR #639, opened 2024-04-09
update more BGS/Relay terminology

bnewbold:
not very high-priority, but noticed these doing some palomar work and did a cleanup.

didn't want to touch the actual bgs package and bigsky code base.

the CLI arg parsing stuff makes it easy to keep the old env vars around, which means this shouldn't impact deploys with old config.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -3,7 +3,7 @@
 
 Run with, eg, `go run ./cmd/bigsky`):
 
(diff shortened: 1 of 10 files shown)


ericvolp12 (approved):
lgtm if it builds and is backwards compatible with existing environment variable names for stuff

Issue #628, opened 2024-03-25
Palomar: Issue with Over-Normalization of Japanese Characters in Search Results


t2hnd:
Hi!

Hi! I've noticed an issue where the precision of search results for Japanese queries in bsky.app is compromised due to the over-normalization of Japanese characters, affecting the accuracy and relevance of search results for native speakers.

An example of such query is "パリ" (Paris in Japanese characters), which receives hits containing "ハリー・ポッター" (Harry potter) or "バリ" (Bali). As a native speaker I don't expect the query "パリ" to match these words because they don't share the character "パ". 
This result can be explained by the behaviour of icu_folding token filters added [here](https://github.com/bluesky-social/indigo/blob/main/search/post_schema.json#L14-L24). This filter normalizes Japanese characters with [different diacritics](https://en.wikipedia.org/wiki/Dakuten_and_handakuten), like the ones in the example "パ", "バ", "ハ". They represents different sounds and used in different words, so normalizing these characters usually lowers precision of Japanese search results.

### Steps to reproduce
Search a word "パリ" in bsky.app or try `https://bsky.social/xrpc/app.bsky.feed.searchPosts?q=%E3%83%91%E3%83%AA&limit=25` (The decoded query is パリ meaning Paris)

As a side note, the same behaviour is observed for other characters such as "ガン" (gun) and "カン" (can).

### Result 
Query receives hits containing  "パリ" along with the ones with "ハリ" or "バリ". The precision of the search result is low because of these hits.

### Expected results
Query receives hits containing  "パリ" only, which has fewer noise and helps to find related information.

### Possible solution
Replace icu_folding filter with asciifoliding filter, which doesn't normalize these Japanese characters. This may, however, affects search results for other languages like Greek that benefits from the icu_folding filter.
Another solution would be to use [Kuromoji analyzer](https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-kuromoji-analyzer.html) specialized for Japanese language but it probably requires per-language search configuration.

I believe addressing this issue will significantly enhance the user experience for Japanese speakers. I appreciate your attention on this issue.

bnewbold:
Hi @t2hnd! Thanks for raising this issue, and your detailed report. Including examples and links to the specific analyzers is very helpful.

We are sorry that the current search experience is not very good for Japanese queries. We will probably need to do more research, but off the top of my head, I can imagine two approaches to this issue:

[...]

PR #640, opened 2024-04-09
palomar: special-case Japanese text indexing using kuromoji

bnewbold:
This is just for posts right now, not profiles (descriptions, display name, etc).

I'm somewhat confident in the indexing approach (separate duplicate fields, gated by text detection). And this seems to work ok for simple cases.

I'm not very confident about all-kanji text and indexing, and mixes of Japanese and non-english character sets. For example, Japanese and Korean (CJK), or Japanese and Thai (non-CJK).

One positive thing is that everything is still being indexed in the regular text fields, using the existing analysis pipeline. So we can revert the query changes if needed, or improve some corner cases using query-time-only techniques.

Closes: https://github.com/bluesky-social/indigo/issues/628

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,11 @@ cborgen: ## Run codegen tool for CBOR serialization
 run-postgres: .env ## Runs a local postgres instance
 	docker compose -f cmd/bigsky/docker-compose.yml up -d
 
+.PHONY: run-dev-opensearch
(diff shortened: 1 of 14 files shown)


ericvolp12 (approved):
LGTM, gonna have a fun time tryna merge this into my working branch but it shouldn't be toooo bad

PR #657, opened 2024-04-22
Per-day new PDS limiter to allow less constrained federation

ericvolp12:
This change adds a per-day new PDS limiter (which can be admin overridden) allowing us to loosen the process around federating new PDSs without risk of being overrun with tons of fraudulent PDSs in a short time.

Admin interface updates include a way to view and update this limit:
CleanShot 2024-04-22 at 11 53 59


whyrusleeping (approved):
seems good enough to me, restarting the process will reset the limit it looks like but thats probably fine for our threat model for now at least

PR #660, opened 2024-04-23
crypto: add lenient versions for JWT parsing

bnewbold:


ericvolp12 (approved):
lgtm

bnewbold:
I still don't love this API shape, but will merge for now and maybe we can revisit when we add proper auth/JWT support directly in the SDK

PR #662, opened 2024-04-25
com.atproto.account.createInviteCode has been renamed to com.atproto.server.createInviteCode

nazo:
https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/server/createInviteCode.json


bnewbold:
good catch, thanks!

PR #663, opened 2024-05-01
palomar: 'OR' for single-token profile searches

bnewbold:
Just sketching, didn't even test this against a local index


ericvolp12 (approved):
LGTM (tested the constructed query on the prod cluster and it looks like it works)

PR #665, opened 2024-05-04
handlr: DNS TXT wrapper for handle resolution

bnewbold:


bnewbold:
@itaru2622 you might be interested in this; see the README.

it is kind of small and weird to include in indigo proper, I might create a separate git repo for it.

ThisIsMissEm:
So I encountered a really weird issue when testing out `did:web` for activitypds, and maybe this could help: essentially the `TXT _atproto.<handle>.domain.example` overwrites the `A <handle>.domain.example` if it comes from `A *.domain.example`

So maybe this would be necessary for `did:web` where you have a server that serves both the TXT records and the A records for each `<handle>.domain.example`?

PR #666, opened 2024-05-05
fallback to clean repo reset in more cases

whyrusleeping:
If we have a filesystem messup for some reason just wipe and reset the repo


ericvolp12 (approved):
lgtm

PR #668, opened 2024-05-07
Acquire semaphore outside of goroutine for resyncs

ericvolp12:
Currently we end up with millions of goroutines waiting on semaphores during a full PDS resync if we run against all production PDSs, this fix waits for the semaphore _before_ spawning the goroutine reducing allocs and the size of the stack significantly.


whyrusleeping (approved):
well now

PR #669, opened 2024-05-09
Index alt text of images embedded in quote posts

ericvolp12:


bnewbold (commented):
I was confused by the title, but this means including alt-text from images embedded in the post, covering the case of posts that have both images *and* a quote-post. It is easy to miss this situation!

Code looks ok; i'd request adding a test case to the test fixture file (`transform-post-fixtures.json`). You should be able to copy/paste one of the existing fixtures and edit it to cover this case.

With a test, :+1: to merge.

bnewbold (approved):

PR #674, opened 2024-05-25
beemo: refactor and add mention notifications

bnewbold:


bnewbold:
@ericvolp12 nothing very high-stakes here but a skim and plus1 would be good if you get a sec

cmd/beemo/firehose_consumer.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	rsc := &events.RepoStreamCallbacks{
+		RepoCommit: func(evt *comatproto.SyncSubscribeRepos_Commit) error {
+			return HandleRepoCommit(ctx, logger, evt, postCallback)
+		},
+		// NOTE: could add other callbacks as needed
+	}
+
+	var scheduler events.Scheduler
```
ericvolp12, line 49: I would recommend using the parallelScheduler tbh, the autoscalingScheduler has been more trouble than it's worth lately.

cmd/beemo/notify_mentions.go
```diff
@@ -0,0 +1,88 @@
...
+func (mc *MentionChecker) ProcessPost(ctx context.Context, did syntax.DID, rkey syntax.RecordKey, post appbsky.FeedPost) error {
+	mc.logger.Debug("processing post record", "did", did, "rkey", rkey)
+
+	for _, facet := range post.Facets {
+		for _, feature := range facet.Features {
+			mention := feature.RichtextFacet_Mention
+			if mention == nil {
+				continue
+			}
+			for _, d := range mc.mentionDIDs {
```
ericvolp12, line 33: Instead of a for loop over a list here, you could make `mentionDIDs` a map and key into it, would be a bit more idiomatic Go if that's the only way we're using that struct (to check membership), turn `mentionDIDs` into a `map[string]struct{}` which is Go's version of a "set"

cmd/beemo/notify_reports.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return err
+		}
+		xrpcc.Auth.AccessJwt = refresh.AccessJwt
+		xrpcc.Auth.RefreshJwt = refresh.RefreshJwt
+
+		// query just new reports (regardless of resolution state)
+		// ModerationQueryEvents(ctx context.Context, c *xrpc.Client, createdBy string, cursor string, includeAllUserRecords bool, limit int64, sortDirection string, subject string, types []string) (*ModerationQueryEvents_Output, error)
+		var limit int64 = 50
+		me, err := toolsozone.ModerationQueryEvents(
```
ericvolp12, line 73: lol lmao, love this method

ericvolp12 (approved):
LGTM aside from a few nits, no need to fix them tho if you don't mind

cmd/beemo/notify_reports.go
bnewbold: yeah, it's a bummer what is fairly reasonable/idiomatic for API design ends up so bad in codegen here. several ways we could re-architect the codegen some day. maybe worth creating a more ergonomic wrapper (with config struct) for a handful of these monster methods.

cmd/beemo/notify_mentions.go
bnewbold: you are correct, but i'm going to be lazy and skip on this refactor in this case. i'll try to use that pattern more in the future

PR #675, opened 2024-05-28
remove blobs package (including from BGS)

bnewbold:
We haven't ever actually used the blob sync stuff in prod. This removes that dead code, and the entire "blobs" package. We can always pull this out of git history if we decide we want it back.

This change is expected to be safe and a no-op to merge and deploy. I verified that `disk-blob-store` is a flag-only option, and that flag does not show up in `grep` results in our internal infra repos.


ericvolp12 (approved):
LGTM

PR #676, opened 2024-05-28
Support #account events in the Relay event stream

ericvolp12:
This change includes the required changes to support `#account` events from PDSs that allow for PDS-directed suspension, takedowns, deactivation, and reactivation of accounts.

Still need to fix cborgen cause it broke _again_...


ericvolp12:
Think I got the cborgen working now, just missing some tests now I think. Can hit those up tomorrow.

whyrusleeping (approved):
LGTM

whyrusleeping (approved):

PR #678, opened 2024-06-03
keep going if we encounter corrupted shard files

whyrusleeping:
If the relay gets shut down very suddenly and things arent properly synced to disk (I think this only happens if there are unexpected power failures on the host machine) then we can get into weird states where some repos have corrupted shard files. 
They keep functioning fine at a high level, but compaction gets impacted and prevented from finishing.

These repos should really just get wiped and resynced, but this change should prevent them from wreaking havoc in the compaction engine.


ericvolp12 (approved):
LGTM

PR #681, opened 2024-06-05
in postgres use optimized where in query

whyrusleeping:
We need to find a good way to run these tests on postgres


bnewbold:
re: postgresql testing

The `atproto` repo (typescript) has some helper scripts for setting up a local postgresql during test invocation. It slows things down a little bit, but in go we could tag the tests which need that, and run them as a separate `make` command (?), and have CI run that and the regular tests together; shouldn't be too bad.

ericvolp12:
Merged this since commit c13ab32 is what's actually in prod rn

PR #682, opened 2024-06-08
Relay Github Action.

oracularhades:
Builds the Relay to a Docker Github package. The action is set to "on: workflow_dispatch", which is manual. I left it out so a team member can set the condition.

This action file can also be used elsewhere!: "# To use this Github action file with another part of this repo, or with another project, change the DOCKERFILE and PACKAGE_NAME variables below."

🦋

Thanks,
Josh.


oracularhades:
Sorry for the extra commits, I was trying to debug this, and didn't realize Github started writing over here instead of my fork: https://bsky.app/profile/oracularhades.com/post/3kuexlcb5wk2a

bnewbold:
The existing `container-bigsky-ghcr.yaml` action seems pretty similar to this. Are you looking for something different/additional with this PR?

PR #689, opened 2024-06-27
relay: code tweaks, README, and helper scripts for self-hosting

bnewbold:
- Makefile command to build dashboard
- small refactors and simplifications of bigsky env var and CLI arg parsing; should be safe to merge and deploy to prod
- disable some Relay features by default when bigsky is run (like "aggregations" and "spidering")

The helper bash scripts are a bit duplicate with `gosky` commands, and aren't super important, I can yank those if they seem too redundant.

### DIFF
--- a/cmd/bigsky/main.go
+++ b/cmd/bigsky/main.go
@@ -154,7 +154,7 @@ func run(args []string) error {
 			EnvVars: []string{"MAX_FETCH_CONCURRENCY"},
 		},
 		&cli.StringFlag{
-			Name:    "environment",
+			Name:    "env",
 			EnvVars: []string{"ENVIRONMENT"},
 			Usage:   "declared hosting environment (prod, qa, etc); used in metrics",
 		},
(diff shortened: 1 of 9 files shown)


cmd/bigsky/main.go
```diff
@@ -217,14 +229,27 @@ func Bigsky(cctx *cli.Context) error {
...
 		otel.SetTracerProvider(tp)
 	}
 
+	return nil
+}
+
+func runBigsky(cctx *cli.Context) error {
+	// Trap SIGINT to trigger a shutdown.
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
ericvolp12, line 246: This bit makes me nervous to add in unless it's been well tested

ericvolp12 (approved):
One note on the signal trapping change but otherwise lgtm

cmd/bigsky/main.go
ericvolp12: Oh NVM that's in the existing version, GH just got confused

ericvolp12:
NVM Github rendered the change weirdly and made it seem like that was new code, it was just moved, LGTM.

PR #695, opened 2024-07-03
more relay hosting tweaks

bnewbold:
This has some low-impact doc and scripts changes, and a couple actually-impactful changes to Relay code:

- I think fixes a bug/type with per-PDS, per-second event rate limits
- wires through some concurrency and limit values as env vars; should not impact the default behaviors
- updates PDS resync numbers (in database and results from API) much more frequently, but doesn't log for all those updates. only impacts actual PDS resync processes

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -1616,8 +1616,10 @@ func (bgs *BGS) ResyncPDS(ctx context.Context, pds models.PDS) error {
 				log.Errorw("failed to enqueue crawl for repo during resync", "error", err, "uid", res.ai.Uid, "did", res.ai.Did)
 			}
 		}
-		if i%100_000 == 0 {
-			log.Warnw("checked revs during resync", "num_repos_checked", i, "num_repos_to_crawl", numReposToResync, "took", time.Now().Sub(resync.StatusChangedAt))
+		if i%100 == 0 {
+			if i%10_000 == 0 {
+				log.Warnw("checked revs during resync", "num_repos_checked", i, "num_repos_to_crawl", numReposToResync, "took", time.Now().Sub(resync.StatusChangedAt))
+			}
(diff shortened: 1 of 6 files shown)


bgs/bgs.go
```diff
@@ -106,7 +106,25 @@ type SocketConsumer struct {
...
+	return BGSConfig{
+		SSL:               true,
+		CompactInterval:   4 * time.Hour,
+		DefaultRepoLimit:  100,
+		ConcurrencyPerPDS: 100,
+		MaxQueuePerPDS:    1_000,
+	}
+}
+
+func NewBGS(db *gorm.DB, ix *indexer.Indexer, repoman *repomgr.RepoManager, evtman *events.EventManager, didr did.Resolver, rf *indexer.RepoFetcher, hr api.HandleResolver, config BGSConfig) (*BGS, error) {
```
ericvolp12, line 127: This should probably receive the config as a pointer and then init it as the default if the config it receives is nil.

bgs/bgs.go
```diff
@@ -1158,7 +1179,7 @@ func (s *BGS) createExternalUser(ctx context.Context, did string) (*models.Actor
 		peering.Host = durl.Host
 		peering.SSL = (durl.Scheme == "https")
 		peering.CrawlRateLimit = float64(s.slurper.DefaultCrawlLimit)
-		peering.RepoLimit = s.slurper.DefaultPerSecondLimit
+		peering.RateLimit = float64(s.slurper.DefaultPerSecondLimit)
```
ericvolp12, line 1182: oops, good catch

bgs/bgs.go
bnewbold: good point, updated to to that

ericvolp12 (approved):
sure

PR #697, opened 2024-07-03
automod: add basic repost rule

foysalit:
TODO:
- [ ] Verify the threshold values are correct
- [ ] Uncomment ozone report line

This PR adds basic checks for relatively new accounts reposting too frequently and reports them to ozone.

### DIFF
--- a/automod/rules/reposts.go
+++ b/automod/rules/reposts.go
@@ -1,6 +1,7 @@
 package rules
 
 import (
+	"strings"
 	"time"
 
 	"github.com/bluesky-social/indigo/automod"
@@ -23,6 +24,12 @@ func TooManyRepostRule(c *automod.RecordContext) error {
 		if age > 30*24*time.Hour {
 			return nil
 		}
+	}
+
+	// Special case for newsmast bridge feeds
+	handle := c.Account.Identity.Handle.String()
+	if strings.HasSuffix(handle, "newsmast.community.ap.brid.gy") {
+		return nil
 	}
 
 	switch c.RecordOp.Collection {
--- a/automod/rules/reposts.go
+++ b/automod/rules/reposts.go
@@ -7,7 +7,9 @@ import (
 	"github.com/bluesky-social/indigo/automod/countstore"
 )
 
-var repostDailyThreshold = 500
+var dailyRepostThresholdWithoutPost = 20
+var dailyRepostThresholdWithLowPost = 100
+var dailyPostThresholdWithHighRepost = 5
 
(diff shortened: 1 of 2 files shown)


automod/rules/reposts.go
```diff
@@ -0,0 +1,42 @@
...
+	if c.Account.Private != nil {
+		age := time.Since(c.Account.Private.IndexedAt)
+		if age > 30*24*time.Hour {
+			return nil
+		}
+	}
+
+	switch c.RecordOp.Collection {
+	case "app.bsky.feed.post":
+		c.Increment("post", did)
```
bnewbold, line 28: I thought we already had a counter like this elsewhere, but guess we don't!

automod/rules/reposts.go
```diff
@@ -0,0 +1,42 @@
...
+var _ automod.RecordRuleFunc = TooManyRepostRule
+
+// looks for accounts which do frequent reposts
+func TooManyRepostRule(c *automod.RecordContext) error {
+
+	did := c.Account.Identity.DID.String()
+	// Don't bother checking reposts from accounts older than 30 days
+	if c.Account.Private != nil {
+		age := time.Since(c.Account.Private.IndexedAt)
+		if age > 30*24*time.Hour {
```
bnewbold, line 21: I have this as 2 weeks in a couple other places, but 30 days is totally reasonable for this rule.

automod/rules/reposts.go
```diff
@@ -0,0 +1,42 @@
+package rules
+
+import (
+	"time"
+
+	"github.com/bluesky-social/indigo/automod"
+	"github.com/bluesky-social/indigo/automod/countstore"
+)
+
+var repostDailyThreshold = 500
```
bnewbold, line 10: Definitely curious if we'll have any triggers at this value! I might have started at 100... but maybe these accounts are going really hard already? And there might be heavy organic re-post use by some users.

bnewbold (approved):

automod/rules/reposts.go
foysalit: I thought we'd have one too but couldn't find one.

automod/rules/reposts.go
foysalit: good point, at least for test, I can set a lower threshold maybe?

automod/rules/reposts.go
foysalit: cool yeah I was going with what Aaron recommended.

PR #707, opened 2024-07-25
configuring initial perDayLimit via cmdline/env-var, to reduce ops

### LINKED ISSUE #684: initial value of perDayLimit (0) causes failures on lots of features.
as described in https://github.com/bluesky-social/atproto/discussions/2454

initial value of perDayLimit causes failures on lot of features, until  its value configured via REST API.
It should have the option or environment variable to set initial perDayLimit value to reduce operations and trobles.

itaru2622:
fix #684

This PR supports you to configure **initial value** of perDayLimit via commandline or environment variable, to reduce ops.

```
cmdline option:  --newpds-perday-limit N
env var:  RELAY_NEWPDS_PERDAY_LIMIT=N
```

the above ops takes efforts only when the the value is not configured before.
so there is no bad side-effect on your existing parameters, nor your runtime ops.

if both of cmdline option and env var are not specified at first deployment, default value 10 is passed by cli.Context

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -112,6 +112,7 @@ type BGSConfig struct {
 	DefaultRepoLimit  int64
 	ConcurrencyPerPDS int64
 	MaxQueuePerPDS    int64
+	InitialNewPDSPerDayLimit int64
 }
 
 func DefaultBGSConfig() *BGSConfig {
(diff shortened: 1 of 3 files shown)


itaru2622:
@bnewbold , maintainers   could you review this PR, please.

itaru2622:
@bnewbold , maintainers could you review this PR, please

PR #709, opened 2024-07-30
lexgen runs all packages together

brianolson:
cross-package references can now make a type know that it needs to include a type field

### DIFF
--- a/lex/gen.go
+++ b/lex/gen.go
@@ -25,6 +25,7 @@ const (
 )
 
 type Schema struct {
+	path   string
 	prefix string
 
 	Lexicon int                    `json:"lexicon"`
@@ -50,9 +51,9 @@ type InputType struct {
 }
 
 type TypeSchema struct {
-	prefix    string
-	id        string
-	defName   string
+	prefix    string // prefix of a major package being processed, e.g. com.atproto
+	id        string // parent Schema.ID
+	defName   string // parent Schema.Defs[defName] points to this *TypeSchema
 	defMap    map[string]*ExtDef
 	needsCbor bool
 	needsType bool
@@ -149,7 +150,6 @@ func (s *Schema) AllTypes(prefix string, defMap map[string]*ExtDef) []outputType
 		if ts.Type == "ref" {
 			refname := ts.Ref
 			if strings.HasPrefix(refname, "#") {
-				fmt.Println("Foo")
 				refname = s.ID + ts.Ref
 			}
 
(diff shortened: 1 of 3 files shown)


lex/gen.go
```diff
@@ -1470,3 +1473,51 @@ func (ts *TypeSchema) writeCborUnmarshalerEnum(name string, w io.Writer) error {
...
+	Prefix    string
+	Outdir    string
+}
+
+var Packages []Package = []Package{
+	Package{"bsky", "app.bsky", "api/bsky"},
+	Package{"atproto", "com.atproto", "api/atproto"},
+	Package{"chat", "chat.bsky", "api/chat"},
+	Package{"ozone", "tools.ozone", "api/ozone"},
+}
```
bnewbold, line 1488: in the near-ish future we are going to want to make it easy for folks to use this tooling for non-bluesky-maintained lexicons (aka, independent lexicons).

in the medium-term, it is most likely that any inter-namespace references would be from independent lexicons to one of these lexicons (aka, independent lexicons referencing `com.atproto.*` or `app.bsky.actor.profile`). so having these hard-coded in is fine.

overall I think having these hard-coded is fine for this PR, but I would guess we might want an additional command more like the previous behavior soon (coming weeks).

lex/gen.go
```diff
@@ -50,9 +51,9 @@ type InputType struct {
 }
 
 type TypeSchema struct {
-	prefix    string
-	id        string
-	defName   string
+	prefix    string // prefix of a major package being processed, e.g. com.atproto
+	id        string // parent Schema.ID
+	defName   string // parent Schema.Defs[defName] points to this *TypeSchema
```
bnewbold, line 56: comments :muscle: !

bnewbold (approved):
I ran this branch against the atproto repo `main` branch and the output looks good/expected to me.

Left some notes about flexibility with the lexgen tool with 3rd party lexicons, which hasn't been taken advantage of much to date, but probably will be soon (coming weeks). I can take adding an additional entrypint / sub-command for that use-case when it is needed.

I only skimmed the actual code changes, assuming that @ericvolp12 or @whyrusleeping will review code quality.

ericvolp12:
Confirming on my end that the changes to lexicons that I see when running lexgen from this branch look good (just out of date/stale stuff that hasn't been needed on the Go side of things yet). I figure you're gonna merge this and then open another PR with the updated lexicons and requisite changes to make stuff build?

Also I do agree with @bnewbold that it may be best to pass the packages into the `Run()` entrypoint in `lex/gen.go` rather than have them hardcoded inside that file. Should be totally fine to define those in `cmd/lexgen` as a default value somewhere that users can override with a CLI flag or something.

PR #726, opened 2024-08-22
split lex/gen.go

brianolson:
gen.go was getting long and weird, split up by major classes
this will make further work easier


ericvolp12 (approved):
LGTM

PR #728, opened 2024-08-27
add rigging for hive to pre-screen images before sending to hive

whyrusleeping:
for now, just going to log if the pre-screen fails at its job


bnewbold (approved):
generally looks good, but would like to have context wired through.

you can get the context from the `RecordContext` with `c.Ctx`:
https://github.com/bluesky-social/indigo/blob/main/automod/visual/hiveai_rule.go#L16

PR #733, opened 2024-08-30
Relay: better buffer sizes (and separate sizes per buffer type), and fix a loop/bug with behind consumers

whyrusleeping:


bnewbold (approved):
Big change, I would maybe have done like 32k -> 4k, but i'm ok with this. Let's deploy from branch to learn.

ericvolp12 (approved):
Ohhhhhhhhhhhhh ok

bnewbold:
This is working good now and we should probably merge soon

PR #736, opened 2024-09-02
:sparkles: Add combo hashtag rule

foysalit:
This PR adds a way for us to takedown accounts that post content with very bad hashtags.
This introduces the need for 2 new sets `potentially-bad-hashtags` and `potentially-bad-hashtag-combos`. 

However, ideally, these would be nested arrays so that each `potentially-bad-hashtags` item would have nested items of `potentially-bad-hashtag-combos` so that we could build hierarchies like 

'test' -> ['check', 'now']
'call' - > ['me', 'her']
- if a post has #test and #check rule will be activated but 'test' and 'me' won't


bnewbold:
I would make this a separate new rule instead of bundling with the current rule.

Jumping straight to takedown is also aggressive. In an urgent situation we can at least start with `!warn` or `!hide` on the overall account. Cleaning up mistakes is *very* laborious, and this kind of rule in particular is going to have a bunch of false-positives.

bnewbold:
we moved rules internally, closing this PR

PR #737, opened 2024-09-02
optimize alloc hot spot

brianolson:
15% of alloc are NewCborWriter() inside MarshalCBOR() in BGS.EventHandler(), cut half of them

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -39,6 +39,7 @@ import (
 	promclient "github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	dto "github.com/prometheus/client_model/go"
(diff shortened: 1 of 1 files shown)


brianolson:
TODO: also update to cbor-gen that makes fewer alloc https://github.com/whyrusleeping/cbor-gen/pull/104

bgs/bgs.go
```diff
@@ -686,11 +687,12 @@ func (bgs *BGS) EventsHandler(c echo.Context) error {
 				return fmt.Errorf("unrecognized event kind")
 			}
 
-			if err := header.MarshalCBOR(wc); err != nil {
+			cborWriter := cbg.NewCborWriter(wc)
```
whyrusleeping, line 690: move this outside of the loop and use `SetWriter`

brianolson:
do #738 instead

PR #738, opened 2024-09-03
serialize outbound once

brianolson:


events/events.go
```diff
@@ -165,6 +172,57 @@ type XRPCStreamEvent struct {
...
+	return obj.MarshalCBOR(cborWriter)
+}
+
+// serialize content into Preserialized cache
+func (evt *XRPCStreamEvent) Preserialize() {
+	if evt.Preserialized != nil {
+		return
+	}
+	var buf bytes.Buffer
+	evt.Serialize(&buf)
```
whyrusleeping, line 224: probably still want to check  (or explicitly ignore if we are extremely confident) errors

whyrusleeping (approved):
LGTM

PR #740, opened 2024-09-04
:saprkles: Auto escalate accounts for high follow/like churn

foysalit:


automod/engine/effects.go
```diff
@@ -128,6 +132,11 @@ func (e *Effects) TakedownAccount() {
 	e.AccountTakedown = true
 }
 
+// Enqueues the entire account to be taken down at the end of rule processing.
```
bnewbold, line 135: comment out of date (copypasta)

automod/engine/metrics.go
```diff
@@ -40,6 +40,11 @@ var actionNewTakedownCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Help: "Number of new flags persisted",
 }, []string{"type"})
 
+var actionNewEscalationCount = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "automod_new_action_escalations",
+	Help: "Number of new flags persisted",
```
bnewbold, line 45: out of date

automod/engine/persist.go
```diff
@@ -57,8 +57,13 @@ func (eng *Engine) persistAccountModActions(c *AccountContext) error {
 	if err != nil {
 		return fmt.Errorf("circuit-breaking takedowns: %w", err)
 	}
+	// @TODO: do we want to check if the account is already escalated?
```
bnewbold, line 60: yeah, we should check here. we should also maybe popular review start as part of account meta? and clear account meta when it changes.

bnewbold (commented):
this generally looks good. this kind of very-impactful change is something we'll want to test carefully in staging; i'm not sure we can get this through in the next week.

automod/rules/interaction.go
```diff
@@ -25,6 +25,7 @@ func InteractionChurnRule(c *automod.RecordContext) error {
 			c.Logger.Info("high-like-churn", "created-today", created, "deleted-today", deleted)
 			c.AddAccountFlag("high-like-churn")
 			c.ReportAccount(automod.ReportReasonSpam, fmt.Sprintf("interaction churn: %d likes, %d unlikes today (so far)", created, deleted))
+			c.EscalateAccount()
```
bnewbold, line 28: I'd be in favor of commenting this out for now, so we can review and get the engine change merged, then do the policy/behavior change separately.

PR #742, opened 2024-09-04
run N threads of compactor

brianolson:
riffing on #741

### DIFF
--- a/bgs/compactor.go
+++ b/bgs/compactor.go
@@ -105,17 +105,25 @@ type CompactorState struct {
 	stats     *carstore.CompactionStats
 }
 
+func (cstate *CompactorState) set(uid models.Uid, did, status string, stats *carstore.CompactionStats) {
+	cstate.latestUID = uid
+	cstate.latestDID = did
+	cstate.status = status
+	cstate.stats = stats
+}
+
 // Compactor is a compactor daemon that compacts repos in the background
 type Compactor struct {
 	q                 *uniQueue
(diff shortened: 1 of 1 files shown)


bgs/compactor.go
```diff
@@ -158,34 +167,15 @@ type compactionStats struct {
...
-
 var errNoReposToCompact = fmt.Errorf("no repos to compact")
 
 // Start starts the compactor
 func (c *Compactor) Start(bgs *BGS) {
 	log.Info("starting compactor")
-	go c.doWork(bgs)
+	c.wg.Add(c.numWorkers)
+	for _ = range c.numWorkers {
+		go c.doWork(bgs, &c.wg)
```
whyrusleeping, line 177: no need to pass the workgroup if its on the compactor object already

bgs/compactor.go
```diff
@@ -263,34 +252,37 @@ func (c *Compactor) doWork(bgs *BGS) {
 	}
 }
 
-func (c *Compactor) compactNext(ctx context.Context, bgs *BGS) (*CompactorState, error) {
+func (c *Compactor) compactNext(ctx context.Context, bgs *BGS) (state CompactorState, err error) {
```
whyrusleeping, line 255: if i could delete one thing from Go, it would be named returns. I really try to avoid using them ever as they always make the code harder to follow

brianolson:
do #744 instead

PR #743, opened 2024-09-04
Pre-allocate ops slices that are appended in a loop where possible.

avalonbits:
This ensures that only one larger allocation is done instead of multiple small ones, reducing pressure on the garbage collector.


brianolson (approved):
Yup. This is a better pattern. Thank you.

bnewbold (approved):
thanks!

PR #744, opened 2024-09-04
multi-threaded compactor with randomness

brianolson:
This is #742 with additional feature of half the compactor workers pop randomly from the backlog

### DIFF
--- a/bgs/compactor.go
+++ b/bgs/compactor.go
@@ -2,7 +2,9 @@ package bgs
 
 import (
 	"context"
+	crypto_rand "crypto/rand"
 	"fmt"
+	"math/rand/v2"
 	"sync"
 	"time"
 
@@ -22,6 +24,7 @@ type uniQueue struct {
 	q       []queueItem
 	members map[models.Uid]struct{}
 	lk      sync.Mutex
+	rnd     *rand.Rand
 }
 
 // Append appends a uid to the end of the queue if it doesn't already exist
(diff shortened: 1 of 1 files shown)


bgs/compactor.go
```diff
@@ -82,20 +85,50 @@ func (q *uniQueue) Remove(uid models.Uid) {
 }
 
 // Pop pops the first item off the front of the queue
-func (q *uniQueue) Pop() (*queueItem, bool) {
+func (q *uniQueue) Pop() (item queueItem, ok bool) {
```
whyrusleeping, line 88: no named returns please

bgs/compactor.go
```diff
@@ -131,9 +173,17 @@ func DefaultCompactorOptions() *CompactorOptions {
 		RequeueLimit:      0,
 		RequeueShardCount: 50,
 		RequeueFast:       true,
+		NumWorkers:        2,
 	}
 }
 
+func newRandFromRoot() *rand.Rand {
```
whyrusleeping, line 180: this is unnecessary, the default rand is already well seeded

bgs/compactor.go
```diff
@@ -217,30 +252,29 @@ func (c *Compactor) Start(bgs *BGS) {
 func (c *Compactor) Shutdown() {
 	log.Info("stopping compactor")
 	close(c.exit)
-	<-c.exited
+	c.wg.Wait()
 	log.Info("compactor stopped")
 }
 
-func (c *Compactor) doWork(bgs *BGS) {
+func (c *Compactor) doWork(bgs *BGS, wg *sync.WaitGroup, strategy NextStrategy) {
```
whyrusleeping, line 259: no need to pass the waitgroup, just use the struct member

bgs/compactor.go
```diff
@@ -263,34 +297,51 @@ func (c *Compactor) doWork(bgs *BGS) {
...
 
-func (c *Compactor) compactNext(ctx context.Context, bgs *BGS) (*CompactorState, error) {
+type NextStrategy int
+
+const (
+	NextInOrder NextStrategy = iota
+	NextRandom
+)
+
+func (c *Compactor) compactNext(ctx context.Context, bgs *BGS, strategy NextStrategy) (state CompactorState, err error) {
```
whyrusleeping, line 307: no named returns please

ericvolp12:
This looks pretty reasonable to me, jumbling up the queue doesn't make me too concerned cause the stuff at the end of the queue is gonna be really quick to compact anyhow. I'm +1 on all of Why's comments.

bgs/compactor.go
brianolson: I think the change here is to drop the `Compactor.exited` member in favor of a sync.WorkGroup which more naturally fits N workers instead of 1

bgs/compactor.go
brianolson: this gives us a *Rand that can be owned and accessed lockless by some thread

bgs/compactor.go
ericvolp12: The waitgroup is on the `*Compactor` object already though isn't it? It doesn't need to be passed into this function as well as a param

brianolson: derp. yes.

bgs/compactor.go
whyrusleeping: we're not performance constrained on the rand here at all, and even if we were we could just use `rand.New(rand.NewSource(time.Now().UnixNano()))`

whyrusleeping (approved):
cool, LGTM. lets give it a shot

PR #746, opened 2024-09-05
add `IdenticalReplyPostSameParentRule` to automod

haileyok:
adds a spam label to accounts that are under 24 hours doing this, otherwise just reports. maybe too hasty on the label, can adjust. (note our client really doesn't make it easy to make an identical post multiple times in a row like this, where even failures will usually close the composer anyway, so its likely to be intentional). same could probably be applied for same author but with a maybe higher limit

### DIFF
--- a/automod/rules/replies.go
+++ b/automod/rules/replies.go
@@ -80,7 +80,8 @@ func IdenticalReplyPostRule(c *automod.RecordContext, post *appbsky.FeedPost) er
 
 // Similar to above rule but only counts replies to the same post. More aggressively applies a spam label to new accounts that are less than a day old.
 var identicalReplySameParentLimit = 3
-var identicalReplySameParentSpamLabel = 24 * time.Hour
+var identicalReplySameParentMaxAge = 24 * time.Hour
+var identicalReplySameParentMaxPosts int64 = 50
 var _ automod.PostRuleFunc = IdenticalReplyPostSameParentRule
 
 func IdenticalReplyPostSameParentRule(c *automod.RecordContext, post *appbsky.FeedPost) error {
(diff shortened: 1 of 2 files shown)


automod/rules/replies.go
```diff
@@ -78,6 +78,39 @@ func IdenticalReplyPostRule(c *automod.RecordContext, post *appbsky.FeedPost) er
...
+		return nil
+	}
+
+	if ParentOrRootIsFollower(c, post) {
+		return nil
+	}
+
+	period := countstore.PeriodDay
+	bucket := c.Account.Identity.DID.String() + "/" + post.Reply.Parent.Uri + "/" + HashOfString(post.Text)
+	c.IncrementPeriod("reply-text-same-post", bucket, period)
```
bnewbold, line 97: nice, this is a bit subtle but you did the right thing: the cardinality of (poster, URI, hash) is going to be huge, but by using `IncrementPeriod` (instead of generic increment), we limit to only last-hour bucket and the counters won't stick around.

automod/rules/replies.go
```diff
@@ -78,6 +78,39 @@ func IdenticalReplyPostRule(c *automod.RecordContext, post *appbsky.FeedPost) er
...
+
+	count := c.GetCount("reply-text-same-post", bucket, period)
+	if count >= identicalReplySameParentLimit {
+		// if account is young, add spam label. otherwise just report.
+		createdAt := c.Account.CreatedAt
```
bnewbold, line 102: It is tempting to find more filters to skip earlier, before the counter is even incremented. As a perf thing. We mostly have "new influx" but will get "new reactivation" as well. We could filter on "older than a day, and more than 50 posts already" just skips? I think we have the post count pre-cached on the AccountMetadata.

Maybe overthinking though, redis is fast.

automod/rules/replies.go
```diff
@@ -78,6 +78,39 @@ func IdenticalReplyPostRule(c *automod.RecordContext, post *appbsky.FeedPost) er
...
+		c.AddAccountFlag("multi-identical-reply-same-post")
```
bnewbold, line 103: I had these all over the place earlier, I think we could just skip or remove adding these flags now? hrm.

automod/rules/replies.go
```diff
@@ -78,6 +78,39 @@ func IdenticalReplyPostRule(c *automod.RecordContext, post *appbsky.FeedPost) er
...
+		c.ReportAccount(automod.ReportReasonSpam, fmt.Sprintf("possible spam (%d identical reply-posts to same post today)", count))
+		if time.Since(*createdAt) < identicalReplySameParentSpamLabel {
```
bnewbold, line 105: I think we want a `nil` check here (aka "if createdAt != nil && [...]"). There might be a helper function which does the age check?

bnewbold (approved):
Nice! This rocks and I think would be safe to merge+deploy as written.

Left some comments but overall good.

automod/rules/replies.go
haileyok: yea tbh think this makes sense in this context. aged acct more likely to be used for something else than this. maybe.

automod/rules/replies.go
haileyok: yup

automod/rules/replies.go
haileyok: tweaked this to hour, think that makes sense.

automod/rules/replies.go
haileyok: 🤷‍♀️ adds context to slack ig, lmk if should rm

bnewbold: oh right!

PR #757, opened 2024-09-12
refactor carstore meta into a class

brianolson:
This should be the "no functional change, just code moving" change.
Future change likely to be making an interface and alternative implementation.

### DIFF
--- a/carstore/bs.go
+++ b/carstore/bs.go
@@ -4,14 +4,11 @@ import (
 	"bufio"
 	"bytes"
 	"context"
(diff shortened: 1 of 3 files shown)


brianolson:
There is also a next patch for turning carstore.CarStore into an interface which carstore.FileCarStore (formerly known as CarStore) implements. Could be merged into this or merged to main sequentially.
https://github.com/brianolson/indigo/pull/1

ericvolp12 (approved):
LGTM

PR #760, opened 2024-09-19
document admin endpoints

brianolson:


ericvolp12 (approved):
LGTM, most of these are in service of the built-in admin WebUI so they may end up changing if requirements for the WebUI changes but they've been relatively stable for the past few months.

PR #765, opened 2024-09-25
syntax: clearer error for trivial empty string case

bnewbold:
Was getting "doesn't match regex" for trivial empty-string situations. having shorter more descriptive errors helps debugging.


brianolson:
`fmt.Errorf("")` with no % formatting should be `errors.New("")`, but otherwise looks good

PR #766, opened 2024-09-26
mst: clarify in comment that fields are nullable, not optional

bnewbold:
See also: https://github.com/bluesky-social/atproto-website/pull/346


brianolson (approved):
ok, _not_ `omitempty`, but it could be `"l":null` or `"t":null`

PR #767, opened 2024-09-29
support tokenize while keeping common censor chars

haileyok:
There are some cases where it's useful to tokenize a string while _not_ splitting on some non-letter chars like `#`, `*`, `-`, or `_`. Unfortunately right now `Tokenize` will split on all of these, making some matching difficult.

This just adds a second `TokenizeTextSkippingCensorChars` for those particular use cases. Also adding `TokenizeTextWithRegex`, so that other cases can be easily covered in the future if they arise.

### DIFF
--- a/automod/keyword/tokenize.go
+++ b/automod/keyword/tokenize.go
@@ -12,24 +12,37 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


bnewbold (approved):

TUNISIA-user:
hey

PR #768, opened 2024-09-30
goat account migration

bnewbold:
This PR adds support for account migration using the goat CLI tool two different ways:

- a bunch of account helper commands which individually enable doing an account migration: creating a new account using service auth and old DID; migrating all the data; swapping identity around
- a helper command that automates all this (starting with a PLC op auth token)

The hope is that the migrate command should work in common cases, but the individual commands allow cleaning up in the case of partial success, and doing more complex stuff.

Missing pieces as future work:

- `did:web` functionality
- detection of current progress and only continue work that needs continuing (eg, don't try to re-create new account if it already exists and can auth with provided credentials)
- proper support for inserting a recovery key when signing PLC op
- more safety validation and guardrails

### DIFF
--- a/cmd/goat/account_migrate.go
+++ b/cmd/goat/account_migrate.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"bytes"
+	"strings"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"time"
@@ -20,6 +22,12 @@ var cmdAccountMigrate = &cli.Command{
 	Usage: "move account to a new PDS. requires full auth.",
 	Flags: []cli.Flag{
 		&cli.StringFlag{
+			Name:     "pds-host",
+			Usage:    "URL of the new PDS to create account on",
+			Required: true,
+			EnvVars:  []string{"ATP_PDS_HOST"},
+		},
+		&cli.StringFlag{
 			Name:     "new-handle",
 			Required: true,
 			Usage:    "handle on new PDS",
@@ -30,6 +38,20 @@ var cmdAccountMigrate = &cli.Command{
 			Required: true,
 			Usage:    "password on new PDS",
 			EnvVars:  []string{"NEW_ACCOUNT_PASSWORD"},
+		},
+		&cli.StringFlag{
+			Name:     "plc-token",
+			Required: true,
+			Usage:    "token from old PDS authorizing token signature",
+			EnvVars:  []string{"PLC_SIGN_TOKEN"},
+		},
+		&cli.StringFlag{
+			Name:  "invite-code",
+			Usage: "invite code for account signup",
+		},
+		&cli.StringFlag{
+			Name:  "new-email",
+			Usage: "email address for new account",
 		},
 	},
 	Action: runAccountMigrate,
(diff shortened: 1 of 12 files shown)


cmd/goat/account_plc.go
```diff
@@ -0,0 +1,169 @@
...
+	} else if err != nil {
+		return err
+	}
+
+	fileBytes, err := os.ReadFile(opPath)
+	if err != nil {
+		return err
+	}
+
+	var op json.RawMessage
```
brianolson, line 156: this doesn't do anything, `json.RawMessage` _is_ `[]byte`

bnewbold: hmm, does it not validate that the bytes are valid JSON when doing this unmarshal? I do also like having the type wrapper, and it allows passing through to the outbound marshal step below in a way that wouldn't work if it was just `[]byte`.

cmd/goat/account_migrate.go
```diff
@@ -0,0 +1,263 @@
...
+			slog.Info("transfered blob", "cid", blobCID, "size", len(blobBytes))
+		}
+		if listResp.Cursor == nil || *listResp.Cursor == "" {
+			break
+		}
+		blobCursor = *listResp.Cursor
+	}
+
+	// display migration status
+	// TODO: should this loop? with a delay?
```
dholms, line 209: yeah you could run this in a loop in the background to give like progress bars for blob uploads or something, but otherwise you can just do it in between steps to ensure that things finished correctly

cmd/goat/account_migrate.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil
+	}
+
+	var unsignedOp IdentitySignPlcOperation_Input
+	if err = json.Unmarshal(credsBytes, &unsignedOp); err != nil {
+		return fmt.Errorf("failed parsing PLC op: %w", err)
+	}
+	unsignedOp.Token = &plcToken
+
+	// TODO: add some safety checks here around rotation key set intersection?
```
dholms, line 235: The new PDS does some checks on your behalf as well. But an extra check here wouldn't hurt. The TS script I wrote doesn't do any extra checks but before we start encouraging other folks to use this script en mass we should probably make it a _bit_ more resilient

cmd/goat/bsky_prefs.go
```diff
@@ -0,0 +1,92 @@
...
+	ctx := context.Background()
+
+	xrpcc, err := loadAuthClient(ctx)
+	if err == ErrNoAuthSession {
+		return fmt.Errorf("auth required, but not logged in")
+	} else if err != nil {
+		return err
+	}
+
+	// TODO: does this crash with unsupported preference types?
```
dholms, line 43: it _shouldn't_

I think the client is more accepting than the server. I believe the PDS will reject requests for preferences not in the `app.bsky` namespace. Which also means that users migrating off of our PDSs shouldn't run into any unsupported prefs

cmd/goat/bsky_prefs.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	var prefsArray []appbsky.ActorDefs_Preferences_Elem
+	err = json.Unmarshal(prefsBytes, &prefsArray)
+	if err != nil {
+		return err
+	}
+
+	// TODO: does this clobber unsupported preference types?
```
dholms, line 83: It shouldn't! But like above, it will reject any preference types that aren't `app.bsky`

cmd/goat/bsky_prefs.go
bnewbold: this is an indigo thing: the default de-serialization will throw an error if it encounters a `$type` which isn't a Lexicon schema known at compile-time.

cmd/goat/bsky_prefs.go
bnewbold: this is partially an indigo issue: unlike in typescript, go data is deserialized in to a struct by default; re-serializing this discards and data/fields which are not in the codegen'd struct. This includes off-Lexicons fields; open union members from unknown Lexicons; and data from fields in newer versions of the Lexicon since the client code was generated.

bnewbold:
Updated comments in the code. Going to merge now to have this basic functionality available.

See also: https://whtwnd.com/bnewbold.net/3l5ii332pf32u

PR #773, opened 2024-10-04
possible improvement of json.RawMessage instead of LexiconTypeDecoder for `unknown` data

bnewbold:
The indigo codebase supports a couple areas of use:

- generic atproto infrastructure code, like the relay, which doesn't care about record schemas
- application code, which does care about application schemas

When writing application code, there is an expectation that all the relevant Lexicon types are known and have pre-generated golang struct types. In that context, LexiconTypeDecoder as a wrapper mostly makes sense. There are some issues with round-tripping (decoding and re-encoding doesn't always yield the same data, if there are new or unexpected fields), but we can ignore that for now.

For the later situation, `LexiconTypeDecoder` doesn't work. If we are only dealing with CBOR blocks (like the relay or raw repo manipulation), it doesn't really matter, can just just avoid serialization, or use the `atproto/data` package for JSON/CBOR round-tripping. But sometimes we are dealing with HTTP APIs and the Lexicon-generated API helpers, and we need to deal with data of unknown type. This can be `unknown`, or it can be open unions. In this case, `LexiconTypeDecoder` doesn't really work.

This PR proposes one way of handling `unknown`: using `json.RawMessage` (which is basically just `[]byte` under the hood). This allows passing arbitrary data around loss-less.

A related approach is: https://github.com/bluesky-social/indigo/pull/485

Having tried this approach, I think something like the above might be better: make it so LexiconTypeDecoder can deal with unknown types.

A related issue we'll need to address is "open unions" when we don't know the type. Maybe we always have an "Other" field in the enum struct and stuff LexiconTypeDecoder in there.

https://github.com/bluesky-social/atproto/pull/2860


repomgr/repomgr.go
```diff
@@ -705,14 +709,18 @@ func (rm *RepoManager) BatchWrite(ctx context.Context, user models.Uid, writes [
...
+				op.Record = ltd.Val
 			}
 
 			ops = append(ops, op)
 		case w.RepoApplyWrites_Update != nil:
 			u := w.RepoApplyWrites_Update
 
-			cc, err := r.PutRecord(ctx, u.Collection+"/"+u.Rkey, u.Value.Val)
+			var ltd lexutil.LexiconTypeDecoder
+			if err = ltd.UnmarshalJSON(*u.Value); err != nil {
```
ericvolp12, line 720: Is this adding a marshalling step in repomanager where we didn't have one before? Wondering if there will be performance impacts on this cause this is critical path on relay write-path iirc.

automod/capture/fetch.go
```diff
@@ -36,8 +37,12 @@ func FetchAndProcessRecord(ctx context.Context, eng *automod.Engine, aturi synta
 		return fmt.Errorf("expected a CID in getRecord response")
 	}
 	recCID := syntax.CID(*out.Cid)
+	var ltd lexutil.LexiconTypeDecoder
```
brianolson, line 40: I think it would be cleaner to use util.JsonDecodeValue()
maybe make a utility function that returns `(cbg.CBORMarshaler, error)` if we want better than `(any, error)` ?

PR #779, opened 2024-10-29
Add avatar to `AccountMeta`

haileyok:
smol change. couple of new rules need avatar cids. currently have to fetch from public api before the rule processes in a separate request which is meh, since we have the info here already.

### DIFF
--- a/automod/engine/account_meta.go
+++ b/automod/engine/account_meta.go
@@ -25,6 +25,7 @@ type AccountMeta struct {
 
 type ProfileSummary struct {
 	HasAvatar   bool
+	Avatar      *string
 	Description *string
 	DisplayName *string
 }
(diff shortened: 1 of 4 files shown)


bnewbold:
I think I was originally worried about redis cache size?

all the `AccountMeta` gets serialized/deserialized using some magic in the redis caching library we use. maybe we should take just the avatar CID as a string and include that? how much metadata about the avatar do you need for your rule? eg, might want banner also if looking for impersonation signals.

or, in the other direction, just include the entire profile view?

haileyok:
ah, didn't realize that. grabbing just the cid would be fine and i have a little function for that already. let me move that over to this pr.

bnewbold (approved):

PR #785, opened 2024-10-31
lexgen 20241030

bnewbold:
@brianolson: there was a lexgen issue in here that I manually worked around. we have a new lexicon which uses a ref directly for the API endpoint "input" schema, which maybe hasn't be exercised before.


brianolson:
Is there a way to fix this so that the next person to run `make lexgen` doesn't undo that patch? e.g. I check this out and run `make lexgen` and ...
```
-func SetUpsertSet(ctx context.Context, c *xrpc.Client, input *SetDefs_Set) (*SetDefs_SetView, error) {
+func SetUpsertSet(ctx context.Context, c *xrpc.Client, input *SetUpsertSet_Input) (*SetDefs_SetView, error) {
```

bnewbold:
@brianolson I have a PR trying to fix lexgen for this here: https://github.com/bluesky-social/indigo/pull/788

PR #792, opened 2024-11-06
Lexicon resolution

bnewbold:
Adds a `ResolveNSID` method to `identity.BaseDirectory`. I think keeping DNS TXT resolution stuff together in the same package makes sense, even if this is a "lexicon" thing, not really an "identity" thing. Could consider renaming that method to `ResolveLexiconNSID` or similar, to clarify it is about NSIDs in the context of Lexicons, not "NSIDs in general" (though they are currently only defined in terms of Lexicons).

Adds a new `ResolvingCatalog` implementation of the `lexicon.Catalog` interface, which does resolution from the live network and adds to a local cache. Also `lexicon.ResolveLexiconData` low-level helper which goes from NSID to Lexicon record data.


bnewbold:
This is ready for review/merge

atproto/identity/nsid.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+)
+
+var (
+	ErrNSIDResolutionFailed = fmt.Errorf("NSID resolution mechanism failed")
+	ErrNSIDNotFound         = fmt.Errorf("NSID not associated with a DID")
+)
+
+// Resolves an NSID to a DID, as used for Lexicon resolution (using "_lexicon" DNS TXT record)
```
brianolson, line 17: no option for `https://*/.well-known/atproto-lexicon` ?

atproto/lexicon/catalog.go
```diff
@@ -72,46 +76,53 @@ func (c *BaseCatalog) AddSchemaFile(sf SchemaFile) error {
...
+		return err
+	}
+	if d.IsDir() {
+		return nil
+	}
+	if !strings.HasSuffix(p, ".json") {
+		return nil
+	}
+	slog.Debug("loading Lexicon schema file", "path", p)
+	f, err := os.Open(p)
```
brianolson, line 99: `os.Open()` won't work for the LoadEmbedFS() case

atproto/identity/nsid.go
bnewbold: nope. we discussed/debated and decided against. I think there is some discussion in: https://github.com/bluesky-social/atproto/discussions/3074

atproto/lexicon/catalog.go
bnewbold: ok, i'll add a test to cover that use-case

atproto/lexicon/resolve.go
```diff
@@ -0,0 +1,62 @@
...
+// The current implementation uses a naive 'getRepo' fetch to the relevant PDS instance, without validating MST proof chain.
+//
+// Calling code should usually use ResolvingCatalog, which handles basic caching and validation of the Lexicon language itself.
+func ResolveLexiconData(ctx context.Context, dir identity.Directory, nsid syntax.NSID) (map[string]any, error) {
+
+	baseDir := identity.BaseDirectory{}
+	did, err := baseDir.ResolveNSID(ctx, nsid)
+	if err != nil {
+		return nil, err
+	}
```
ericvolp12, line 26: Is there a reason we use the BaseDirectory here and not the identity directory? I guess to avoid adding ResolveNSID as a required method on the directory interface for now?

ericvolp12: Never mind I see we suggest you use `ResolvingCatalog` anyhow

atproto/lexicon/resolving_catalog.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func NewResolvingCatalog() ResolvingCatalog {
+	return ResolvingCatalog{
+		Base:      NewBaseCatalog(),
+		Directory: identity.DefaultDirectory(),
+	}
+}
+
+func (rc *ResolvingCatalog) Resolve(ref string) (*Schema, error) {
```
ericvolp12, line 26: Can this string be anything? Should it be a URL or something and should we validate that?

atproto/lexicon/resolve.go
bnewbold: yeah, basically putting that method on `BaseDirectory` to start. would probably make sense to add to caching and other implementations, and thus the interface, but doing baby steps.

atproto/lexicon/resolving_catalog.go
bnewbold: it is a "Lexicon reference", which is an NSID optionally followed by name-fragment. I'm planning to add a syntax parser for that identifier, but still hashing out the specification of the syntax (what we have written on atproto.com is probably too restrictive).

atproto/lexicon/catalog.go
bnewbold: think I got this fixed. thanks for the catch!

PR #803, opened 2024-11-13
make compaction worker count configurable

whyrusleeping:
Want to be able to tweak this manually so we can turn it higher during off-peak if necessary


ericvolp12 (approved):
LGTM

PR #805, opened 2024-11-13
relay api key

brianolson:
in case we need to play favorites to keep the system up, allow API keys with multiple tiers of priority and the ability to shed lower levels

### DIFF
--- a/bgs/admin.go
+++ b/bgs/admin.go
@@ -647,3 +647,114 @@ func (bgs *BGS) handleAdminRequestCrawl(e echo.Context) error {
 
 	return bgs.slurper.SubscribeToPds(ctx, host, true, true) // Override Trusted Domain Check
 }
+
+// Whoooo are the users in your neighborhood?
+type ApiKey struct {
+	// ApiKey goes in header: `Authorization: Bearer {}`
+	ApiKey string `json:"key"`
+	// Priority; more is more (unkeyed == 0)
+	Priority int32 `json:"prio"`
+}
+
+// set by
+type ApiKeySet struct {
+	Keys []ApiKey `json:"keys"`
+}
+
+// ApiKeys is the in-memory version
+type ApiKeys struct {
+	Keys map[string]ApiKey `json:"keys"`
+}
+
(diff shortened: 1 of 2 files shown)


bgs/admin.go
```diff
@@ -647,3 +647,114 @@ func (bgs *BGS) handleAdminRequestCrawl(e echo.Context) error {
...
+}
+
+// ApiKeys is the in-memory version
+type ApiKeys struct {
+	Keys map[string]ApiKey `json:"keys"`
+}
+
+func NewApiKeys() *ApiKeys {
+	out := new(ApiKeys)
+	out.Keys = make(map[string]ApiKey)
```
ericvolp12, line 671: Can we do this more idiomatically?
```suggestion
	out := &ApiKeys{Keys: make(map[string]ApiKey)}
```

bgs/bgs.go
```diff
@@ -87,6 +88,10 @@ type BGS struct {
 
 	// Management of Compaction
 	compactor *Compactor
+
+	apiKeys        atomic.Pointer[ApiKeys]
```
ericvolp12, line 92: Is there a reason we use this instead of a lock?

ericvolp12: We don't use these atomic semantics anywhere else and I'm a bit wary of throwing them in here now over using a standard mutex pattern.

brianolson: Ideally atomics are faster; there are a few large system NUMA cases where they are less good than usual, but I'm not sure they're ever slower than sync.Mutex ? I think in the optimistic case where there is no contention they're much faster.
They don't provide fairness of unlock to multiple lockers like sync.Mutex, but that doesn't matter here.

brianolson:
The good news is that I can demonstrate that it works. With no setup, works as present, no api key needed. With api key set uploaded, can set threshold priority up and down and allow/disallow various api keys (or unkeyed access).

The messy news is that it is currently stateless and I should do more work to add it to the local database. Otherwise the API keys need to be re-uploaded after any reboot.

PR #809, opened 2024-11-14
ident: have redis caching dir only log on 'write', not error

bnewbold:


brianolson (approved):
if we want to make hidden redis-cache downtime less hidden, add a counter (and maybe we'll add an alert on it incrementing too fast); otherwise LGTM

PR #810, opened 2024-11-15
relay splitter daemon + pebble persistence

brianolson:
stack PR on top of #807 
Add Pebble db based persister
clean up tests in events/


events/pebblepersist.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if seq < 0 {
+		// persist with longer key {prev 8 byte key}{time}{int32 extra counter}
+		pp.prevSeqExtra++
+		var key [20]byte
+		binary.BigEndian.PutUint64(key[:8], uint64(pp.prevSeq))
+		binary.BigEndian.PutUint64(key[8:16], nowMillis)
+		binary.BigEndian.PutUint32(key[16:], pp.prevSeqExtra)
+
+		err = pp.db.Set(key[:], blob, pebble.Sync)
```
whyrusleeping, line 51: error ignored

whyrusleeping: oh, we should just ignore -1 sequence number events (info, error, etc)

events/pebblepersist.go
```diff
@@ -0,0 +1,182 @@
...
+func (pp *PebblePersist) SetEventBroadcaster(broadcast func(*XRPCStreamEvent)) {
+	pp.broadcast = broadcast
+}
+
+func (pp *PebblePersist) GetLast(ctx context.Context) (*XRPCStreamEvent, error) {
+	iter, err := pp.db.NewIterWithContext(ctx, &pebble.IterOptions{})
+	if err != nil {
+		return nil, err
+	}
+	ok := iter.Last()
```
whyrusleeping, line 130: is the iter ordered by default?

events/pebblepersist.go
```diff
@@ -0,0 +1,182 @@
...
+	expired := uint64(nowMillis - retention.Milliseconds())
+	iter, err := pp.db.NewIterWithContext(ctx, &pebble.IterOptions{})
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+	todel := make(chan []byte, 100)
+	go func() {
+		for xkey := range todel {
+			pp.db.Delete(xkey, nil)
```
whyrusleeping, line 168: is there a batch delete option? or a range delete?

pds/server.go
```diff
@@ -456,18 +457,7 @@ func (s *Server) HandleResolveDid(c echo.Context) error {
...
-	UpdatedAt   time.Time
-	DeletedAt   gorm.DeletedAt `gorm:"index"`
-	Handle      string         `gorm:"uniqueIndex"`
-	Password    string
-	RecoveryKey string
-	Email       string
-	Did         string `gorm:"uniqueIndex"`
-	PDS         uint
-}
+type User = pdsdata.User
```
whyrusleeping, line 460: feels like a superfluous change to include in this PR

events/pebblepersist.go
brianolson: Yup

pds/server.go
brianolson: Yeah, but I was working on the unit tests, and the tests were in a different package which was weird, so I fixed that and then had to move a data definition to fix a circular dependency.

events/pebblepersist.go
brianolson: Yes! rushed and didn't see it the first time. done.

brianolson:
TODO: I think we still have a TODO of using the pebble store's "get last entry" to set the sequence we fetch `?since={}` from the upstream firehose

brianolson:
merged as #818

PR #812, opened 2024-11-15
allow carstore to use multiple directories, round robin style

whyrusleeping:


cmd/bigsky/main.go
```diff
@@ -312,8 +317,16 @@ func runBigsky(cctx *cli.Context) error {
...
 
-	os.MkdirAll(filepath.Dir(csdir), os.ModePerm)
-	cstore, err := carstore.NewCarStore(csdb, csdir)
+	csdirs := []string{csdir}
+	if paramDirs := cctx.StringSlice("carstore-shard-dirs"); len(paramDirs) > 0 {
+		csdirs = paramDirs
+	}
+
+	for _, csd := range csdirs {
+		os.MkdirAll(filepath.Dir(csd), os.ModePerm)
```
brianolson, line 326: check errors?

brianolson (commented):
should really check mkdir error on startup, otherwise looking good

ericvolp12 (approved):
Think this LGTM

PR #814, opened 2024-11-16
:bug: Fix typo in moderation review state check

foysalit:


bnewbold (commented):
The typo fix is good (thanks), but I want to verify how the strings come through before moving to the full NSID ref (I vaguely remember being surprised that they were coming back this way in API responses).

foysalit:
hmm... I just tested this against staging and here's the logline I got 

```
{"time":"2024-11-18T15:45:11.572752Z","level":"WARN","msg":"reviewState","did":"did:plc:3mfcl5hdb3v7gycyp47ol35m","err":"tools.ozone.moderation.defs#reviewEscalated"}
```

from `logger.Warn("reviewState", "err", *rd.Moderation.SubjectStatus.ReviewState)`

bnewbold (approved):
you were totally right here! sorry this took forever to confirm

PR #816, opened 2024-11-16
add a user cache on the bgs

whyrusleeping:
should cut ~3% of all DB usage on the prod relay

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -528,7 +528,7 @@ type User struct {
 	Tombstoned bool
 
 	// UpstreamStatus is the state of the user as reported by the upstream PDS
-	UpstreamStatusT string `gorm:"index"`
+	UpstreamStatus string `gorm:"index"`
 
 	lk sync.Mutex
 }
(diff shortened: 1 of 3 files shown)


bgs/bgs.go
```diff
@@ -136,6 +140,8 @@ func NewBGS(db *gorm.DB, ix *indexer.Indexer, repoman *repomgr.RepoManager, evtm
 	db.AutoMigrate(models.PDS{})
 	db.AutoMigrate(models.DomainBan{})
 
+	uc, _ := lru.New[string, *User](1_000_000)
```
bnewbold, line 143: probably doesn't need to be that big, our DAU is only like... oh right, haha.

bnewbold (approved):
I don't have a great intuition for what the impact of all this mutex churn will be, but it is way way less than a db transaction so this seems like a great easy win.

bgs/bgs.go
whyrusleeping: yeah, this is our *hourly* active user count lately lol

PR #823, opened 2024-11-19
let ATP_PLC_HOST override default plc host

lun-4:
I've been working on spinning up my own bluesky stack (including the did:plc reference server) and found out that only some commands on `goat` support a `-plc-directory` setting, but most don't because they just use `DefaultDirectory()` instead (via `resolveIdent`). this led me to making a patch to support `ATP_PLC_HOST` globally so all goat commands or other atproto tooling can operate consistently without requiring a rebuild-from-source to change defaults


bnewbold:
I personally don't like the pattern of having env vars impact libraries in a way which is not legible or explicit to the overall process (eg, does not show up in manpage or CLI docs). This goes doubly when adding/removing/updating the behavior.

In this case I think there is a clear path for how commands can configure their own directory.

One way to make it easier to "override" inner defaults like this would be to have the default directory return a concrete type, instead of interface. Eg, `DefaultBaseDirectory()` which returns a `BaseDirectory` that can be mutated (and simply wrapped with a caching directory if desired).

PR #825, opened 2024-11-19
Rewind cursor a bit before reconnection for some relay hosts to smooth over reconnection

ericvolp12:
We see some jerk when the Relay restarts caused by high concurrency of event processing for some high traffic hosts, where some in-progress event handlers die while the cursor is advanced ahead of them and the next event for that repo triggers a catch-up/resync of the repo. This change rewinds the cursor a bit on reconnect to smooth out that process.


whyrusleeping (approved):
For the record, i think theres an edge case where if the relay restarts too fast, we might end up ratcheting the cursor back multiple times and persisting it

PR #827, opened 2024-11-19
non-archival relay

whyrusleeping:
real jank right now, but doesnt keep all the repo data, just does basic event validation

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -1188,6 +1188,7 @@ func (bgs *BGS) handleRepoTombstone(ctx context.Context, pds *models.PDS, evt *a
 	}).Error; err != nil {
 		return err
 	}
+	u.SetTombstoned(true)
 
 	if err := bgs.db.Model(&models.ActorInfo{}).Where("uid = ?", u.ID).UpdateColumns(map[string]any{
 		"handle": nil,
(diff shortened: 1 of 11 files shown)


bgs/bgs.go
```diff
@@ -1155,6 +1155,7 @@ func (bgs *BGS) handleRepoTombstone(ctx context.Context, pds *models.PDS, evt *a
 	}).Error; err != nil {
 		return err
 	}
+	u.SetTombstoned(true)
```
whyrusleeping, line 1158: this is an irrelevant fix of some other thing

carstore/bs.go
```diff
@@ -606,7 +606,11 @@ func (ds *DeltaSession) CloseWithRoot(ctx context.Context, root cid.Cid, rev str
 	case *FileCarStore:
 		return ocs.writeNewShard(ctx, root, rev, ds.user, ds.seq, ds.blks, ds.rmcids)
 	case *NonArchivalCarstore:
-		return nil, ocs.updateLastCommit(ctx, ds.user, rev, root)
+		slice, err := blocksToCar(ctx, root, rev, ds.blks)
```
brianolson, line 609: case on type should probably be a new interface method that both implement

audreyt:
Fixed the lint error over at https://github.com/bluesky-social/indigo/pull/876

whyrusleeping:
Thanks @audreyt !

PR #830, opened 2024-11-19
automod: configurable circuit breaker thresholds

haileyok:


bnewbold (approved):
a Go thing is that code should be resilient when empty structs are passed. eg, an empty `EngineConfig{}` would have the `nil` value for all the config values. we should probably detect and handle that. this can be an issue for test code, third part code, etc.

but this should work for now

bnewbold:
test that is failing might be related to above. might want to check if values are nil/empty when doing the actual value checks, and insert a default there

PR #834, opened 2024-11-20
Revert "stream resync to indexer crawler"

bnewbold:
Reverts bluesky-social/indigo#826

(I pressed merge on the wrong PR)


brianolson:
no need for this, the prior PR merge was fine and doesn't need to be reverted

PR #836, opened 2024-11-21
Fix staticcheck SA2000, S1011 and S1039

egonelbre:
- fix staticcheck SA2000
- fix staticcheck S1011
- fix unnecssary fmt.Sprintf (staticcheck S1039)


brianolson (commented):
This looks all good! Thanks!

brianolson (approved):
re checked, still good, merging...

PR #838, opened 2024-11-21
events: document EventPersistence.Playback behavior for since parameter, fix PebblePersist accordingly

### LINKED ISSUE #837: Inconsistent (off-by-one) handling of `cursor` parameter for Firehose depending on `EventPersistence` implementation.
### The Situation

I copied code from Palomar to subscribe to the Firehose. That includes persisting the cursor of the last processed message to a database (and also periodically). *Note*: This is also used by, e.g. Rainbow when saving the cursor to disk.
I'm receiving a message with sequence number `N` and persist `N` to the database. On next startup, I read `N` from the database and pass it as `cursor=N` to `com.atproto.sync.subscribeRepos`. I noticed that the first message I receive has sequence number `N`, not `N+1`, which means I process that message twice.
This was an easy fix on my end (and I have a PR to fix `search/firehose.go`), but I was curious if there were other components affected, so I dug a bit deeper.

I didn't find good documentation for `com.atproto.sync.subscribeRepos`, but I did find this: https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/sync/subscribeRepos.json#L13
...which states `"The last known event seq number to backfill from."`.
My gut feeling tells me if I already _know_ the sequence number, I _don't_ want it included in the response, i.e., I want everything `> cursor`.

###  Checking the Implementation

The Firehose runs `EventPersistence.Playback(ctx, cursor, ...)` with the user-provided cursor: https://github.com/bluesky-social/indigo/blob/main/events/events.go#L360
Checking all implementations of that interface, I get:
- `DbPersistence` correctly queries for `seq > ?`: https://github.com/bluesky-social/indigo/blob/main/events/dbpersist.go#L379
- `DiskPersistence` ... not 100% sure, but I think this correctly seeks to an event with `evt.Seq > since` here: https://github.com/bluesky-social/indigo/blob/main/events/diskpersist.go#L245-L257
- `PebblePersist` **incorrectly** queries for `LowerBound: [8]byte{since}`: https://github.com/bluesky-social/indigo/blob/main/events/pebblepersist.go#L123
    The docs on that say:
    ```
    // LowerBound specifies the smallest key (inclusive) that the iterator will
	// return during iteration. If the iterator is seeked or iterated past this
	// boundary the iterator will return Valid()==false. Setting LowerBound
	// effectively truncates the key space visible to the iterator.
    ```
    The first element of the iterator is returned, i.e., this returns all events with `evt.Seq >= since`.
- `EventRingBuffer` (I think) correctly skips while `evt.Seq <= since`: https://github.com/bluesky-social/indigo/blob/main/splitter/ringbuf.go#L116-L118
- `MemPersister` is probably correct, if that `TODO` is telling the truth: https://github.com/bluesky-social/indigo/blob/main/events/persist.go#L76-L80
- `YoloPersister` does not support playback.

### Solutions

1. It would be nice to have documentation for `com.atproto.sync.subscribeRepos` that states whether it should return everything `> since` or `>= since`.
2.  The `EventPersistence` interface should have documentation to specify how `since` should be handled.
3. The implementations should all behave the same way.

I can submit PRs for `2` and `3`, probably. Let me know :)

mrd0ll4r:
This fixes #837

### DIFF
--- a/events/pebblepersist.go
+++ b/events/pebblepersist.go
@@ -118,7 +118,8 @@ func eventFromPebbleIter(iter *pebble.Iterator) (*XRPCStreamEvent, error) {
 
 func (pp *PebblePersist) Playback(ctx context.Context, since int64, cb func(*XRPCStreamEvent) error) error {
 	var key [8]byte
-	binary.BigEndian.PutUint64(key[:], uint64(since))
+	// Add 1 to the cursor because IterOptions.LowerBound is inclusive.
+	binary.BigEndian.PutUint64(key[:], uint64(since+1))
 
 	iter, err := pp.db.NewIterWithContext(ctx, &pebble.IterOptions{LowerBound: key[:]})
 	if err != nil {
--- a/events/persist.go
+++ b/events/persist.go
@@ -11,6 +11,8 @@ import (
 // Note that this interface looks generic, but some persisters might only work with RepoAppend or LabelLabels
 type EventPersistence interface {
 	Persist(ctx context.Context, e *XRPCStreamEvent) error
+
+	// Playback calls cb on persisted events with sequence number > since.
 	Playback(ctx context.Context, since int64, cb func(*XRPCStreamEvent) error) error
 	TakeDownRepo(ctx context.Context, usr models.Uid) error
 	Flush(context.Context) error


mrd0ll4r:
Scratch that, I'll investigate further...

PR #839, opened 2024-11-21
simplify CrawlDispatcher.mainLoop()

brianolson:
I am pretty sure that this is the same operation, but much simpler, so we don't confuse ourselves every time we try to read this code


brianolson:
found the bug in this, incomplete, will re-fix later

PR #840, opened 2024-11-21
add memcached for did cache to relay

brianolson:
tired of waiting for the did cache to warm up when relay restarts? wait no longer! put the cache in memcached (still also have local in-process cache, but backed by "L2" cache in memcached)

### DIFF
--- a/cmd/bigsky/main.go
+++ b/cmd/bigsky/main.go
@@ -190,6 +190,10 @@ func run(args []string) error {
 			EnvVars: []string{"RELAY_DID_CACHE_SIZE"},
 			Value:   5_000_000,
 		},
+		&cli.StringSliceFlag{
+			Name:    "did-memcached",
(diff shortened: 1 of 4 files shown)


ericvolp12 (approved):
LGTM, makes me wonder if we want the cache sizes to be configurable per-tier though.

brianolson:
memcached is configured when starting that service; the L1 DID cache is set by --did-cache-size (default 5 million entries)

bnewbold:
just for visiblity, we have some pretty similar code for caching DID and handle resolutions in the `atproto/identity` package (in redisdir sub-package): https://github.com/bluesky-social/indigo/tree/main/atproto/identity

that was written after relay was written, so relay code doesn't use it. would be good to get it migrated over though. if there are specific features needed in the identity package (eg, bypass handle resolution, because relay doesn't care about that), we can add them

PR #846, opened 2024-11-22
relay --time-seq for firehouse loosely bound to now

brianolson:


events/diskpersist.go
```diff
@@ -443,7 +447,16 @@ func (dp *DiskPersistence) doPersist(ctx context.Context, j persistJob) error {
 	b := j.Bytes
 	e := j.Evt
 	seq := dp.curSeq
-	dp.curSeq++
+	if dp.timeSequence {
+		nextSeq := time.Now().UnixMicro()
```
whyrusleeping, line 451: scroll down about 40 lines to discover why this may break the relay

brianolson: right. will need to keep a separate event counter if seq isn't it, and the playback logic needs some touchup too

brianolson: fixed fragmentation via event counter, and fixed replay

PR #853, opened 2024-11-29
syntax: ParseDID performance

bnewbold:
we are apparently burning a bunch of CPU on this regex in prod, so let's bypass for the most common case (valid did:plc).

Would be nice to have a benchmark demonstrating perf gains (or not).

I did some manually testing with the CLI helper and inserted `panic()` statements to confirm that valid DID PLC stays on fast path, and other DIDs don't.

Feedback welcome if there is a more idiomatic/performant golang way to do this... this is still doing unicode stuff, I assume, and maybe casting to bytes would be faster? though probably doesn't impact the actual hot path much. `strings.ContainsAny` or `strings.ContainsFunc` could be used here I guess but :shrug:

### DIFF
--- a/atproto/syntax/cmd/atp-syntax/main.go
+++ b/atproto/syntax/cmd/atp-syntax/main.go
@@ -22,6 +22,12 @@ func main() {
 			ArgsUsage: "<tid>",
 			Action:    runParseTID,
 		},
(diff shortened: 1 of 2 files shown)


whyrusleeping (approved):
Nice

PR #854, opened 2024-11-29
identity: hack flag to skip handle resolution

bnewbold:
Motivation here is to slash DNS lookups as part of cached DID resolution. As noted in the doc comment, we should loop back around and update the "Directory" interface/API to support this usecase better.

To actually deploy this in discover, we should:

- confirm that we don't actually use handle resolution anywhere in discover. this will always return DID lookups with "invalid handle"
- configure the redis directory "invalid handle TTL" to have the same duration as "hit" TTL
- set this flag to "true" on the inner/wrapped base directory

This will only impact the "resolve identity by DID" path, not any "resolve identity by handle" paths.


bnewbold:
note: now planning to *not* implement this method

bnewbold:
merged this functionality in a separate PR

PR #855, opened 2024-11-29
backfill: add failback to fetch CAR file from PDS (if relay errors)

bnewbold:
The main motivation here is that non-archival relay is returning a 4xx error for fetching repos, and we want to try fetching those from the actual PDS when that happens. This code adds a new code branch when a relay CAR fetch fails to do an identity lookup to find the account's PDS instance, and fetches the CAR from there.

For the search code specifically, it re-uses an existing identity directory, to reduce double-resolution.

This also refactors how fetch URLs are constructed to use just hostnames.

UPDATE: should probably *not* merge this to `main` until Jaz can review

### DIFF
--- a/backfill/backfill.go
+++ b/backfill/backfill.go
@@ -386,6 +386,7 @@ func (b *Backfiller) BackfillRepo(ctx context.Context, job Job) (string, error) 
 			slog.Warn("repo CAR fetch from PDS failed", "did", repoDID, "since", job.Rev(), "pdsHost", pdsHost, "err", err)
 			return "repo CAR fetch from PDS failed", err
 		}
+		slog.Info("repo CAR fetch from PDS successful", "did", repoDID, "since", job.Rev(), "pdsHost", pdsHost, "err", err)
 	}
 
 	numRecords := 0
--- a/backfill/backfill.go
+++ b/backfill/backfill.go
@@ -12,8 +12,11 @@ import (
 	"time"
 
 	"github.com/bluesky-social/indigo/api/atproto"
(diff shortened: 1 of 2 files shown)


whyrusleeping (approved):
This looks correct to me

backfill/backfill.go
```diff
@@ -356,10 +346,47 @@ func (b *Backfiller) BackfillRepo(ctx context.Context, job Job) (string, error)
...
+		log = log.With("retry_count", job.RetryCount())
+	}
+	log.Info(fmt.Sprintf("processing backfill for %s", repoDID))
+
+	// first try with Relay endpoint
+	r, err := b.fetchRepo(ctx, repoDID, job.Rev(), b.RelayHost)
+	if err != nil {
+		slog.Warn("repo CAR fetch from relay failed", "did", repoDID, "since", job.Rev(), "relayHost", b.RelayHost, "err", err)
+		// fallback to direct PDS fetch
+		ident, err := b.Directory.LookupDID(ctx, syntax.DID(repoDID))
```
futurGH, line 376: It'd be nice to have the option to include the PDS endpoint in `job` to skip this call (presumably with most ways to acquire DIDs in the first place, you'd also already have the associated PDS)

bnewbold: In this case I think they are coming down the firehose and the full identity isn't available.

But yeah a way to hook in pre-resolved identities further up the pipeline would be a good refactor of this overall package.

brianolson (approved):
LGTM

ericvolp12 (approved):
LGTM

PR #856, opened 2024-11-30
replace ipfs/go-log with stdlib log/slog

brianolson:


cmd/bigsky/main.go
```diff
@@ -292,6 +294,8 @@ func runBigsky(cctx *cli.Context) error {
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 
+	// TODO: set slog default from param/env
```
bnewbold, line 297: I think this will be an unexpected degradation from current config, so I would recommend setting this up for relay (bigsky) specifically. There is some example code in `palomar`'s `main.go`.

cmd/bigsky/main.go
```diff
@@ -486,7 +490,8 @@ func runBigsky(cctx *cli.Context) error {
 	// set up metrics endpoint
 	go func() {
 		if err := bgs.StartMetrics(cctx.String("metrics-listen")); err != nil {
-			log.Fatalf("failed to start metrics endpoint: %s", err)
+			log.Error("failed to start metrics endpoint", "err", err)
+			os.Exit(1)
```
bnewbold, line 494: any style reason to use `Exit(1)` vs `Exit(-1)`? I to the later by default but don't super care either way.

cmd/gosky/main.go
```diff
@@ -79,6 +79,9 @@ func run(args []string) {
 			EnvVars: []string{"ATP_PLC_HOST"},
 		},
 	}
+
+	// TODO: slog.SetDefault from param/env
```
bnewbold, line 83: way less important for `gosky` vs relay, IMO, though @whyrusleeping might care? might be helpful for some debug stuff

events/consumer.go
```diff
@@ -108,7 +109,14 @@ func (sr *instrumentedReader) Read(p []byte) (int, error) {
 	return n, err
 }
 
-func HandleRepoStream(ctx context.Context, con *websocket.Conn, sched Scheduler) error {
+// HandleRepoStream
+// con is source of events
+// sched gets AddWork for each event
+// log may be nil for default logger
```
bnewbold, line 115: :pinched_fingers:

events/dbpersist_test.go
```diff
@@ -16,15 +16,10 @@ import (
 	pds "github.com/bluesky-social/indigo/pds/data"
 	"github.com/bluesky-social/indigo/repomgr"
 	"github.com/bluesky-social/indigo/util"
-	logging "github.com/ipfs/go-log/v2"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 )
 
-func init() {
-	logging.SetAllLoggers(logging.LevelDebug)
```
bnewbold, line 25: I think there is a way to call `slog.SetDefaultLogger`, just to know that, but don't need to do it here and now

bnewbold (approved):
truely god's work :saluting_face:

cmd/bigsky/main.go
brianolson: I checked the source for stdlib log.Fatal and it's Exit(1)

cmd/gosky/main.go
brianolson: I have a second PR coming with this, at least for relay

cmd/bigsky/main.go
brianolson: I have a second PR coming with this

events/dbpersist_test.go
brianolson: a little extra util code should set the logger at something that writes to `testing.T.Log()`

brianolson:
chaser: integrate with main()s to set log file path, rotation rules, log level from env vars
https://github.com/brianolson/indigo/pull/2

Issue #857, opened 2024-11-30
Sonar is crashing due to an invalid memory address or nil pointer dereference


jrevillas:
After running for a few days, my Sonar instance is crashing with the following error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xa19ec0]

goroutine 68 [running]:
github.com/bluesky-social/indigo/sonar.(*Sonar).HandleRepoCommit(0x4000380300, {0xe8f160, 0x4000a6aa20}, 0x40007d5970)
	/app/sonar/sonar.go:230 +0xa60
github.com/bluesky-social/indigo/sonar.(*Sonar).HandleStreamEvent(0x4000380300, {0xe8f198, 0x400048bdb0}, 0x40007cd9e0)
	/app/sonar/sonar.go:111 +0x1130
github.com/bluesky-social/indigo/events/schedulers/sequential.(*Scheduler).AddWork(0x40000b4480, {0xe8f198, 0x400048bdb0}, {0x78f?, 0x60?}, 0x40007cd9e0)
	/app/events/schedulers/sequential/sequential.go:51 +0x70
github.com/bluesky-social/indigo/events.HandleRepoStream({0xe8f198?, 0x400048a140?}, 0x4000316580, {0xe878b0, 0x40000b4480})
	/app/events/consumer.go:202 +0x618
main.Sonar.func5()
	/app/cmd/sonar/main.go:200 +0xb0
created by main.Sonar in goroutine 1
	/app/cmd/sonar/main.go:197 +0xaa0
```

The current implementation (feceb364) seems to be unable to process a commit:
```shell
$ cat sonar_cursor_bad.json | jq .
{
  "last_seq": 838239253,
  "last_seq_processed_at": "2024-11-29T17:47:47.946669887Z"
}
```

As a result, Sonar is unable to recover and make progress on the next run because the cursor is still pointing at a commit it can't process.

jrevillas:
I just found the problematic repository: `did:plc:5hdeww3gu2q2cqcnkmtiesg2`.

PR #859, opened 2024-11-30
laputa lives on

ellie-idb:
this is a *SUPER* wip PR of getting laputa working as a viable alternative to the TypeScript PDS. the motivation behind a lot of this is that people may want to run a PDS on low-power embedded devices (think: <1GB of RAM), such as myself. Go works great here, and this experimental PDS only uses about <100kb of RAM. perfect for my use cases!

lots to be done here, and lots of cleanup to be done here, but this is Just Good Enough :tm: to federate with the app view. off the top of my head: 
* tons of handlers are fully not implemented
* it's definitely not secure because it doesn't enforce invite codes 💀 
* echo is a very mid HTTP handling library. sorry guys. i just don't love it.
* i'm probably not emitting some events in the right sequence, namely the #account event
* the PDS is emitting events to the firehose that are causing upstream consumers to crash (see: #857, https://bsky.app/profile/tom.frontpage.team/post/3lc3u6lalps2a).
* misc errors seen in development with relays failing to accept some repo ops due to a base mismatch
* subscribeRepos currently doesn't backfill cursors, meaning that all events emitted before a consumer connects are silently dropped (VERY BAD)

let me know if there's any interest in this! i'd love to see some of this work get upstreamed 😄


bnewbold:
I would recommend just forking: grab the laputa code, call it something else, and develop in a separate repo. I'm leaning towards just entirely removing laputa from this repo all together.

PR #861, opened 2024-12-03
better get-record

whyrusleeping:


bnewbold (approved):
This looks fine as a first step. You might also need to handle non-struct record types, aka "generic data fetch", there is some code in `goat` to do that. I hope/intend to break that out in to an api helper package

PR #862, opened 2024-12-04
fix RWLock usage

brianolson:
this needs to be a full lock, not just an R-lock because _sometimes_ it writes to the value


ericvolp12 (approved):
LGTM

PR #866, opened 2024-12-05
slog setup util for indigo main() functions

brianolson:
add integration into main()s for slog, set level, output file, log rotation rules, text or json
also connect to ipfs log used by some dependencies.


brianolson:
this replaces https://github.com/brianolson/indigo/pull/2

PR #868, opened 2024-12-06
syntax: path parsing helper

bnewbold:
helper for parsing a repo path (combination of collection and rkey) in to separate NSID and RecordKey


ericvolp12 (approved):
LGTM

PR #869, opened 2024-12-06
schema-agnostic API helpers

bnewbold:
These are some variants of "typed" atproto sync/repo helpers from `api/atproto` which use `json.RawMessage` instead of the lexutil `LexiconTypeDecoder` wrapper.

These are mostly used from tools like `goat` which work with arbitrary record data, when we don't have lexgen types (or they aren't imported, or whatever). They can also be used for things like preferences, where the type in an open union is rapidly changing, and we don't want to "clobber" (drop) data from new lexicons the client code isn't familiar with.

You can see these in action: https://github.com/bluesky-social/indigo/pull/870

@whyrusleeping you might want to do similar things in `gosky`.

Big picture we'll maybe want to move lexgen in this direction at some point? Relatedly, having LexiconTypeDecoder and open unions have an "unknown schema" option with raw data (possibly `map[string]any` instead of `json.RawMessage`; the later only makes sense for API endpoints, not record data).

### DIFF
--- a/api/agnostic/actorgetPreferences.go
+++ b/api/agnostic/actorgetPreferences.go
@@ -1,0 +1,28 @@
+// Copied from indigo:api/atproto/actorgetPreferences.go
+
+package agnostic
+
+// schema: app.bsky.actor.getPreferences
+
+import (
+	"context"
+
+	"github.com/bluesky-social/indigo/xrpc"
+)
+
+// ActorGetPreferences_Output is the output of a app.bsky.actor.getPreferences call.
+type ActorGetPreferences_Output struct {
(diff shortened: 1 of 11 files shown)


api/agnostic/actorgetPreferences.go
```diff
@@ -0,0 +1,28 @@
+// Copied from indigo:api/atproto/actorgetPreferences.go
```
brianolson, line 1: api/bsky (in a few of these files)

brianolson (approved):
double check the comments on where things were copied from, a couple were bsky not atproto, otherwise makes sense

PR #876, opened 2024-12-13
Use keyed field in DbCID initialization.

audreyt:
fixes "struct literal uses unkeyed fields" in lint.


whyrusleeping:
Hah, thanks!

PR #879, opened 2024-12-17
feat(cmd/goat): XRPC subcommand

bleonard252:
This PR introduces a generic XRPC subcommand, which allows executing XRPC queries and procedures on your PDS, or via the `atproto-proxy` header (which is set with the `--proxy` flag). This should aid in the development of XRPC endpoints, and it may be helpful for power users.

A screenshot of the command being run in a terminal, with the proxy DID redacted.

```
NAME:
   goat xrpc - use an XRPC endpoint

USAGE:
   goat xrpc [command options] <nsid> [paramKey=paramValue...]

DESCRIPTION:
   Execute an XRPC query or procedure via your PDS. Procedure inputs should be provided via standard input (i.e. a pipe, or redirection with <).

OPTIONS:
   --proxy value            the service to proxy to, referred to by its DID and service ID
   --type value             the MIME type of the request body for a procedure (default: "application/json")
   --procedure, --post, -p  execute an XRPC procedure (POST request) (default: false)
   --query, -q, --get, -g   execute an XRPC query (GET request) (default: false)
   --help, -h               show help
```

I set procedure and query to be options instead of subcommands so that lexicon resolution can be added later. If a lexicon resolves for a given NSID, the lexicon will tell it which one (procedure or query) it needs to be.

### DIFF
--- a/cmd/goat/main.go
+++ b/cmd/goat/main.go
@@ -38,6 +38,7 @@ func run(args []string) error {
 		cmdSyntax,
 		cmdCrypto,
 		cmdPds,
+		cmdXRPC,
 	}
 	return app.Run(args)
 }
--- a/cmd/goat/xrpc.go
+++ b/cmd/goat/xrpc.go
@@ -18,6 +18,7 @@ import (
 var cmdXRPC = &cli.Command{
 	Name: "xrpc",
 	Usage: "use an XRPC endpoint",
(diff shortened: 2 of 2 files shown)


bnewbold:
I've had something like this on my TODO list for a while! I haven't dug in super deep to your implementation here, but I really like the HTTPie "interface" and copied that in an earlier atproto CLI tool: https://gitlab.com/bnewbold/adenosine/-/blob/main/extra/adenosine.1.md

PR #884, opened 2024-12-19
Add PDS host argument to blob export

ngerakines:
This PR introduces the `--pds-host` argument to the `blob export` command chain.

This was needed in order to export the blobs from an old PDS when by DID document was pointing to the new one already.

### DIFF
--- a/cmd/goat/blob.go
+++ b/cmd/goat/blob.go
@@ -78,14 +78,14 @@ func runBlobExport(cctx *cli.Context) error {
 		return err
 	}
 
-	pds_host := cctx.String("pds-host")
-	if pds_host == "" {
-		pds_host = ident.PDSEndpoint()
+	pdsHost := cctx.String("pds-host")
+	if pdsHost == "" {
+		pdsHost = ident.PDSEndpoint()
 	}
 
 	// create a new API client to connect to the account's PDS
 	xrpcc := xrpc.Client{
-		Host: pds_host,,
+		Host: pdsHost,
 	}
 	if xrpcc.Host == "" {
 		return fmt.Errorf("no PDS endpoint for identity")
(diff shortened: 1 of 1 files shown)


cmd/goat/blob.go
```diff
@@ -73,9 +78,14 @@ func runBlobExport(cctx *cli.Context) error {
 		return err
 	}
 
+	pds_host := cctx.String("pds-host")
+	if pds_host == "" {
+		pds_host = ident.PDSEndpoint()
+	}
+
```
ngerakines, line 85:
```suggestion
	pdsHost := cctx.String("pds-host")
	if pdsHost == "" {
		pdsHost = ident.PDSEndpoint()
	}

```

cmd/goat/blob.go
```diff
@@ -28,6 +28,11 @@ var cmdBlob = &cli.Command{
 					Aliases: []string{"o"},
 					Usage:   "directory to store blobs in",
 				},
+				&cli.StringFlag{
+					Name:    "pds-host",
+					Usage:   "URL of the PDS to export blobs from (overrides DID doc)",
+					EnvVars: []string{"ATP_PDS_HOST"},
```
bnewbold, line 34: nit pick, but this could collide with an env var with the same name used for login, in a confusing way. basically we use `ATP_PDS_HOST` to hint login/auth. I think we should use `ATP_BLOB_HOST` here, and `blob-host` as the arg name. Or maybe `ATP_PDS_BLOB_HOST`, but I like the shorter `ATP_BLOB_HOST` in the public API.

bnewbold (approved):
left a nit request, but otherwise helpful and looks good!

cmd/goat/blob.go
bnewbold: also, I guess if a DID was provided (for the account), and `ATP_BLOB_HOST` was provided, we could just skip identity resolution? that could help with situations where the identity is broken for whatever reason.

ngerakines: Yeah, given that, maybe the easiest thing to do would be to just remove the env va?
```suggestion
```

PR #886, opened 2024-12-22
feat(automod): refactor slur processing

lizthegrey:
The old automod code relied upon cut-pasting a bunch of `[hhhhhh][eeeeeee][iiiiiiil]{2}[oooooooooo0]` strings around when the regexes could be a lot more readable by separating the unicode substitution and badword checking steps. This attempts to unify the code into one place.


devankestel (approved):

bnewbold:
this looks interesting/promising, but it will probably be January (2025) before we can take a real look

lizthegrey:
> this looks interesting/promising, but it will probably be January (2025) before we can take a real look

Enjoy your holiday break! There's no urgency, this is just maintainability of future badword detection rather than any urgent need to add more badwords.

lizthegrey:
Welcome back, happy new year @bnewbold

PR #889, opened 2024-12-24
use syntax.ParseRepoPath when appropriate

bnewbold:
This helper was added recently. Rip out old copy-pasta versions and start actually using it.


ericvolp12 (approved):
Sure

PR #891, opened 2024-12-30
scylla blockstore

brianolson:
also toy sqlite blockstore

this is a squash-rebase of #882

### DIFF
--- a/carstore/README.md
+++ b/carstore/README.md
@@ -1,0 +1,41 @@
+# Carstore
+
+Store a zillion users of PDS-like repo, with more limited operations (mainly: firehose in, firehose out).
+
+## [ScyllaStore](scylla.go)
+
+Blocks stored in ScyllaDB.
+User and PDS metadata stored in gorm (PostgreSQL or sqlite3).
+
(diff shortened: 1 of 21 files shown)


carstore/README.md
```diff
@@ -0,0 +1,41 @@
...
+FileCarStore was the first production carstore and used through at least 2024-11.
+
+## [SQLiteStore](sqlite_store.go)
+
+Experimental/demo.
+Blocks stored in trivial local sqlite3 schema.
+Minimal reference implementation from which fancy scalable/performant implementations may be derived.
+
+```sql
+CREATE TABLE IF NOT EXISTS blocks (uid int, cid blob, rev varchar, root blob, block blob, PRIMARY KEY(uid,cid))
```
ericvolp12, line 25: This will store all blocks for a user in a single partition FYI, you'd want `PRIMARY KEY((uid, cid)) for each block to be its own partition.

ericvolp12 (approved):
LGTM assuming it doesn't impact operation of the NA Relay

PR #893, opened 2025-01-01
bump golang version to 1.23; fix slog syntax issues

bnewbold:
This is just periodic maintenance, there isn't anything specific in 1.23 we want/need.

go v1.24 is expected in February 2025; we usually wait a month or two before upgrading (though we didn't get around to 1.23 until now).

The newer `go vet` turned up some slog syntax issues.


brianolson (approved):
I wish the golang version was single-sourced from go.mod or that we at least had a script to check consistency; otherwise LGTM

PR #897, opened 2025-01-03
:sparkles: Add new typedefs from ozone api changes

foysalit:


bnewbold (approved):
Thanks!

We really will need to update how codegen works for those huge parameter lists; they should be structs which means `nil` will be used by default and we don't need to update a bunch of service/command code every time. that might not happen for a bit though.

PR #900, opened 2025-01-07
iterated client SDK

bnewbold:
This is just an interface sketch for now. Three distinct things:

- `APIClient` interface for atproto clients which talk to a single HTTP service. `BaseAPIClient` is an implementation with modular auth implementations
- `NetClient` interface (could be `SyncClient`?) for resolving records or repo CAR files from the overall network, eg by resolving a DID to PDS, or talking to an archival mirror
- `PDSClient` (not included yet) which has extra affordances on top of `APIClient` to do service proxying, labeler headers, etc.


atproto/client/apiclient.go
```diff
@@ -0,0 +1,195 @@
+package client
```
jcalabro, line 1: Re: a question from the PR description:

> should this package be named atproto/atclient instead of atproto/client? "client" is such a generic term

We could also just have this be at the `gitub.com/bluesky-social/indigo/atproto` level. `atproto.APIClient` sounds nice IMO. Not really sure I feel strongly though.

Another thought was `github.com/bluesky-social/indigo/atproto/at`, that way `at` is a relatively unique name. 🤷

atproto/client/apiclient.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Creates a simple APIClient for the provided host. This is appropriate for use with unauthenticated ("public") atproto API endpoints, or to use as a base client to add authentication.
+//
+// Uses [http.DefaultClient], and sets a default User-Agent.
+func NewAPIClient(host string) *APIClient {
+	return &APIClient{
+		Client: http.DefaultClient,
+		Host:   host,
+		Headers: map[string][]string{
+			"User-Agent": []string{"indigo-sdk"},
```
jcalabro, line 46: Should this come with the short version number like we use in other places?

atproto/client/apiclient.go
```diff
@@ -0,0 +1,195 @@
...
+	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+		var eb ErrorBody
+		if err := json.NewDecoder(resp.Body).Decode(&eb); err != nil {
+			return &APIError{StatusCode: resp.StatusCode}
+		}
+		return eb.APIError(resp.StatusCode)
+	}
+
+	if out == nil {
+		return nil
```
jcalabro, line 86: We should discard the response body in this case (same with `Post`, `LexDo`, etc.)

atproto/client/apiclient.go
```diff
@@ -0,0 +1,195 @@
...
+	resp, err := c.Do(ctx, req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+		var eb ErrorBody
+		if err := json.NewDecoder(resp.Body).Decode(&eb); err != nil {
+			return &APIError{StatusCode: resp.StatusCode}
```
jcalabro, line 119: Might be helpful to just set the full response body string as the `Message` on this field? Otherwise, the response body is being eaten by our SDK in the case that we return something other than valid JSON.

atproto/client/apiclient.go
```diff
@@ -0,0 +1,195 @@
...
+	return nil
+}
+
+// Full-featured method for atproto API requests.
+//
+// TODO: this does not currently parse API error response JSON body to [APIError], thought it might in the future.
+func (c *APIClient) Do(ctx context.Context, req *APIRequest) (*http.Response, error) {
+
+	if c.Client == nil {
+		c.Client = http.DefaultClient
```
jcalabro, line 142: I get a bit sus when assigning to a pointer field in an environment where there could be multiple goroutines running (this line could be a data race)

I feel as though we should assume that the `c.Client` is non-nil and valid given that we supply one on `NewAPIClient`

atproto/client/apierror.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+type APIError struct {
+	StatusCode int
+	Name       string
+	Message    string
+}
+
+func (ae *APIError) Error() string {
+	if ae.StatusCode > 0 && ae.Name != "" && ae.Message != "" {
```
jcalabro: Control flow could be simplified if we lift the status code check out

```go
if ae.StatusCode > 0 {
    // .. all the other ifs
}
return "API request failed"
```

atproto/client/lexclient.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+
+	var body io.Reader
+	if bodyobj != nil {
+		if rr, ok := bodyobj.(io.Reader); ok {
+			body = rr
+		} else {
+			b, err := json.Marshal(bodyobj)
```
jcalabro: Wondering if we can get in to a weird case down this code path where we marshal to json here internally, but a different `inpenc` was passed. It might be better for us to force the content type to json in this branch, and to only set it to inpenc if we don't take this branch?

atproto/client/params.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// Flexibly parses an input map to URL query params (strings)
+func ParseParams(raw map[string]any) (url.Values, error) {
+	out := make(url.Values)
+	for k := range raw {
+		switch v := raw[k].(type) {
+		case nil:
+			out.Set(k, "")
+		case bool:
```
jcalabro: For the bool, string, and integer primitives, and encoding.TextMarshaler types, can we just use `fmt.Sprint(v)`?

atproto/client/params.go
```diff
@@ -0,0 +1,58 @@
...
+			if v {
+				out.Set(k, "true")
+			} else {
+				out.Set(k, "false")
+			}
+		case string:
+			out.Set(k, v)
+		case int, uint, int8, int16, int32, int64, uint8, uint16, uint32, uint64:
```
jcalabro: Missing `uintptr`

atproto/client/params.go
```diff
@@ -0,0 +1,58 @@
...
+					case int, uint, int8, int16, int32, int64, uint8, uint16, uint32, uint64:
+						out.Add(k, fmt.Sprintf("%d", elem))
+					case encoding.TextMarshaler:
+						out.Add(k, fmt.Sprintf("%s", elem))
+					default:
+						return nil, fmt.Errorf("can't marshal query param type: %T", v)
+					}
+				}
+			} else {
+				return nil, fmt.Errorf("can't marshal query param type: %T", v)
```
jcalabro: Might be good to include the field name in this error to make it easier for our users?

atproto/client/apiclient.go
bnewbold: I thought about adding that, but 1) it adds a (small) dependency to this SDK library (most invocations of `versioninfo` are in application code, not library code) and 2) it creates a privacy concern by default, in that a git hash in the user-agent is a lot of bits of entropy narrowing down to a specific user, and is likely to get included in logs.

My general feeling is that app developers should be expected to set a User-Agent, overriding this, and can make their own privacy and versioning decisions.

atproto/client/apiclient.go
bnewbold: I'd be open to re-exporting symbols from the parent `atproto` package, but think the code should still be organized with sub-packages. An example of this is in `../automod/pkg.go`.

atproto/client/apiclient.go
bnewbold: doesn't the `defer resp.Body.Close()` cover that in all these cases?

atproto/client/apiclient.go
bnewbold: A couple thoughts on this:

* if it is an atproto service, and the request is an XRPC endpoint, the error response should be a JSON error body. anything else isn't really an `APIError`. eg, if you connect to a non-atproto service by accident and get HTML back, that seems like a distinct error type; or generally Ok to not try and capture as an `APIError` message
* I particularly don't want to capture a bunch of random HTML, or a huge string (like an exception trace) and include that in the error `Message`
* the one case i'd be sympathetic to capturing is a single line of `text/plain` text. but i'm not sure how common that is in the real world, and detecting and handling just that case feels a bit complex, and maybe even unexpected behavior

I think we've had a couple places where we returned `text/plain` strings as `echo` error messages; I think we should try to refactor and avoid doing that from our own atproto services.

atproto/client/apiclient.go
bnewbold: yeah, this also felt a bit weird to me and I almost didn't include it. it is assigning a fixed value (the default client), so I don't think race conditions would actually cause any problems? and even with `NewAPIClient`, i'd expect that folks might construct the struct manually (eg, this is semi-common with `http.Client{}` for better or worse).

But agree it is icky. It also feels weird to have a call like this mutate the struct at all.

I can think of two options (other than just leaving as-is):

- return a descriptive error
- use `http.DefaultClient` for the request, without updating the `APIClient` struct

any thoughts?

atproto/client/lexclient.go
bnewbold: good call! in particular that "application/json" wasn't getting set

atproto/client/params.go
bnewbold: oh nice, I hadn't used `Sprint()` before.

I interpreted this as keeping the code flow, but simplifying the individual `Sprintf` calls.

atproto/client/apiclient.go
jcalabro: Sounds fine in theory, but in practice we have layers in the middle of our atproto services. Say an haproxy starts returning a 503 - is that message going to be accurately captured, or are our clients going to lose that message?

Most RPC systems I've seen swallow errors they don't expect, and unexpected errors can happen. Failing to pass those error messages to clients leads to confusion during incidents.

atproto/client/apiclient.go
jcalabro: Eh...I'm fine with keeping it as-is over either of those options

atproto/client/apiclient.go
jcalabro: I don't think so. In the case that the user passes in `nil` for `out`, and we do in fact get an HTTP body back from the server, we need to drain the `resp.Body`, which we fail to do in this `return nil` case.

atproto/client/params.go
jcalabro: I meant that we could simplify control flow because all primitives as well as `TextMarshaller` could just be one `case` statement. Not that big of a deal though.

atproto/client/params.go
jcalabro: Was this resolved? I still don't see it in the list

jcalabro: Ugh nvm, github was showing me an old diff, this looks good

atproto/client/apiclient.go
jcalabro: Another [example](https://github.com/bluesky-social/indigo/pull/1059/files#diff-65c763652fe4ceddce6b73141a569d8fa8d9b79c55dd38b548cb1c2794d352dbR28) of when we're not getting a valid xrpc json body back, and all we'd have is the response code (no message)

atproto/client/apiclient.go
bnewbold: oh, right, good call!

atproto/client/params.go
bnewbold: got it.

I partially collapsed these: combined all the specific types in to one `case`, but left the `TextMarshaller` interface as a separate `case`.

atproto/client/apiclient.go
bnewbold: What is it that haproxy would send back though? Probably just the default status page, which doesn't really add anything beyond the status code, right? And this error handling code does pass back the status code.

I think dumping the full request body, which is often HTML, doesn't really help users of this library figure out what is going on. If there was a terse string giving more information about the error, which could be piped through and displayed to a user, that could help in some cases, but that isn't what usually happens. When I have debugged this sort of thing with our services, it is super important (critical!) to get back 4xx error "name" and message, and that is what these methods do. But 5xx error messages are usually not helpful or meaningful; it is necessary to go to the server logs which generated the response and look for a trace or more info there.

This overall issue is the reason `Do()` doesn't try to parse error responses in to `APIError`. Though it should probably be more consistent.

I do want to get this right: it is a pain when libraries swallow necessary debugging info. But I also don't think this package should return the full HTTP response or something like that: i'm not sure what calling code would even do in that situation. Truncate the first few hundred bytes? I think for cases there is a reverse proxy in the mix, and an error occurred at that layer, passing back just the status code is basically the best we can do.

bnewbold: An example I thought of where a proxy swallowing errors is annoying is limits on request body size or overall header size. Those are meaningful messages for debugging, and probably come back as HTTP 400, which isn't very informative.

I'm still not sure how or what info to extract and pass back to calling code. Maybe we should use verbose logging of a snippet of non-JSON response body?

Also not sure if this needs to be addressed up-front with this package, or if it can be improved down the line. If we need to change the function signature of these methods then that does push towards fixing up front.

jcalabro: I don't think this would necessitate changing any function signatures, so yeah we can kick the can on it. I'm not sure why we can't just read the first N bytes of the response body and stick it in message field though? The thought being more information is better than none.

I.e. if you get back an haproxy error log, you know the problem is in haproxy.

bnewbold: Mulled this over some more. I'd be open to sticking the first KByte or so of text in an additional field on `APIError`, if the content type is `text/plain` or `text/html`.

But I also think this could be done as a follow-on to this PR, maybe after getting some real-world experience with debugging this kind of issue.

I think this is the last point to discuss on this PR? I'd lean towards merging now.

jcalabro: I had this issue with [twirp](https://github.com/twitchtv/twirp) and it bit us in production several times. It's frustrating to debug an incident when your client is swallowing the HTTP response bodies in the case of unexpected 500 level errors.

Since it will be backwards compatible to add better diagnostics, it's not a blocker, but I would still advocate for making those response bodies accessible so our clients aren't left scratching their heads when things go wrong.

atproto/client/apiclient.go
jcalabro: Eh. No strong preference. Re-exporting symbols sounds not great IMO. Fine as is if you ask me

atproto/client/apiclient.go
bnewbold: created a tracking issue: https://github.com/bluesky-social/indigo/issues/1099

PR #904, opened 2025-01-08
allow admin to add PDS that are ssl or not

brianolson:
BGS and Slurper allow mixed SSL or not upstreams
-crawl-ssl-policy|RELAY_CRAWL_NONSSL = any|require|none|prod
/admin/pds/crawl can add either; public requestCrawl normally SSL only
BGS AllowPDSProxies for nested relay setups
-allow-pds-proxy|RELAY_ALLOW_PDS_PROXY = true

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -65,9 +65,9 @@ type BGS struct {
 
 	hr api.HandleResolver
 
-	// TODO: work on doing away with this flag in favor of more pluggable
-	// pieces that abstract the need for explicit ssl checks
-	ssl bool
+	ssl SlurperSSLStance
+
+	config BGSConfig
 
 	crawlOnly bool
 
@@ -117,12 +117,13 @@ type SocketConsumer struct {
 }
 
 type BGSConfig struct {
-	SSL                  bool
+	SSL                  SlurperSSLStance
 	CompactInterval      time.Duration
 	DefaultRepoLimit     int64
 	ConcurrencyPerPDS    int64
 	MaxQueuePerPDS       int64
 	NumCompactionWorkers int
+	VerboseAPILog        bool
 
 	// NextCrawlers gets forwarded POST /xrpc/com.atproto.sync.requestCrawl
 	NextCrawlers []*url.URL
@@ -130,7 +131,7 @@ type BGSConfig struct {
 
 func DefaultBGSConfig() *BGSConfig {
 	return &BGSConfig{
-		SSL:                  true,
+		SSL:                  SlurperRequireSSL,
 		CompactInterval:      4 * time.Hour,
 		DefaultRepoLimit:     100,
 		ConcurrencyPerPDS:    100,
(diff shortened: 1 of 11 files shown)


bgs/bgs.go
```diff
@@ -941,7 +944,9 @@ func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *event
 			return fmt.Errorf("rebase was true in event seq:%d,host:%s", evt.Seq, host.Host)
 		}
 
-		if host.ID != u.PDS && u.PDS != 0 {
+		if bgs.config.AllowPDSProxies {
```
ericvolp12, line 947: Would be neat if we preconfigured the allowed hosts here via env var or something to avoid having the relay connected to itself or weird cases like that.

bnewbold:
Ok with this for now to get things unblocked, but would love to un-jank all this more in the future, including all the `wss://` vs `https://` wrangling throuhgout the codebase.

Maybe storing hostnames separately?

Also a bit worried if it becomes possible to consume from the same host both with SSL and without.

brianolson:
major tweak: 'allow relay' is now a per-source _admin_ action. No more global flag.

snuck this bit of data in next to the rate-limit admin stuff
'rate limit' admin stuff could grow to 'configure PDS/source', or could keep it broken up into API endpoint per function (block, crawl, ratelimit, allow-relay, etc)

also the dashboard now uses the admin version of `requestCrawl` instead of the public xrpc version

bgs/bgs.go
```diff
@@ -941,7 +943,9 @@ func (bgs *BGS) handleFedEvent(ctx context.Context, host *models.PDS, env *event
 			return fmt.Errorf("rebase was true in event seq:%d,host:%s", evt.Seq, host.Host)
 		}
 
-		if host.ID != u.PDS && u.PDS != 0 {
+		if host.RelayAllowed {
+			// don't check that source is canonical PDS, allow intermediate relays
+		} else if host.ID != u.PDS && u.PDS != 0 {
```
ericvolp12, line 948: What are the implications here of skipping the `createExternalUser` bit? Do we end up not creating a row for users that are being echoed by a Relay?

brianolson: yeah
image

brianolson: this is a little weird, as it turns out that narelay _does_ keep a little bit of per-user state so that it can emit a warning if there's a discontinuity in a user's feed ( See #915 )

ericvolp12: I think we'd still probably want to keep track of the users being proxied by a Relay so we can ban them and/or change their upstream status and/or count them right?

brianolson: mehhhh, ok, this needs some more consideration, there's _some_ useful stuff inside `.createExternalUser()` and I can't just throw out the path of code that was under "user from weird place, wat"

brianolson:
this PR may be obsoleted by other ongoing work, but not going to close it _yet_

PR #905, opened 2025-01-09
simple proxy of listRepos to rainbow upstream

brianolson:


brianolson:
supercedes #892

whyrusleeping (approved):
LGTM, not ideal but maybe we can do some caching later

PR #908, opened 2025-01-15
add GET /admin/repo/takedowns

brianolson:
API gap, we could add a repo takedown, and repeal it, but we couldn't list what was active

### DIFF
--- a/bgs/admin.go
+++ b/bgs/admin.go
@@ -116,6 +116,51 @@ func (bgs *BGS) handleAdminReverseTakedown(e echo.Context) error {
 	return nil
 }
 
+type ListTakedownsResponse struct {
+	Dids   []string `json:"dids"`
+	Cursor int64    `json:"cursor,omitempty"`
+}
+
+func (bgs *BGS) handleAdminListRepoTakeDowns(e echo.Context) error {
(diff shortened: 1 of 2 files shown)


bgs/admin.go
```diff
@@ -116,6 +116,51 @@ func (bgs *BGS) handleAdminReverseTakedown(e echo.Context) error {
...
+		var did string
+		err := rows.Scan(&id, &did)
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "oops").WithInternal(err)
+		}
+		out.Dids = append(out.Dids, did)
+		out.Cursor = id
+	}
+	if len(out.Dids) < limit {
+		out.Cursor = 0
```
ericvolp12, line 159: Should we omit the cursor from the response when it's empty? Not sure if it makes sense to return a cursor 0 since that's still valid.

brianolson: I don't feel super strogly; it would take a bit of rework from 'greater than previous last id' to 'greater-equal 0 or (last id plus one)'

ericvolp12: Ah okay omitempty makes that work

PR #910, opened 2025-01-16
automod: persist some firehose events to ozone

bnewbold:
This is some rework of: https://github.com/bluesky-social/indigo/pull/708

Specific features:

- automod engine feature for persisting record events (create/update/delete) to Ozone (if a rule requests this)
- rules to track ozone event on record subjects, and then persist future record updates to Ozone
- optional logic to persist account and identity events to ozone


bnewbold:
@foysalit summarizing changes I made over in this branch/PR:

- synced some recent changes from `main`
- renamed and shuffled some code around; eg "forward" not "reroute"
- made "record op" persist stuff just another record persist behavior, not something controlled via general event forwarding. so basically controlled by enabling a rule, not configuring hepa
- did keep "persist account history" as a hepa-level flag (not a rule), but also made that flag much more explicit about what it is doing.

I think this is ready to go; I will do a bit of testing, then merge this branch to `main` and close the other PR.

bnewbold:
at least in staging, the de-dupe query doesn't seem to work. I get:

```
XRPC ERROR 400: InvalidRequest: Invalid event type
```

when using event types like `tools.ozone.moderation.defs#identityEvent`

bnewbold:
another problem i'm seeing with "delete" record ops is that some ozone events require a "strong ref"; but with a deletion there is no CID for the record, so a strongref can't be created.

PR #912, opened 2025-01-20
Bolson/vm3 ex250120

brianolson:
this is mostly just to get the bulid system to build things for me to test deploy


bnewbold:
@brianolson can we close this?

PR #913, opened 2025-01-20
make PDS table unique on host name

brianolson:
fix old data that had dups on PDS host name

### DIFF
--- a/bgs/bgs.go
+++ b/bgs/bgs.go
@@ -141,7 +141,10 @@ func DefaultBGSConfig() *BGSConfig {
 
 func NewBGS(db *gorm.DB, ix *indexer.Indexer, repoman *repomgr.RepoManager, evtman *events.EventManager, didr did.Resolver, rf *indexer.RepoFetcher, hr api.HandleResolver, config *BGSConfig) (*BGS, error) {
 	logger := slog.Default().With("system", "bgs")
-	err := fixupDupPDSRows(db, logger)
+	// TODO: 2025-02-04 hopefully about a month from now this migration will have run everywhere and it can be deleted.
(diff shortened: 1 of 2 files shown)


bgs/bgs.go
```diff
@@ -202,6 +206,68 @@ func NewBGS(db *gorm.DB, ix *indexer.Indexer, repoman *repomgr.RepoManager, evtm
...
+				logger.Info("dup PDS", "host", host, "from", xPDSId, "to", minPDSId)
+				err = db.Exec("UPDATE users SET pds = ? WHERE pds = ?", minPDSId, xPDSId).Error
+				if err != nil {
+					return fmt.Errorf("failed to update user pds %d -> %d: %w", xPDSId, minPDSId, err)
+				}
+				err = db.Exec("UPDATE actor_infos SET pds = ? WHERE pds = ?", minPDSId, xPDSId).Error
+				if err != nil {
+					return fmt.Errorf("failed to update actor_infos pds %d -> %d: %w", xPDSId, minPDSId, err)
+				}
+				err = db.Exec("DELETE FROM pds WHERE id = ?", xPDSId).Error
```
ericvolp12, line 261: Do we want to do these in a transaction?

brianolson: kinda? I put it effectively into relay 'boot code' before threads of work are started. I guess the bad case is multiple `bigsky` processes pointed at the same PostgreSQL, but I bet it handles that badly in _many_ ways.

brianolson:
migration wrapped in a transaction, note TODOs and how we should delete this migration later when the INFO log shows that all the deployments are starting up cleanly.

PR #914, opened 2025-01-21
MST implementation with support for operation inversion

bnewbold:
This is fairly similar to `indigo/mst`, but written from scratch in a more go-idiomatic style, with the intent of supporting record op inversion (for the "inductive firehose" proposal).

Goal is to have something that passes full interop tests with TypeScript codebase, and can validate commits on the firehose. Would take more perf work after that, and probably review, to get this in a state where it could be in the hot path in prod.


bnewbold:
Interesting: tests are passing for me locally (including clearing test caches), but failing in CI. Will need to dig in to that.

An open design question is how to split "repo" code from "MST" code. Most of this comes down to managing records to/from CAR slices, repo exports, blockstores, etc. And concept of "record operations": is that an MST concept, or a repo concept? Is testing inversion of ops something that happens as a batch from repo code, or by MST code?

bnewbold:
The MST part of this PR is complete enough, and I think the `mst.Tree` API is stable enough, that i'm marking this ready to review.

The included `repo-tool` command can read the prod firehose and at least partially validate MST stuff.

The repo parts, including verifying commit signatures, are close, but I can complete that in a follow-up PR.

atproto/repo/sync.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, op := range ops {
+		if err := mst.InvertOp(&invTree, &op); err != nil {
+			// print the *non-inverted* tree
+			//mst.DebugPrintTree(repo.MST.Root, 0)
+			return nil, err
+		}
+	}
+	// TODO: compare against previous commit for this repo?
+	_, err = invTree.RootCID()
+
```
brianolson, line 103: this would be the part where we use the new firehose `prev` data?

atproto/repo/mst/operation.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (op *Operation) IsDelete() bool {
+	if op.Value == nil && op.Prev != nil {
+		return true
+	}
+	return false
+}
+
+// Mutates the tree, returning a full `Operation`
+func ApplyOp(tree *Tree, path string, val *cid.Cid) (*Operation, error) {
```
brianolson, line 38: why not member of Tree? `someTree.ApplyOp(path, val)` ?

brianolson: Why isn't this two functions? `someTree.InsertCid(val)` and `someTree.RemovePath(path)`

brianolson (commented):
broad pattern comment: lots of things that I would expect to be member functions of structs are instead top level functions that take the struct as first argument

atproto/repo/sync.go
brianolson: found it in follow-on PR

atproto/repo/mst/car.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if err := bs.Put(ctx, blk); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	if len(br.Roots) < 1 {
+		return nil, nil, fmt.Errorf("CAR file missing root CID")
+	}
+	commitCID := br.Roots[0]
```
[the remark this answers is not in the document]
bnewbold: this file moved up a level, but added the check there

atproto/repo/mst/operation.go
bnewbold: The Operation code is a bit funky because it is sort of a separate type/concept from the MST. maybe this code belongs better in the `atproto/repo` package? This is why I didn't want to put the methods on `Tree` directly; they don't feel like part of the `Tree` API.

I think it is helpful to keep the `Operation` abstraction around distinct from the lower-level insert/remove methods. This is all clearer at the `Repo` abstraction layer, so i'll move the code up to there.

PR #915, opened 2025-01-21
narelay: drop noisy warning on no prior user data

brianolson:
this fixed like 90% of WARN messages on a narelay I was running


carstore/nonarchive.go
```diff
@@ -130,7 +138,10 @@ func (cs *NonArchivalCarstore) NewDeltaSession(ctx context.Context, user models.
 		return nil, err
 	}
 
-	if since != nil && *since != lastShard.Rev {
+	if lastShard == nil {
+		// ok, no previous user state to refer to
+		lastShard = &commitRefZero
```
brianolson, line 143: this is where we _don't_ emit a warning message two lines later

ericvolp12 (approved):
sure

PR #917, opened 2025-01-23
remove automod "notifications event" support

bnewbold:
The idea was to integrate the automod framework with other services
which do push notifications. This never happened, and this code is
dead/untested, so removing for now.

This kind of functionality might be useful (we have lots of interaction
spam!) and could be pulled back in from git history if we ever need it.


foysalit (approved):
looks good!

PR #926, opened 2025-02-02
goat: update to v3 of urfav/cli

bnewbold:
Transition guide here: https://cli.urfave.org/migrate-v2-to-v3/

Overall updates are good/fine, but the big win and motivation here is finally having CLI arg parsing order be more flexible.

Eg:

```shell
# --did is a boolean flag

goat resolve --did atproto.com
goat resolve atproto.com --did
```

In v2, only the first is allowed, while with v3 the later works.

I'll probably do this update/refactor across all our commands soon.

### DIFF
--- a/cmd/goat/account.go
+++ b/cmd/goat/account.go
@@ -11,14 +11,14 @@ import (
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/bluesky-social/indigo/xrpc"
 
-	"github.com/urfave/cli/v2"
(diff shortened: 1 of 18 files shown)


brianolson (approved):
a little tedious to change but probably better?

brianolson:
cli v3 is still "beta1", maybe wait for full 3.0.0

bnewbold:
this got out of date, going to close until there is consensus to go forward

PR #928, opened 2025-02-03
Hydrate DeleteRecords events with deleted Record

### LINKED ISSUE #927: Context for deleted records on Firehose (`com.atproto.sync.subscribeRepos`)
I'm not certain this is the right repo to raise this request; I'm raising it here since indigo implements the Firehose, and this request should not require any changes the the API.

Note: all of my examples will be using the Jetstream JSON schema, but this applies to the CBOR `com.atproto.sync.subscribeRepos` too.

### The Problem

When an atproto record is deleted, the firehose event for the deletion references only the `rkey` of the event being deleted. This requires that, in order to handle the deletion, the user must store every record (of that type) for the users being monitored, even if their use-case otherwise does not require it.

#### Example

A simple "follower count" widget, that shows the current follower count for a user.

When an `app.bsky.graph.follow` record is _created_, it includes the `did` of the user being followed, via `commit.record.subject`, enabling the widget to simply increment its count when the DID matches the user being monitored.

However, when a follow record is _deleted_, it only includes the `rkey`. In order to decrement your counter, _you would need to store every matching follow record_, and do look-ups against the `rkey`.

This is probably fine for a single user, but it doesn't scale as you want to monitor the follower counts of more and more users. You ultimately need to store the entire follower graph, just to keep an accurate by-user count.

Worse: if you do try to store the entire follower graph, that's not possible via just the Firehose: you'd need to "backfill" all previous follow records from the HTTP API, which due to rate-limits would likely take _days_ already, and will only get worse.

This is the case irrespective of the type of record being deleted, which is why it's an atproto issue, not a bsky schema issue.

## The Solution

atproto already allows commit objects to contain an _optional_ `Record` field, which is populated for `create` and `update` operations.

**My request is, when a record is being _deleted_, the existing `record` field should contain the full record being deleted.**

This would solve the problem, without any schema change.

nicktelford:
To enable Firehose users to action deleted records without necessarily storing the records being deleted (e.g. by decrementing counters, etc.), we include the record being deleted in the Firehose.

This is only enabled when `hydrateRecords` is enabled, as per `UpdateRecords`.

Currently, BigSky has `hydrateRecords` set to `false`. We set this to `true` so that the Firehose provided by BigSky hydrates records for `update` and `delete` events.

Closes #927

### DIFF
--- a/cmd/bigsky/main.go
+++ b/cmd/bigsky/main.go
@@ -482,7 +482,7 @@ func runBigsky(cctx *cli.Context) error {
 		if err := ix.HandleRepoEvent(ctx, evt); err != nil {
 			slog.Error("failed to handle repo event", "err", err)
 		}
-	}, false)
+	}, true)
 
 	prodHR, err := api.NewProdHandleResolver(100_000, cctx.String("resolve-address"), cctx.Bool("force-dns-udp"))
 	if err != nil {
--- a/repomgr/repomgr.go
+++ b/repomgr/repomgr.go
@@ -312,6 +312,23 @@ func (rm *RepoManager) DeleteRecord(ctx context.Context, user models.Uid, collec
 	}
 
 	rpath := collection + "/" + rkey
+
+	op := RepoOp{
+		Kind:       EvtKindDeleteRecord,
+		Collection: collection,
+		Rkey:       rkey,
+	}
+
+	if rm.hydrateRecords && rm.events != nil {
+		ocid, rec, err := r.GetRecord(ctx, rpath)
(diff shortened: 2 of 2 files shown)


nicktelford:
I'm not entirely convinced that this is enough. Specifically, I don't think changing `hydrateRecords = true` in BigSky is the correct way to have the Firehose hydrate delete records. This is my first attempt at understanding the Indigo codebase, so I may have misunderstood a few things.

bnewbold:
I commented on the linked issue; will probably close this PR at some point.

PR #941, opened 2025-02-13
Add pagination to relay dash

ericvolp12:


ts/bgs-dash/src/components/Dash/Dash.tsx
```diff
@@ -1290,6 +1299,101 @@ const Dash: FC<{}> = () => {
             </tbody>
           </table>
         </div>
+        {pdsList && pdsList!.length > pageSize && (
```
devinivy, line 1302: The `!` here and below _shouldn't_ be needed since `pdsList` being truthy should narrow the type.  But if the type checker is yelling for some reason without it then probably not worth wrestling with.

devinivy (approved):

PR #951, opened 2025-02-21
medsky radical trim of bigsky (relay)

brianolson:
sync 1.1 induction firehose


cmd/medsky/events/cbor_gen.go
```diff
@@ -0,0 +1,303 @@
+// Code generated by github.com/whyrusleeping/cbor-gen. DO NOT EDIT.
```
bnewbold, line 1: should probably update `indigo;gen/main.go` to keep this file up-to-date? especially as we have some tweaks and deprecations coming soon.

eg, we'll be nuking the `#handle`, `#migrate`, and `#tombstone` messages; and adding `#sync`

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+	admin.POST("/subs/killUpstream", bgs.handleAdminKillUpstreamConn)
+	admin.POST("/subs/setPerDayLimit", bgs.handleAdminSetNewPDSPerDayRateLimit)
+
+	// Domain-related Admin API
+	admin.GET("/subs/listDomainBans", bgs.handleAdminListDomainBans)
+	admin.POST("/subs/banDomain", bgs.handleAdminBanDomain)
+	admin.POST("/subs/unbanDomain", bgs.handleAdminUnbanDomain)
+
+	// Repo-related Admin API
+	admin.POST("/repo/takeDown", bgs.handleAdminTakeDownRepo)
```
bnewbold, line 284: doesn't need to happen now, but over time i'd like to move most of the admin endpoints to Lexicons. could be `com.atproto.admin.*`, or `io.bigsky.*` or similar. we can totally prototype and sketch as raw HTTP API endpoints, but once we start having interop (eg, Ozone sends takedowns to relay), or folks start independently running relays with their own tooling, I think it is good to formalize the APIs.

For example, this one (repo takedown, and reverseTakedown below) should probably be the existing `com.atproto.admin.updateSubjectStatus` endpoint.

(this isn't a priority yet, just bookmarking some of these)

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+			if sendHeader {
+				ctx.Response().WriteHeader(500)
+			}
+		}
+	}
+
+	// TODO: this API is temporary until we formalize what we want here
+
+	e.GET("/xrpc/com.atproto.sync.subscribeRepos", bgs.EventsHandler)
+	e.GET("/xrpc/com.atproto.sync.requestCrawl", bgs.HandleComAtprotoSyncRequestCrawl)
```
bnewbold, line 260: it is just an XRPC "procedure", so we should only support "POST"

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+	admin.POST("/subs/unbanDomain", bgs.handleAdminUnbanDomain)
+
+	// Repo-related Admin API
+	admin.POST("/repo/takeDown", bgs.handleAdminTakeDownRepo)
+	admin.POST("/repo/reverseTakedown", bgs.handleAdminReverseTakedown)
+	admin.GET("/repo/takedowns", bgs.handleAdminListRepoTakeDowns)
+
+	// PDS-related Admin API
+	admin.POST("/pds/requestCrawl", bgs.handleAdminRequestCrawl)
+	admin.GET("/pds/list", bgs.handleListPDSs)
```
bnewbold, line 290: "listing PDS instances" in particular i'd like to make a public Lexicon endpoint. I can take on the task of specifying it.

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+
+	e.GET("/xrpc/com.atproto.sync.subscribeRepos", bgs.EventsHandler)
+	e.GET("/xrpc/com.atproto.sync.requestCrawl", bgs.HandleComAtprotoSyncRequestCrawl)
+	e.POST("/xrpc/com.atproto.sync.requestCrawl", bgs.HandleComAtprotoSyncRequestCrawl)
+	e.GET("/xrpc/com.atproto.sync.listRepos", bgs.HandleComAtprotoSyncListRepos)
+	e.GET("/xrpc/com.atproto.sync.getLatestCommit", bgs.HandleComAtprotoSyncGetLatestCommit)
+	e.GET("/xrpc/_health", bgs.HandleHealthCheck)
+	e.GET("/_health", bgs.HandleHealthCheck)
+	e.GET("/", bgs.HandleHomeMessage)
+
```
bnewbold, line 267: doesn't have to be in this PR / iteration, but we'll want to have `getRepo` as an endpoint on the relay, which to start does an HTTP redirect to upstream (eg, PDS)

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+	if !bgs.ssl {
+		e.Use(middleware.LoggerWithConfig(middleware.LoggerConfig{
+			Format: "method=${method}, uri=${uri}, status=${status} latency=${latency_human}\n",
+		}))
+	} else {
+		e.Use(middleware.LoggerWithConfig(middleware.DefaultLoggerConfig))
+	}
+
+	// React uses a virtual router, so we need to serve the index.html for all
```
bnewbold, line 220: in theory we could spin out the dash UI to a separate microservice, like we do with Ozone (Ozone UI and Ozone Backend are separate processes, separate hostnames/ports, etc).

can keep it as-is if working fine, just mentioning in case it would simplify things

cmd/medsky/bgs/bgs.go
```diff
@@ -0,0 +1,1440 @@
...
+// on any platform, but there's no great way to weave this timeout without
+// adding another parameter to the (at time of writing) long signature of
+// NewServer.
+const serverListenerBootTimeout = 5 * time.Second
+
+type BGS struct {
+	db      *gorm.DB
+	slurper *Slurper
+	events  *events.EventManager
+	didd    identity.Directory
```
bnewbold, line 54: I usually use `dir` as the short variable name for directories; could reduce confusion with the old did directory code to update it?

cmd/medsky/main.go
```diff
@@ -0,0 +1,478 @@
...
+	//	} else {
+	//		prevResolver = mr
+	//	}
+	//
+	//	cachedidr = plc.NewCachingDidResolver(prevResolver, time.Hour*24, cctx.Int("did-cache-size"))
+	//}
+
+	// TODO: add memcached layer for shared external cache
+	baseDir := identity.BaseDirectory{
+		SkipHandleVerification: true,
```
bnewbold, line 335: yup yup!

bnewbold (commented):
Left a bunch of comments from browsing/skimming over the code.

I should do a more focused review of just the core inductive verification stuff, and the account lifecycle stuff: the actually important bits!

Will need to add handling of the `#sync` message once this PR lands:
https://github.com/bluesky-social/atproto/pull/3391

... which might happen later today (Friday), or might be monday.

util/cbor.go
```diff
@@ -1,12 +1,11 @@
 package util
 
 import (
-	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	mh "github.com/multiformats/go-multihash"
 )
 
-func CborStore(bs blockstore.Blockstore) *cbor.BasicIpldStore {
```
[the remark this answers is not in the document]
brianolson: upgrading to non-deprecated boxo is separately a good idea; but I noticed that a few places use IpldBlockstore and it's a smaller simpler interface which has all the parts we ever actually use, so propagating it seems safe.
A fourth option is to define our own Blockstore interface with the subset we care about, which matches parts of an external Blockstore interface, and maybe a util wrapper with no-op implementations if we need to pass it to an external API (which it seems we never need to do because we have brought enough CAR/MST pieces in-house).

cmd/medsky/events/cbor_gen.go
brianolson: This is `subscribeRepos` stuff. Ts shouldn't even be in an `events` package but next to the rest of the protocol stuff in `atproto`

cmd/medsky/bgs/bgs.go
brianolson: added 302 handler for getRepo

cmd/medsky/bgs/bgs.go
brianolson: it's kinda weird; I never build the UI in development; a point in favor of separable admin-UI would be to have one UI fan out to multiple relays

brianolson:
work moved to #961

PR #953, opened 2025-02-21
Fix Race Conditions In All Slurper.handleConnection Callbacks

jcalabro:
Hi there!

I was curious about your code and pulled it down to take a look. When I went to run the tests, I noticed that there was a single race condition, which is resolved with this commit.

The idea is that the cursor's value should only be updated after acquiring `activeSub`'s lock.

Also, a small suggestion, it might be a good idea to add the race detector to CI? Didn't want to push that up to y'all in a PR out of the blue though since that feels like bad open source form :smile:


jcalabro:
Yo @bnewbold, thoughts on this one from a few weeks ago? Also, thoughts on adding the race detector to tests CI?

jcalabro:
This is no longer relevant now that sync 1.1 is far along 🐎

PR #954, opened 2025-02-22
lexgen with sync v1.1 lexicons, and other tweaks (2025-02-21)

bnewbold:


bnewbold:
Got a :heavy_plus_sign: from why, going ahead with merge.

PR #955, opened 2025-02-23
identity package tweaks

bnewbold:
- add user-agent
- ensure we drain HTTP requests in more code paths (allows connections/streams to be re-used more efficiently)


ericvolp12 (approved):
LGTM

PR #957, opened 2025-02-25
syntax: more datetime examples

bnewbold:


bnewbold:
these have gotten rolled up in other PRs

PR #961, opened 2025-02-28
relay sync1.1

brianolson:
relay radical rehacking for sync1.1 induction firehose
drops all `carstore` code for keeping archive of all repos
drops all `indexer` code for running `getRepo` on all those repos
drops a bunch of other stuff that fell out being deprecated and unneeded.
adds sync1.1 protocol features for checking `#commit`.`prevData` and `#sync`
internally vendors a bunch of code that we should re-share back out to common code

formerly known as #951


cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+			Name:    "disk-persister-dir",
+			Usage:   "set directory for disk persister (implicitly enables disk persister)",
+			EnvVars: []string{"RELAY_PERSISTER_DIR"},
+		},
+		&cli.StringFlag{
+			Name:    "admin-key",
+			EnvVars: []string{"RELAY_ADMIN_KEY", "BGS_ADMIN_KEY"},
+		},
+		&cli.IntFlag{
+			Name:    "max-metadb-connections",
```
bnewbold, line 106: `max-db-connections`? "metadb" doesn't make sense as a differentiation any more

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+			EnvVars: []string{"MAX_METADB_CONNECTIONS"},
+			Value:   40,
+		},
+		&cli.StringFlag{
+			Name:    "resolve-address",
+			EnvVars: []string{"RESOLVE_ADDRESS"},
+			Value:   "1.1.1.1:53",
+		},
+		&cli.BoolFlag{
+			Name:    "force-dns-udp",
```
bnewbold, line 116: not used

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+			Name:    "admin-key",
+			EnvVars: []string{"RELAY_ADMIN_KEY", "BGS_ADMIN_KEY"},
+		},
+		&cli.IntFlag{
+			Name:    "max-metadb-connections",
+			EnvVars: []string{"MAX_METADB_CONNECTIONS"},
+			Value:   40,
+		},
+		&cli.StringFlag{
+			Name:    "resolve-address",
```
bnewbold, line 111: not used

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+			EnvVars: []string{"RESOLVE_ADDRESS"},
+			Value:   "1.1.1.1:53",
+		},
+		&cli.BoolFlag{
+			Name:    "force-dns-udp",
+			EnvVars: []string{"FORCE_DNS_UDP"},
+		},
+		&cli.IntFlag{
+			Name:    "max-fetch-concurrency",
```
bnewbold, line 120: not used (many of these options seem to be not used, i'm not going to check any more)

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+				attribute.Int64("ID", 1),
+			)),
+		)
+		otel.SetTracerProvider(tp)
+	}
+
+	return nil
+}
+
+func runBigsky(cctx *cli.Context) error {
```
bnewbold, line 255: `runRelay`?

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+		if err := setRelaySettingBool(db, "time-seq", pOpts.TimeSequence); err != nil {
+			return err
+		}
+	} else {
+		if pOpts.TimeSequence != storedTimeSeq {
+			return fmt.Errorf("time-seq stored as %v but param/env set as %v", storedTimeSeq, pOpts.TimeSequence)
+		}
+	}
+
+	dp, err := events.NewDiskPersistence(dpd, "", db, pOpts)
```
bnewbold, line 343: huh, I would have assumed we'd be switching to pebble persister instead of disk persister? this will be difficult to change once deployed (I assume) so important to decide soon

cmd/relay/repomgr/repomgr.go
```diff
@@ -0,0 +1,398 @@
...
+
+// VerifyCommitSignature get's repo's registered public key from Identity Directory, verifies Commit
+// hostname is just for metrics in case of error
+func (rm *RepoManager) VerifyCommitSignature(ctx context.Context, commit *atrepo.Commit, hostname string, hasWarning *bool) error {
+	if rm.directory == nil {
+		return nil
+	}
+	xdid, err := syntax.ParseDID(commit.DID)
+	if err != nil {
+		commitVerifyErrors.WithLabelValues(hostname, "sig1").Inc()
```
bnewbold, line 370: these `sig1`, `sig2` labels are pretty opaque; more helpful/meaninful names for human operators who don't maintain this code trying to figure out what is going on?

cmd/relay/repomgr/repomgr.go
```diff
@@ -0,0 +1,398 @@
...
+		return fmt.Errorf("bad car DID, %w", err)
+	}
+	ident, err := rm.directory.LookupDID(ctx, xdid)
+	if err != nil {
+		if rm.AllowSignatureNotFound {
```
bnewbold, line 375: should check if the `err` is an `ErrNotFound` before allowing here, right?

cmd/relay/repomgr/repomgr.go
```diff
@@ -0,0 +1,398 @@
...
+
+	log               *slog.Logger
+	inductionTraceLog *slog.Logger
+
+	directory identity.Directory
+
+	maxRevFuture       time.Duration
+	ErrRevTooFarFuture error
+
+	// AllowSignatureNotFound enables counting messages without findable public key to pass through with a warning counter
```
bnewbold, line 61: what is the use-case for this?

cmd/relay/repomgr/repomgr.go
```diff
@@ -0,0 +1,398 @@
...
+
+	return &RepoManager{
+		userLocks:         make(map[models.Uid]*userLock),
+		log:               slog.Default().With("system", "repomgr"),
+		inductionTraceLog: inductionTraceLog,
+		directory:         directory,
+
+		maxRevFuture:           maxRevFuture,
+		ErrRevTooFarFuture:     ErrRevTooFarFuture,
+		AllowSignatureNotFound: true, // TODO: configurable
```
bnewbold, line 37: should definitely not be `true` by default?

cmd/relay/events/consumer.go
```diff
@@ -0,0 +1,372 @@
...
+		switch header.Op {
+		case EvtKindMessage:
+			switch header.MsgType {
+			case "#commit":
+				var evt comatproto.SyncSubscribeRepos_Commit
+				if err := evt.UnmarshalCBOR(r); err != nil {
+					return fmt.Errorf("reading repoCommit event: %w", err)
+				}
+
+				if evt.Seq < lastSeq {
```
bnewbold, line 216: I think it should be  `evt.Seq <= lastSeq`, and if that happens the message should just be dropped, right? we don't want to get in a loop re-processing stale events if the upstream accidentally re-emits them?

cmd/relay/events/consumer.go
```diff
@@ -0,0 +1,372 @@
...
+				}
+
+				lastSeq = evt.Seq
+
+				if err := sched.AddWork(ctx, evt.Did, &XRPCStreamEvent{
+					RepoSync: &evt,
+				}); err != nil {
+					return err
+				}
+			case "#handle":
```
bnewbold, line 244: deprecated (also a bunch of the other event handlers)

cmd/relay/main.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	app.Flags = []cli.Flag{
+		&cli.BoolFlag{
+			Name: "jaeger",
+		},
+		&cli.StringFlag{
+			Name:    "db-url",
+			Usage:   "database connection string for BGS database",
+			Value:   "sqlite://./data/bigsky/bgs.sqlite",
```
bnewbold, line 70: is there still sqlite support? I thought that got dropped

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1376 @@
...
+			}
+		}
+	}
+
+	repoCommitsResultCounter.WithLabelValues(host.Host, "ok").Inc()
+	return nil
+}
+
+func (bgs *BGS) handleSync(ctx context.Context, host *models.PDS, evt *comatproto.SyncSubscribeRepos_Sync) error {
+	// TODO: actually do something with #sync event
```
bnewbold, line 1088: we should at least be updating the metadata for the account (DID) here, right?

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1376 @@
...
+		if err != nil {
+			return err
+		}
+		since = &sval
+	}
+
+	ctx, cancel := context.WithCancel(c.Request().Context())
+	defer cancel()
+
+	// TODO: authhhh
```
bnewbold, line 536: I don't think this is a TODO

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1376 @@
...
+		return nil
+	}
+
+	if ustatus == events.AccountStatusSuspended {
+		bgs.log.Debug("dropping commit event from suspended user", "did", evt.Repo, "seq", evt.Seq, "pdsHost", host.Host)
+		repoCommitsResultCounter.WithLabelValues(host.Host, "susu").Inc()
+		return nil
+	}
+
+	if ustatus == events.AccountStatusDeactivated {
```
bnewbold, line 981: also account status `throttled`

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1376 @@
...
+			return err
+		}
+
+		if subj.PDS != host.ID {
+			repoCommitsResultCounter.WithLabelValues(host.Host, "noauth").Inc()
+			return fmt.Errorf("event from non-authoritative pds")
+		}
+	}
+
+	if u.GetTombstoned() {
```
bnewbold, line 1009: I'm pretty sure we just don't / shouldn't have a concept of "tombstoned" accounts in the relay. If the DID is tombstoned, that is basically account status `deleted`

bnewbold (commented):
got through some review. I still want to look at:

- the actual generated postgresql schemas. getting those right seems important. the gorm models are still spread through many files, seems like they could all be in `cmd/relay/models/models.go`?
- account state and lifecycle
- rate-limiting

it feels like there is still a lot of low-hanging fruit for code that could just be ripped out? eg, all the deprecated event types. there seemed to be a lot of config arguments which are not used.

I was surprised the disk persister is still used, instead of the pebble persister used in rainbow.

cmd/relay/README.md
```diff
@@ -0,0 +1,339 @@
...
+atproto Relay Service
+===============================
+
+*NOTE: "Relays" used to be called "Big Graph Servers", or "BGS", or "bigsky". Many variables and packages still reference "bgs"*
+
+This is the implementation of an atproto Relay which is running in the production network, written and operated by Bluesky.
+
+In atproto, a Relay subscribes to multiple PDS hosts and outputs a combined "firehose" event stream. Downstream services can subscribe to this single firehose a get all relevant events for the entire network, or a specific sub-graph of the network. The Relay maintains a mirror of repo data from all accounts on the upstream PDS instances, and verifies repo data structure integrity and identity signatures. It is agnostic to applications, and does not validate data against atproto Lexicon schemas.
+
+This Relay implementation is designed to subscribe to the entire global network. The current state of the codebase is informally expected to scale to around 20 million accounts in the network, and thousands of repo events per second (peak).
```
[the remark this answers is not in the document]
brianolson: I re-read it just now and I think most of it is still valid and I deleted the obviously obsolete stuff.

events/consumer.go
```diff
@@ -126,13 +126,20 @@ func HandleRepoStream(ctx context.Context, con *websocket.Conn, sched Scheduler,
...
 		defer t.Stop()
+		failcount := 0
 
 		for {
 
 			select {
 			case <-t.C:
 				if err := con.WriteControl(websocket.PingMessage, []byte{}, time.Now().Add(time.Second*10)); err != nil {
 					log.Warn("failed to ping", "err", err)
+					failcount++
```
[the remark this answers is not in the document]
brianolson: this is not the active code, see cmd/relay/evenst/consumer.go

models/models.go
```diff
@@ -104,7 +104,7 @@ type FollowRecord struct {
 type PDS struct {
 	gorm.Model
 
-	Host       string
+	Host       string `gorm:"unique"`
```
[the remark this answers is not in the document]
brianolson: yeah, I should probably rollback most of the changes outside of `cmd/relay`

cmd/relay/events/consumer.go
```diff
@@ -0,0 +1,372 @@
...
+
+type RepoStreamCallbacks struct {
+	RepoCommit    func(evt *comatproto.SyncSubscribeRepos_Commit) error
+	RepoSync      func(evt *comatproto.SyncSubscribeRepos_Sync) error
+	RepoHandle    func(evt *comatproto.SyncSubscribeRepos_Handle) error
+	RepoIdentity  func(evt *comatproto.SyncSubscribeRepos_Identity) error
+	RepoAccount   func(evt *comatproto.SyncSubscribeRepos_Account) error
+	RepoInfo      func(evt *comatproto.SyncSubscribeRepos_Info) error
+	RepoMigrate   func(evt *comatproto.SyncSubscribeRepos_Migrate) error
+	RepoTombstone func(evt *comatproto.SyncSubscribeRepos_Tombstone) error
```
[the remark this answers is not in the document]
brianolson: I don't think I can just delete them, I think I'd need to add code to explicitly ignore them and maybe count obsolete messages as a warning.

cmd/relay/events/diskpersist.go
```diff
@@ -0,0 +1,983 @@
...
+	"github.com/bluesky-social/indigo/api/atproto"
+	"github.com/bluesky-social/indigo/cmd/relay/models"
+	arc "github.com/hashicorp/golang-lru/arc/v2"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"gorm.io/gorm"
+)
+
+type DiskPersistence struct {
```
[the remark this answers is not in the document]
brianolson: I think the disk persist may actually be more efficient. pebble is good at random access, but maybe a little mediocre at sequential access with rotating discard which this explicitly does perfectly.

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1376 @@
+package bgs
```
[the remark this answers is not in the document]
brianolson: it's still package bgs, just indigo/cmd/relay/bgs instead of indigo/bgs.
I think the happy path a couple weeks from now is to clobber cmd/relay/*** back out into the root indigo/*** level things

cmd/relay/main.go
brianolson: the big thing pebble persister really did was be _new_ code without any cruft, and it doesn't re-sequence events which is what rainbow wants. I think DiskPersistence is still efficient and good.

cmd/relay/repomgr/repomgr.go
brianolson: maybe either `ErrDIDNotFound` (404) or `ErrDIDResolutionFailed` (5xx) should both be allow, but just now looking at the error return paths in the identity code there are a lot of different error conditions without a clear categorization (maybe the other broad category is "some part was there but invalid"). I think for now at this level "it didn't work, continue anyway" might be an okay strategy that we could refine later?

cmd/relay/repomgr/repomgr.go
brianolson: PLC is down. New repo exists in PDS but hasn't forwarded DID to PLC yet?

cmd/relay/repomgr/repomgr.go
brianolson: I think we can maybe refine which versions of "not found" are allowed or not, but until then it has to broadly be on in prod and is at worst a warning.

cmd/relay/main.go
brianolson: it's not working right now because some transactions changed, but it's maybe not far off from being fixed?

cmd/relay/bgs/bgs.go
brianolson: yup. ooooold todo. cancelled.

cmd/relay/bgs/bgs.go
brianolson: yeah, old code, I basically haven't had time to update the account status flow yet, that part seemed to have been in flux up until last week

cmd/relay/events/consumer.go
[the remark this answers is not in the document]
bnewbold: I took a stab at removing these in the existing packages here:
https://github.com/bluesky-social/indigo/pull/967

Unknown event types should be ignored by default (so we can evolve the protocol), and I think most of the existing code/packages does that. We should also have fewer concerns about things like "old event types in persist storage" with this relay, which will have a fresh deploy? Keeping around the old types/hooks and just making all the functions a no-op seems fine though.

cmd/relay/events/diskpersist.go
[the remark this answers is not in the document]
bnewbold: ok, good to know and fine with me.

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+This is an atproto [https://atproto.com] relay instance, running the 'bigsky' codebase [https://github.com/bluesky-social/indigo]
+
+The firehose WebSocket path is at:  /xrpc/com.atproto.sync.subscribeRepos
+`
+
+func (bgs *BGS) HandleHomeMessage(c echo.Context) error {
+	return c.String(http.StatusOK, homeMessage)
+}
+
+type AuthToken struct {
```
bnewbold, line 344: Storing auth tokens in the database like this is old code.

Do we need this functionality? Seems like it could just be an env var, maybe with multiple values.

cmd/relay/bgs/models.go
```diff
@@ -0,0 +1,8 @@
+package bgs
+
+import "gorm.io/gorm"
+
+type DomainBan struct {
+	gorm.Model
+	Domain string
```
bnewbold, line 7: `unique` constraint on the domain?

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+	u.UpstreamStatus = v
+}
+
+func (u *User) GetUpstreamStatus() string {
+	u.lk.Lock()
+	defer u.lk.Unlock()
+	return u.UpstreamStatus
+}
+
+type UserPreviousState struct {
```
bnewbold, line 471: can this just be merged on to the existing `Users` table? I guess this table has higher churn and it might be good to keep that separate

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+	Uid models.Uid   `gorm:"column:uid;primaryKey"`
+	Cid models.DbCID `gorm:"column:cid"`
```
bnewbold, line 473: I think a better column name would be good; `data_cid`? `commit_data` or `commit_data_cid`?

cmd/relay/bgs/bgs.go
bnewbold: I think the rough expected behavior is:

- parse out the commit object
- verify the message matches the fields in the commit object
- fetch user metadata
    - if account isn't active, drop event
    - if rev is old/bad, drop event
- update user metadata with sync info
- emit the event

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+	if err != nil {
+		return
+	}
+	_ = bgs.didd.Purge(ctx, *ati)
+}
+
+// createExternalUser is a mess and poorly defined
+// did is the user
+// host is the PDS we received this from, not necessarily the canonical PDS in the DID document
+// TODO: rename? This also updates users, and 'external' is an old phrasing
```
bnewbold, line 1135: seems like a good time to rename this, can update comments around what it does.

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+	ban, err := bgs.domainIsBanned(ctx, durl.Host)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check pds ban status: %w", err)
+	}
+
+	if ban {
+		return nil, fmt.Errorf("cannot create user on pds with banned domain")
+	}
+
+	if peering.ID == 0 {
```
bnewbold, line 1183: I assume this is old code, maybe from the days when BGS had indexing and would "spider" to discover accounts? or if there is a bulk catch-up happening?

would be good if you could clarify what the expected behavior is here ("maybe this never happens").

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1375 @@
...
+
+	// TODO: request this users info from their server to fill out our data...
+	u := User{
+		Did:         did,
+		PDS:         peering.ID,
+		ValidHandle: false,
+	}
+
+	err = bgs.db.Transaction(func(tx *gorm.DB) error {
+		res := tx.Model(&models.PDS{}).Where("id = ? AND repo_count < repo_limit", peering.ID).Update("repo_count", gorm.Expr("repo_count + 1"))
```
bnewbold, line 1262: I don't really understand the `repo_count < repo_limit` clause here... is this how account limits are enforced? the increment fails if the clause filters out the relevant row?

my intuition is that the account should be created (inserted), but with local account state "throttled", if the overall PDS is over quota. instead of silently being dropped; the later is hard for external folks to debug.

bnewbold (commented):
it looks like adding per-account rate-limits would currently mean a few `slidingwindow.Limiter` in-process for each of tens of millions of accounts (and growing). do you think that will scale, for us and for others (eg, with less RAM)? if they could expire out of memory, then the total number of counters could be more like "DAU" than "MAU".

the limits we have discussed are total bytes of data, and number of record ops, per account (DID). and then also the number of "expensive" events (like a `#sync` which is not a no-op; or an `#identity` event), or "broken" events (MST does not invert, for example).

---

overall account lifecycle state tracking:

- should store "local" account state as a string, not boolean flags
- if local state is "active', then default to upstream state (which may or may not be "active"). otherwise, use local state
- could store an "until" timestamp in account status, for temporary throttled situations

cmd/relay/bgs/bgs.go
brianolson: yeah, there's an env var. currently it sets the db; we could just hold it in memory and stop bouncing off the db; I'd be pretty happy about that actually....

cmd/relay/bgs/bgs.go
brianolson: yeah, that's my motivation. I think Postgres works better with the high churn on separate smaller rows from the relatively low-churn data.

util/cbor.go
```diff
@@ -1,12 +1,11 @@
 package util
 
 import (
-	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	mh "github.com/multiformats/go-multihash"
 )
 
-func CborStore(bs blockstore.Blockstore) *cbor.BasicIpldStore {
+func CborStore(bs cbor.IpldBlockstore) *cbor.BasicIpldStore {
```
bnewbold, line 8: I think you should split out these changes to function signatures (here, and `indigo:mst` and `indigo:repo`) to a separate PR, if they aren't touched by the `indigo:cmd/relay` code.

events/consumer.go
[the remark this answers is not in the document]
bnewbold: if this doesn't impact the `indigo:cmd/relay` code, can you pull it out of this PR?

cmd/relay/bgs/bgs.go
brianolson: I did a bunch of reorg and commenting on this function as a whole and I think it's better now

brianolson:
force-push should be progress up to 59ae24fe15a1233763e1ff3f8c482ea11c2071fc sqashed and rebased onto main

cmd/relay/repomgr/repomgr.go
bnewbold: I think if the PLC client can connect to the PLC directory and authoritatively determined that the PLC just doesn't exist at all, we should just drop the event, right? I don't think we should have a mechanism where PDS instances can spam the relay database with bogus DIDs which have never existed in any form.

It is maybe different if there are other identity errors, like network error connecting to the directory, in which case the DID might or might not exist, and we might make a different call to accept the DID even if we don't know it's disposition.

cmd/relay/repomgr/repomgr.go
bnewbold: I think if PLC is down, and the identity isn't cached, the network is basically expected to pause? especially for firehose consumers, where they can catch up when it comes back up (presumably within minutes, PLC is critical infrastructure).

cmd/relay/bgs/bgs.go
bnewbold: might be hard to change this now, but I think at least a comment explaining what this is the CID of is important; we have referred to the CID of the entire commit object previously.

cmd/relay/bgs/bgs.go
bnewbold: I think this code got moved around, but this "throttled" state didn't get implemented yet.

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1306 @@
...
+	case events.AccountStatusTakendown:
+		if err := bgs.db.Model(Account{}).Where("id = ?", u.ID).Update("upstream_status", events.AccountStatusTakendown).Error; err != nil {
+			return fmt.Errorf("failed to set user taken down status: %w", err)
+		}
+		u.SetUpstreamStatus(events.AccountStatusTakendown)
+		// TODO: set Account takedown in db? -- bolson 2025
+	case events.AccountStatusDeleted:
+		// TODO: tweak model to mark user deleted? -- bolson 2025
+		if err := bgs.db.Model(&Account{}).Where("id = ?", u.ID).UpdateColumns(map[string]any{
+			"tombstoned":      true,
```
bnewbold, line 1251: the `tombstoned` and `handle` columns of this table/model are gone now.

we should treat `deleted` the same as all the other non-active account statuses, especially now that the relay doesn't hold a mirror/archive. accounts can enter the `deleted` state and then exit it and become `active` again, according to the spec.

cmd/relay/bgs/handlers.go
```diff
@@ -0,0 +1,203 @@
...
+			}(blob)
+		}
+	}
+
+	return s.slurper.SubscribeToPds(ctx, host, true, false, nil)
+}
+
+func (s *BGS) handleComAtprotoSyncListRepos(ctx context.Context, cursor int64, limit int) (*comatprototypes.SyncListRepos_Output, error) {
+	// Filter out tombstoned, taken down, and deactivated accounts
+	q := fmt.Sprintf("id > ? AND NOT tombstoned AND NOT taken_down AND (upstream_status is NULL OR (upstream_status != '%s' AND upstream_status != '%s' AND upstream_status != '%s'))",
```
bnewbold, line 111: `tombstoned` column is gone

cmd/relay/bgs/bgs.go
```diff
@@ -0,0 +1,1306 @@
...
+
+var homeMessage string = `
+d8888b. d888888b  d888b  .d8888. db   dD db    db
+88  '8D   '88'   88' Y8b 88'  YP 88 ,8P' '8b  d8'
+88oooY'    88    88      '8bo.   88,8P    '8bd8'
+88~~~b.    88    88  ooo   'Y8b. 88'8b      88
+88   8D   .88.   88. ~8~ db   8D 88 '88.    88
+Y8888P' Y888888P  Y888P  '8888Y' YP   YD    YP
+
+This is an atproto [https://atproto.com] relay instance, running the 'bigsky' codebase [https://github.com/bluesky-social/indigo]
```
bnewbold, line 332: "`relay` codebase"

```
.########..########.##..........###....##....##
.##.....##.##.......##.........##.##....##..##.
.##.....##.##.......##........##...##....####..
.########..######...##.......##.....##....##...
.##...##...##.......##.......#########....##...
.##....##..##.......##.......##.....##....##...
.##.....##.########.########.##.....##....##...
```

cmd/relay/main.go
```diff
@@ -0,0 +1,484 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"gorm.io/plugin/opentelemetry/tracing"
+)
+
+func init() {
```
bnewbold, line 44: this entire `init()` can just be yanked if using the slog setup function now

cmd/relay/main.go
bnewbold: I would at least update the default from `.../bigsky/...` to `relay` so dev databases don't get clobbered in a weird way

cmd/relay/events/pebblepersist.go
```diff
@@ -0,0 +1,262 @@
+package events
```
bnewbold, line 1: could yank this file if not using pebble persister?

brianolson: result of vendoring events/...

cmd/relay/events/consumer.go has #sync which should go upstream
cmd/relay/events/events.go has #sync which should go upstream

diskpersist.go has real changes; changes in db model; time-ish seq

But yes, I can un-vendor a couple types I didn't use.

brianolson:
force push squashes prior commits and fixes merge conflicts with main

PR #968, opened 2025-03-04
refactor event persist implementations in to sub-packages

bnewbold:
the motivation is to not pull in things like `carstore` in to the dependency tree of every firehose consumer.


brianolson (approved):
LGTM (i was a little worried of merge conflicts with my other work, but this is just moves so it's fine)

PR #973, opened 2025-03-04
remove `./cmd/laputa`

bnewbold:
This was a partial/prototype PDS implementation in golang. Some of that code is still partially used in internal tests, but is very far out of data from the atproto spec.

This removes the command which references that library code. Can always pull this back out of git history if we need it.


bnewbold:
@whyrusleeping said ok

PR #977, opened 2025-03-06
use ipld cbor Blockstore interface

brianolson:
We already used the ipld-cbor Blockstore interface in a few places, and it's a simpler smaller interface that does all we need and isn't deprecated; spread its usage around a bit to simplify some things I was working on elsewhere in relay code


bnewbold (approved):
note that this currently only hits the old MST+repo packages, not the new ones under `indigo:atproto/repo` and `indigo:atproto/repo/mst`

whyrusleeping (approved):

PR #983, opened 2025-03-12
add sync event

bnewbold:
Goal with this is a very small/minimal PR which won't conflict with either Relay v1.1 branch, or https://github.com/bluesky-social/indigo/pull/967

This doesn't deprecate/remove anything, just adds cborgen for existing lexicon types, and callbacks for event handlers.


brianolson (approved):
a bunch of these are the same as what I did on the relay branch, so should merge cleanly with little or no diff

PR #989, opened 2025-03-17
treat sync1.1 errors as warnings

brianolson:


cmd/relay/bgs/validator.go
```diff
@@ -129,36 +132,56 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
 	hasWarning := false
 	commitVerifyStarts.Inc()
-	logger := slog.Default().With("did", msg.Repo, "rev", msg.Rev, "seq", msg.Seq, "time", msg.Time)
+	logger := slog.Default().With("host", hostname, "did", msg.Repo, "rev", msg.Rev, "seq", msg.Seq, "time", msg.Time)
 
 	did, err := syntax.ParseDID(msg.Repo)
 	if err != nil {
 		commitVerifyErrors.WithLabelValues(hostname, "did").Inc()
-		return nil, err
+		if !val.Sync11ErrorsAreWarnings {
```
bnewbold, line 140: I think this one (invalid DID syntax) should probably always be an error (eg, not a sync v1.1 change)

cmd/relay/bgs/validator.go
```diff
@@ -129,36 +132,56 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
+			return nil, err
+		} else {
+			logger.Warn("invalid did", "err", err)
+		}
 	}
 	rev, err := syntax.ParseTID(msg.Rev)
 	if err != nil {
 		commitVerifyErrors.WithLabelValues(hostname, "tid").Inc()
-		return nil, err
+		if !val.Sync11ErrorsAreWarnings {
```
bnewbold, line 149: this one (revision TID syntax) also always an error (not a sync v1.1 change)

cmd/relay/bgs/validator.go
```diff
@@ -305,46 +376,79 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
 func (val *Validator) HandleSync(ctx context.Context, host *models.PDS, msg *atproto.SyncSubscribeRepos_Sync) (newRoot *cid.Cid, err error) {
 	hostname := host.Host
 	hasWarning := false
+	logger := slog.Default().With("host", hostname, "did", msg.Did, "rev", msg.Rev, "seq", msg.Seq, "time", msg.Time)
 
 	did, err := syntax.ParseDID(msg.Did)
 	if err != nil {
 		syncVerifyErrors.WithLabelValues(hostname, "did").Inc()
-		return nil, err
+		if !val.Sync11ErrorsAreWarnings {
```
bnewbold, line 384: I think all these `#sync` event checks could be errors-not-warnings. The whole event is new in sync v1.1, so there shouldn't be any problem with old implementations outputting events which don't validate. It is only the `#commit` events we need to handle with warning-not-error.

bnewbold (approved):
overall looks good, and a couple good little cleanups along the way.

I flagged a couple places we probably don't need warn-versus-error: `#sync` events aren't emitted by pre-sync1.1 software at all, and DID syntax in `#commit` events can always be enforced strictly (not a sync1.1 change)

cmd/relay/bgs/validator.go
brianolson: this is where I thought the discussion went yesterday, that we have to allow mediocre implementation of sync1.1 for a while

cmd/relay/bgs/validator.go
```diff
@@ -305,46 +368,75 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
 		commitVerifyErrors.WithLabelValues(hostname, "car").Inc()
-		return nil, err
+		if !val.Sync11ErrorsAreWarnings {
+			return nil, err
+		} else {
+			logger.Warn("invalid car", "err", err)
+		}
 	}
 
 	if commit.Rev != rev.String() {
```
devinivy, line 419: If there was an error in loading the CAR, I would expect `commit` to be `nil`.  If that's the case, falling through on the "invalid car" warn would cause a nil pointer exception here.

cmd/relay/bgs/validator.go
```diff
@@ -305,46 +368,75 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
+		} else {
+			logger.Warn("message did != commit.did")
+		}
 	}
 
 	err = val.VerifyCommitSignature(ctx, commit, hostname, &hasWarning)
 	if err != nil {
 		// signature errors are metrics counted inside VerifyCommitSignature()
-		return nil, err
+		logger.Warn("invalid sig", "err", err)
```
devinivy, line 439: I believe this now always warns instead of always errors.

cmd/relay/bgs/validator.go
brianolson: fixed (and in commit handler)

cmd/relay/bgs/validator.go
```diff
@@ -409,7 +409,7 @@ func (val *Validator) HandleSync(ctx context.Context, host *models.PDS, msg *atp
 	commit, err := atrepo.LoadCARCommit(ctx, bytes.NewReader([]byte(msg.Blocks)))
 	if err != nil {
 		commitVerifyErrors.WithLabelValues(hostname, "car").Inc()
-		if !val.Sync11ErrorsAreWarnings {
+		if !val.Sync11ErrorsAreWarnings || commit == nil {
```
devinivy, line 412: I would expect that if `err != nil` that the `commit` is going to be `nil`.  Is that not the case?  I think that means we always have to error if we fail to read the commit from the CAR, there's no case where we'll end-up warning here.

brianolson: ok, yeah, content and err are mutually exclusive, cleaned up

bnewbold (commented):
there are a few more instances where a check might error, resulting in nil / zero-like values.

especially for the MST inversion case, I think the easiest fix is to split code out to a sub-method which fails or succeeds all together, and then warn-or-error on that overall result.

cmd/relay/bgs/validator.go
```diff
@@ -305,40 +367,68 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
 	if err != nil {
 		syncVerifyErrors.WithLabelValues(hostname, "tid").Inc()
-		return nil, err
+		if !val.Sync11ErrorsAreWarnings {
+			return nil, err
+		} else {
+			logger.Warn("invalid rev", "err", err)
+		}
 	}
 	if rev.Time().After(time.Now().Add(val.maxRevFuture)) {
```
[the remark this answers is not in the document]
brianolson: in HandleCommit I'd changed it to assume valid DID and rev TID are basic; propagated that here too

cmd/relay/bgs/validator.go
```diff
@@ -305,40 +367,68 @@ func (val *Validator) VerifyCommitMessage(ctx context.Context, host *models.PDS,
...
 	if err != nil {
 		commitVerifyErrors.WithLabelValues(hostname, "car").Inc()
-		return nil, err
+		if val.Sync11ErrorsAreWarnings {
+			logger.Warn("invalid car", "err", err)
+		}
+		return nil, fmt.Errorf("invalid car, %w", err)
 	}
 
 	if commit.Rev != rev.String() {
```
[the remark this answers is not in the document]
brianolson: it's okay because we do always error out on that error

brianolson:
ok, I think I cleaned up all the places where we should bail on a sequence of steps if there's an error, bundled things up into two new inner functions to organize that.

PR #990, opened 2025-03-17
rainbow requestCrawl forwarding overhaul

brianolson:
each forwarding recipient has a processing thread and buffering chan

adds API:
```
POST /v1/requestRequestCrawl body={"url":""}
```

### DIFF
--- a/splitter/splitter.go
+++ b/splitter/splitter.go
@@ -71,6 +71,10 @@ type SplitterConfig struct {
 	PebbleOptions *pebblepersist.PebblePersistOptions
 
 	MaxRequestCrawlForwardErrors int
+
+	AuthTokens []string
+
+	SkipRequestCrawlPing bool
 }
 
 func (sc *SplitterConfig) XrpcRootUrl() string {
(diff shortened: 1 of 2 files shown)


splitter/splitter.go
```diff
@@ -251,6 +276,7 @@ func (s *Splitter) StartWithListener(listen net.Listener) error {
 	e.GET("/xrpc/_health", s.HandleHealthCheck)
 	e.GET("/_health", s.HandleHealthCheck)
 	e.GET("/", s.HandleHomeMessage)
+	e.POST("/v1/requestRequestCrawl", s.RequestRequestCrawlHandler)
```
bnewbold, line 279: feels like we should require auth (admin auth) on this endpoint.

eg, possible security concerns:
- URL isn't sanitized, so could be internal URLs (eg, 10.0.0.1), resulting in SSRF problems
- there are no rate-limits, so a bad actor could create huge numbers of clients, consuming both rainbow resources and request load on target services
- loop detection

bnewbold (commented):
I might have missed some earlier conversation on this setup, but it feels complex to have a dynamic "subscribe to requestCrawl" setup like this, requiring implementation both in rainbow and the consuming service.

Could we just past a list of upstreams to `rainbow`, and have it always push to those? We'd need to reconfigure every time we add a service, but that doesn't seem to bad.

splitter/splitter.go
brianolson: we should definitely have an auth token; I was assuming intra-datacenter usage with a frontend router that would only pipe `subscribeRepos` in from the outside, but auth token is good

brianolson:
OR, this branch but also with etcd client for discovery of requestCrawl forwarding destinations
https://github.com/bluesky-social/indigo/pull/993

bnewbold:
I'm definitely leaning more towards this PR over the etcd one. Though still wish we could just fan-out the requestCrawl to multiple hosts; and not have `rainbow` restarts/deploys be so scary.

PR #997, opened 2025-03-21
nicer log messages & dev relay

brianolson:
dev relay default to everything under ./data/ (sqlite and disk persist)

### DIFF
--- a/cmd/relay/events/schedulers/parallel/parallel.go
+++ b/cmd/relay/events/schedulers/parallel/parallel.go
@@ -24,7 +24,8 @@ type Scheduler struct {
 	lk     sync.Mutex
 	active map[string][]*consumerTask
 
-	ident string
+	identLogKey string
+	ident       string
 
 	// metrics
 	itemsAdded     prometheus.Counter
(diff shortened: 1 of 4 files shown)


cmd/relay/main.go
```diff
@@ -81,7 +81,7 @@ func run(args []string) error {
 		},
 		&cli.StringFlag{
 			Name:    "disk-persister-dir",
-			Usage:   "set directory for disk persister (implicitly enables disk persister)",
+			Usage:   "set directory for disk persister (default: \"./data/event-buffer\")",
```
bnewbold, line 84: maybe simpler to set a `Value:` here on the `cli.StringFlag`?

cmd/relay/events/schedulers/parallel/parallel.go
```diff
@@ -24,7 +24,8 @@ type Scheduler struct {
 	lk     sync.Mutex
 	active map[string][]*consumerTask
 
-	ident string
+	identLogKey string
```
bnewbold, line 27: making this configurable feels pretty convoluted to me. can we just call it "host" or something?

bnewbold (commented):
definitely happy about more log context, and improving the `--disk-persister-dir` argument.

left some comments about implementation details

cmd/relay/events/schedulers/parallel/parallel.go
brianolson: scheduler was presented as being more generic than that, not explicitly about a "host", but it is actually always used that way, so, ok

brianolson: ah, because in relay code I'm trying to move log lines about "host" to be "pds". I think "host" is too generic and if more log lines gather on "pds" it will be easier to search for problem sources in logs. I think some other users of parallel scheduler are using it to process the full relay firehose, so a different log key would make sense there.

PR #998, opened 2025-03-22
use tinyblockstore in repo ingest path

whyrusleeping:


brianolson:
Yup. That's pretty much what it's for.

PR #999, opened 2025-03-22
dont use cursed car v2

whyrusleeping:


brianolson:
```
cmd/gosky/debug.go:36 	"github.com/ipld/go-car/v2"
mst/mst_test.go:19 	"github.com/ipld/go-car/v2"
repo/repo.go:16 	"github.com/ipld/go-car/v2"
```
but the majority of usage is plain go-car. weird and suspicious to have split usage of this library.

PR #1000, opened 2025-03-22
Streaming repo parsing

whyrusleeping:
First pass at a streaming repo parse utility. This iterates the repo in mst order, waiting for the blocks it needs to be read from the passed in stream.

It currently doesnt have a good backpressure mechanism, will need to add that next. I kinda wanted to get this to be a generator pattern type thing instead of using goroutines, but since the blocks can be reordered it gets sticky... we will see

### DIFF
--- a/repo/repo.go
+++ b/repo/repo.go
@@ -1,18 +1,20 @@
 package repo
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"sync"
 
 	"github.com/bluesky-social/indigo/atproto/repo"
 	lexutil "github.com/bluesky-social/indigo/lex/util"
 	"github.com/bluesky-social/indigo/mst"
+	"github.com/bluesky-social/indigo/repo/carutil"
 	"github.com/bluesky-social/indigo/util"
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
-	"github.com/ipld/go-car/v2"
 	cbg "github.com/whyrusleeping/cbor-gen"
 	"go.opentelemetry.io/otel"
 )
@@ -73,17 +75,25 @@ func (uc *UnsignedCommit) BytesForSigning() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+const repoBlockBufferSize = 128 << 10
+
+var repoBlockBufferPool = &sync.Pool{
+	New: func() any {
+		return make([]byte, repoBlockBufferSize)
+	},
+}
+
 func IngestRepo(ctx context.Context, bs cbor.IpldBlockstore, r io.Reader) (cid.Cid, error) {
 	ctx, span := otel.Tracer("repo").Start(ctx, "Ingest")
 	defer span.End()
 
-	br, err := car.NewBlockReader(r)
+	br, root, err := carutil.NewReader(bufio.NewReader(r))
 	if err != nil {
 		return cid.Undef, fmt.Errorf("opening CAR block reader: %w", err)
 	}
 
 	for {
-		blk, err := br.Next()
+		blk, err := br.NextBlock(repoBlockBufferPool, repoBlockBufferSize)
 		if err != nil {
 			if err == io.EOF {
 				break
@@ -96,7 +106,7 @@ func IngestRepo(ctx context.Context, bs cbor.IpldBlockstore, r io.Reader) (cid.C
 		}
 	}
 
-	return br.Roots[0], nil
+	return root, nil
 }
 
 func ReadRepoFromCar(ctx context.Context, r io.Reader) (*Repo, error) {
--- a/repo/stream.go
+++ b/repo/stream.go
@@ -1,0 +1,162 @@
+package repo
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"sync"
+
+	"github.com/bluesky-social/indigo/mst"
+	"github.com/bluesky-social/indigo/repo/carutil"
+	"github.com/bluesky-social/indigo/util"
+	block "github.com/ipfs/go-block-format"
+	cid "github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel"
+)
+
+type waitingBlockstore struct {
+	lk             sync.Mutex
+	blockWaits     map[cid.Cid]chan block.Block
+	otherBlocks    map[cid.Cid]block.Block
+	streamComplete bool
+}
+
+func newWaitingBlockstore() *waitingBlockstore {
+	return &waitingBlockstore{
+		blockWaits:  make(map[cid.Cid]chan block.Block),
+		otherBlocks: make(map[cid.Cid]block.Block),
+	}
+}
+
+func (bs *waitingBlockstore) Get(ctx context.Context, cc cid.Cid) (block.Block, error) {
+	bs.lk.Lock()
+
+	if blk, ok := bs.otherBlocks[cc]; ok {
+		delete(bs.otherBlocks, cc)
+		bs.lk.Unlock()
+		return blk, nil
+	}
+
+	if bs.streamComplete {
+		bs.lk.Unlock()
+		return nil, ErrMissingBlock
+	}
+
+	bw, ok := bs.blockWaits[cc]
+	if ok {
+		bs.lk.Unlock()
+		return nil, fmt.Errorf("somehow already have active wait for block in question: %s", cc)
+	}
+
+	bw = make(chan block.Block, 1)
+
+	bs.blockWaits[cc] = bw
+
+	bs.lk.Unlock()
+
+	select {
+	case blk, ok := <-bw:
+		if !ok {
+			return nil, ErrMissingBlock
+		}
+
+		return blk, nil
+	case <-ctx.Done():
+		return nil, ctx.Err()
+
+	}
+}
+
+var ErrMissingBlock = fmt.Errorf("block was missing from archive")
+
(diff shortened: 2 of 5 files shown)


repo/repo.go
```diff
@@ -97,6 +107,8 @@ func IngestRepo(ctx context.Context, bs cbor.IpldBlockstore, r io.Reader) (cid.C
 		}
 	}
 
+	bufrPool.Put(bufr)
```
ericvolp12, line 110: What happens if the function returns before this line? Do we need to defer putting the buffer back in the pool?

whyrusleeping: yeah deferring the put probably makes sense, in general its not a huge deal if we miss putting something back in the pool in an error case, but this isnt inconvenient

backfill/backfill.go
```diff
@@ -323,8 +325,8 @@ func (e *FetchRepoError) Error() string {
 	return fmt.Sprintf("failed to get repo: %s (%d)", reason, e.StatusCode)
 }
 
-// Fetches a repo CAR file over HTTP from the indicated host. If successful, parses the CAR and returns repo.Repo
-func (b *Backfiller) fetchRepo(ctx context.Context, did, since, host string) (*repo.Repo, error) {
+// Fetches a repo CAR file over HTTP from the indicated host.
+func (b *Backfiller) fetchRepo(ctx context.Context, did, since, host string) (io.ReadCloser, error) {
```
bnewbold, line 329: I like this API change :heavy_plus_sign:.

Could add to the doc comment that calling code is responsible for closing the returned reader (though it is a non-exported method).

repo/stream.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/bluesky-social/indigo/mst"
+	"github.com/bluesky-social/indigo/repo/carutil"
+	"github.com/bluesky-social/indigo/util"
+	block "github.com/ipfs/go-block-format"
+	cid "github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel"
+)
+
+const bufPoolBlockSize = 1024
```
bnewbold, line 19: is this value justified, or just a guess? probably fine as a starting point, though might be worth adding logging, loading up a CAR file, and see what the distribution looks like. I would guess a lot of records are a bit over 1kbyte and 2k or 4k might reduce allocation churn. but maybe memory efficiency is more important for RPi bulk streaming (for example)

repo/stream.go
```diff
@@ -0,0 +1,162 @@
...
+	blk, err := bs.readUntilBlock(ctx, cc)
+	if err != nil {
+		return nil, err
+	}
+
+	return blk, nil
+}
+
+func (bs *readStreamBlockstore) View(cc cid.Cid, cb func([]byte) error) error {
+	blk, err := bs.get(context.TODO(), cc)
```
bnewbold, line 92: seems worth wiring through the `ctx`. the only call point I can see has the `ctx` right there

repo/stream.go
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+var ErrMissingBlock = fmt.Errorf("block was missing from archive")
+
+func (bs *readStreamBlockstore) Put(ctx context.Context, blk block.Block) error {
+	return fmt.Errorf("put is not needed")
+}
+
+func StreamRepoRecords(ctx context.Context, r io.Reader, prefix string, setRev func(string), cb func(k string, c cid.Cid, v []byte) error) error {
```
bnewbold: The `setRev` could use a doc string here, and worth saying that `cb` is the visitor function. And which can be `nil`.

Should declare if `prefix` can be an exact match (eg, if the prefix is the full path, does that record get returned).

I think there should also be a description of error handling: if `cb` returns an error, does walking stop immediately, or does the first/last error get returned for the overall `StreamRepoRecords`?

repo/stream.go
bnewbold: I'm also not familiar with functions with this name and signature, probably worth a quick comment. "calls provided callback with the value corresponding to the given CID".

whyrusleeping: unfortunately the View interface comes from somewhere else and changing that would be a PITA

repo/stream.go
whyrusleeping: Its a bit of a guess, but I do think most records are under 1k, likes, follows, most posts

uniphil: oh i have this one handy:

size buckets

buckets are "record block sizes less than"

counts are all records across morel pds (nov snapshot)

sum of records over 1KB: 3.8%
over 2K: 0.4%
over 4K: 0.02%

uniphil: wait i did the numbers shifted at the end. 1kb bar is <1kb.

sum of records over 1KB: 0.4%
over 2K: 0.02%
over 4K: 0.001%

repo/stream.go
```diff
@@ -0,0 +1,179 @@
...
+	var sc SignedCommit
+	if err := cst.Get(ctx, root, &sc); err != nil {
+		return fmt.Errorf("loading root (%s) from blockstore (other blocks: %d): %w", root, len(bs.otherBlocks), err)
+	}
+
+	if sc.Version != ATP_REPO_VERSION && sc.Version != ATP_REPO_VERSION_2 {
+		return fmt.Errorf("unsupported repo version: %d", sc.Version)
+	}
+
+	// TODO: verify that signature
```
bnewbold: it would be good to have a way to handle commit signature validation in the future.

what if `setRev` was a commit callback, passed the entire SignedCommit, and could return an error? the existing code can easily pluck off just the `rev` field. in the future code could resolve the identity and verify the signature; if it doesn't match an error would get piped through.

repo/stream.go
bnewbold: great, 1 KB seems fine then (512 might also work)

repo/repo.go
```diff
@@ -76,17 +78,28 @@ func (uc *UnsignedCommit) BytesForSigning() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+var bufrPool = &sync.Pool{
+	New: func() any {
+		return bufio.NewReader(nil)
+	},
+}
+
 func IngestRepo(ctx context.Context, bs cbor.IpldBlockstore, r io.Reader) (cid.Cid, error) {
```
bnewbold, line 87: I checked and this function does have at least one happy-path test case, loading a CAR file from disk in `testing/sig_test.go`. that exercises the new carutil code a bit.

events/consumer.go
```diff
@@ -162,6 +164,13 @@ func HandleRepoStream(ctx context.Context, con *websocket.Conn, sched Scheduler,
 		return nil
 	})
 
+	cr := new(cbg.CborReader)
```
bnewbold, line 167: can you talk a bit about this change? not obvious why `CborReader` needs to get added here (or why it wasn't previously). does the UnmarchalCBOR below check the reader type and wrap it if necessary?

also not immediately obvious what the ramification of reusing the instrumented reader between loop iterations is... more accurate metrics? no change? will it impact rate-limiting or back-pressure behaviors?

backfill/gormstore.go
```diff
@@ -180,7 +180,10 @@ func (j *Gormjob) BufferOps(ctx context.Context, since *string, rev string, ops
 	switch j.state {
 	case StateComplete:
 		return false, nil
-	case StateInProgress, StateEnqueued:
+	case StateEnqueued:
+		// if the repo is enqueue, but not actively being backfilled, just ignore events for it for now
```
bnewbold, line 184: this sounds fine (save fewer unnecessary events in database), but i'm not sure how the concurrency works out in vortex.

backfill/backfill.go
```diff
@@ -426,36 +422,38 @@ func (b *Backfiller) BackfillRepo(ctx context.Context, job Job) (string, error)
...
+	setRev := func(s string) {
+		rev = s
+	}
+
 	// Producer routine
+	var streamRecordsError error
 	go func() {
+		defer r.Close()
 		defer close(recordQueue)
-		if err := r.ForEach(ctx, b.NSIDFilter, func(recordPath string, nodeCid cid.Cid) error {
```
bnewbold, line 432: thinking out loud: there is a new possible error condition where the CAR file and repo starts being processed before it is fully downloaded. that means the network connection being interrupted can result in partial processing (some records getting processed but not all).

that was probably possible in theory before (if something interrupted record processing after the MST had already been fully loaded), but maybe becomes more frequent?

but I guess the backfiller would track that condition and come back around later and re-attempt processing.

whyrusleeping:
> I think the service most at risk of unseen/unexpected disruption would be vortex.

Yeah, for a new deployment of vortex it might hit this, but in steady state it shouldn't IIUC

backfill/gormstore.go
whyrusleeping: This just saves memory, we were previously queueing up all the firehose events for the repo, but when we finally got around to fetching the repo, it already included those

backfill/backfill.go
whyrusleeping: Yeah, that would result in the repo backfill being marked as an error and it would get retried later, and the duplicate records from the second fetch would just get ignored

events/consumer.go
whyrusleeping: This one is entirely about memory allocations, the cborgen code can look through the interface to see if its a CborReader, and wraps it if it doesnt find what its looking for. This just prevents that extra allocation. 

Reusing the instrumentedReader is fine as all its doing is tracking things the same way every time. I could make another one each loop but its identical operation (the thing outside of its reader is stateless)

repo/stream.go
whyrusleeping: Yeah, for now i'm happy to skip on this but agree would be a nice to have.

PR #1005, opened 2025-03-29
refactor api/*.go in to PLC and handle packages

bnewbold:
This is older code for working with PLC, and resolving handles.

This doesn't touch the code itself, just moves out of `api/*.go` to other packages, and then marks those packages as deprecated.

Having this in `api/*.go` was weird because everything else under `api/` is codegen (lexgen).


ericvolp12 (approved):
Sure

PR #1010, opened 2025-04-04
relay refactor

bnewbold:
Still a bunch to do here; this isn't ready for review or experimentation, just showing progress.


bnewbold:
I merged some patches which came up along the way to main, and did a rebase of this branch.

cmd/relay/relay/models/models.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"time"
+
+	"gorm.io/gorm"
+)
+
+type DomainBan struct {
+	gorm.Model
+	Domain string `gorm:"unique"`
+}
```
devinivy, line 12: Does gorm automatically add a primary key, or could we potentially make `domain` a pkey?

cmd/relay/relay/models/models.go
```diff
@@ -0,0 +1,103 @@
...
+type AccountRepo struct {
+	// references Account.UID, but not set up as a foreign key
+	UID uint64 `gorm:"column:uid;primarykey"`
+	Rev string `gorm:"column:rev;not null"`
+
+	// The CID of the entire signed commit block. Sometimes called the "head"
+	CommitCID string `gorm:"column:commit_cid;not null"`
+
+	// The CID of the top of the repo MST, which is the 'data' field within the commit block. This becomes 'prevData'
+	CommitData string `gorm:"column:commit_data;not null"`
```
devinivy, line 98: I have a tiny brain itch w/ `commit_cid` vs. `commit_data` both being CIDs but only reflected in one of their names.  Not a biggie though.

bnewbold: Agree! `CommitDataCID`/`commit_data_cid` work for you?

cmd/relay/relay/models/models.go
bnewbold: It does when the struct extends `gorm.Model`: that adds `ID`, `CreatedAt`, `UpdatedAt`, `DeletedAt`: https://gorm.io/docs/models.html#gorm-Model

I didn't really update this model (it is simple), but think it should be more explicit like the other models, and easy to do this update.

PR #1013, opened 2025-04-06
go.mod: specify full version (1.23.0)

bnewbold:
For the automatic go toolchain stuff to work, we need the full version, including patch (not just minor)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/bluesky-social/indigo
 
-go 1.23.0
+go 1.23
+toolchain go1.23.8
 
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/bluesky-social/indigo
 
-go 1.23
+go 1.23.0
 
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2


jcalabro (approved):
I try and avoid using `*.0` releases - usually like to yank in the bug releases that come through on patches. Though since this is partially at least consumed as a library, are we intentionally trying to target oldest still-supported versions?

Perhaps 1.23.8 would be better?

bnewbold:
I'm trying to remember why I actually wanted this change; I had a TODO note from weeks ago.

Looking over the docs:

- https://go.dev/ref/mod#go-mod-file-go
- https://go.dev/doc/toolchain

I think maybe what I actually wanted was a `toolchain` line; and needed full version (not just patch) in that context (eg, `go1.23.0` works but `go1.23` does not). What I vaguely remember was having an Ubuntu 22.04 (or 24.04?) server with go 1.22, and couldn't compile something from this repo. With the recent go toolchain stuff, I was able to get it working by setting a `toolchain` line so it would download an updated toolchain, but it needed the patch version specified?

So maybe this PR should be:

```
go 1.23
toolchain go1.23.8
```

This would keep the "minimum golang version" low (to maximize reuse/compatibility when folks are using `indigo` as an SDK in other projects), but ensure a more recent toolchain is used when we do builds for images/executables and our own prod services.

We do tend to wait on actual upgrades until there is a bugfix patch version of the new go toolchain.

(this PR in particular isn't urgent for me right now, but I am in favor of having our versioning situation clean both for us and downstream projects)

bnewbold:
Ok, hit this again today and think I can reproduce what happened:

```
ubuntu@host:~$ go version
go version go1.22.2 linux/amd64

ubuntu@host:~$ head -n1 /etc/os-release 
PRETTY_NAME="Ubuntu 24.04.2 LTS"

ubuntu@host:~/indigo$ go build ./cmd/relay/
go: downloading go1.23 (linux/amd64)
go: download go1.23 for linux/amd64: toolchain not available
```

If I set:

```
go 1.23
toolchain go1.23.8
```

then it works as expected: downloads newer toolchain and builds.

jcalabro (approved):
Cool! I actually didn't know this; I've never knowingly used the auto-toolchain stuff before (I pin my versions and typically manage it all myself, but I also don't work across many Go codebases at once, so this is a neat feature for those that do!)

Issue #1007, opened 2025-03-29
proposed refactor of `events/`


bnewbold:
make `XRPCStreamEvent` just about repo stream (firehose) messages; remove unused `Labels*` stuff

refactor `events` package:

- `stream/`: interfaces and functions: EventPersistence, Scheduler, HandleRepoStream, RepoStreamCallbacks, Scheduler (and possibly re-export XRPCStreamEvent?)
- `stream/events/`: just serialization structs and basic methods off them: XRPCStreamEvent
- `stream/persist/*`: implementations of EventPersistence
- `stream/schedulers/*`: implementations of Scheduler
- `stream/eventmgr/`: EventManager

move `bgs/` to `cmd/bigsky/libbgs/`

ericvolp12:
Don't we use the `labels` stuff in Vortex for label ingestion? Will DM you the usage.

bnewbold:
ack on label ingest usage, I hadn't checked that codebase. fine not changing `XRPCStreamEvent` for now.

the core thing with the `events/` directory is that a whole lot of code depends on the `XRPCStreamEvent` struct for different use-cases. that is fine/good, but I think should be untangled from `EventManager`.

PR #1018, opened 2025-04-07
refactor EventManager in to a sub-package

bnewbold:
This is just moving around code, updating imports, and removing some unused types. There should be no changes in behavior.

This is part of https://github.com/bluesky-social/indigo/issues/1007

We'd need to update some code in another repo to match this; but it is also just import updates there, not that big a deal.

On the other hand, this isn't hugely important. I did this refactor as part of the new relay project, and might just keep a vendored copy of the code there if this doesn't get merged.


bnewbold:
This got out of date, going to close and can try again another time.

PR #1021, opened 2025-04-08
search: handle `#identity` events, instead of `#handle`

bnewbold:
`#handle` events have been deprecated for a while, have been removed from the upstream lexicon schemas (eg, next `lexgen` will remove from indigo), and won't be passed through by new relay.

I'm not really sure how palomar is even working today? But this updates to use the `#identity` event for handle updates instead.


bnewbold:
this actually should not have the additional directory lookup in there

PR #1023, opened 2025-04-10
rainbow: small refactors, and improve request forwarding

bnewbold:
A bit of this is moving code around so I could understand which pieces were doing what. Then improved how the request proxying and requestCrawl forwarding stuff works, as part of new relay ops.

None of the core configuration or behaviors should change: I tweaked some argument names, but not the env var names, so nothing about how we deploy/operate rainbow should change.

No behavior of the core code (ringbuf, upstream websocket subscription, or serving to clients) has changed.

### DIFF
--- a/cmd/rainbow/main.go
+++ b/cmd/rainbow/main.go
@@ -52,7 +52,7 @@ func run(args []string) error {
 		&cli.StringFlag{
 			Name:    "upstream-host",
 			Value:   "http://localhost:2470",
-			Usage:   "simple hostname (no URI scheme) of the upstream host (eg, relay)",
+			Usage:   "URL (schema and hostname, no path) of the upstream host (eg, relay)",
 			EnvVars: []string{"ATP_RELAY_HOST", "RAINBOW_RELAY_HOST"},
 		},
 		&cli.StringFlag{
@@ -90,9 +90,16 @@ func run(args []string) error {
 			EnvVars: []string{"RAINBOW_PERSIST_BYTES", "SPLITTER_PERSIST_BYTES"},
 		},
 		&cli.StringSliceFlag{
+			// TODO: better name for this argument
 			Name:    "next-crawler",
-			Usage:   "forward POST requestCrawl to this url, should be machine root url and not xrpc/requestCrawl, comma separated list",
+			Usage:   "forward POST requestCrawl to these hosts (schema and host, no path) in addition to upstream-host. Comma-separated or multiple flags",
 			EnvVars: []string{"RAINBOW_NEXT_CRAWLER", "RELAY_NEXT_CRAWLER"},
+		},
+		&cli.StringFlag{
+			Name:    "collectiondir-host",
+			Value:   "http://localhost:2510",
+			Usage:   "host (schema and hostname, no path) of upstream collectiondir instance, for com.atproto.sync.listReposByCollection",
+			EnvVars: []string{"RAINBOW_COLLECTIONDIR_HOST"},
 		},
 		&cli.StringFlag{
 			Name:    "env",
@@ -101,8 +108,9 @@ func run(args []string) error {
 			EnvVars: []string{"ENVIRONMENT"},
 		},
 		&cli.BoolFlag{
-			Name:  "enable-otel-otlp",
-			Usage: "enables OTEL OTLP exporter endpoint",
+			Name:    "enable-otel-otlp",
+			Usage:   "enables OTEL OTLP exporter endpoint",
+			EnvVars: []string{"RAINBOW_ENABLE_OTEL_OTLP", "ENABLE_OTEL_OTLP"},
 		},
 		&cli.StringFlag{
 			Name:    "otel-otlp-endpoint",
@@ -160,6 +168,7 @@ func runSplitter(cctx *cli.Context) error {
 
 	persistPath := cctx.String("persist-db")
 	upstreamHost := cctx.String("upstream-host")
+	collectionDirHost := cctx.String("collectiondir-host")
 	nextCrawlers := cctx.StringSlice("next-crawler")
 
 	var spl *splitter.Splitter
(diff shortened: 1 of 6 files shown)


cmd/rainbow/main.go
```diff
@@ -91,69 +92,80 @@ func run(args []string) {
...
+		},
+		&cli.BoolFlag{
+			Name:  "enable-otel-otlp",
+			Usage: "enables OTEL OTLP exporter endpoint",
+		},
+		&cli.StringFlag{
+			Name:    "otel-otlp-endpoint",
+			Usage:   "OTEL traces export endpoint",
+			Value:   "http://localhost:4318",
+			EnvVars: []string{"OTEL_EXPORTER_OTLP_ENDPOINT"},
```
ericvolp12, line 111: Should this just have no default value and then disable itself when not supplied? I'm not sure having a default here is useful to us since we generally don't run otel collectors on localhost.

cmd/rainbow/main.go
```diff
@@ -2,56 +2,57 @@ package main
...
-			Name:    "splitter-host",
-			Value:   "bsky.network",
+			Name:    "log-level",
+			Usage:   "log verbosity level (eg: warn, info, debug)",
+			EnvVars: []string{"RAINBOW_LOG_LEVEL", "GO_LOG_LEVEL", "LOG_LEVEL"},
+		},
+		&cli.StringFlag{
+			Name:    "upstream-host",
+			Value:   "http://localhost:2470",
+			Usage:   "simple hostname (no URI scheme) of the upstream host (eg, relay)",
```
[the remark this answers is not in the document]
bnewbold: good catch! updated the code but hadn't updated the comment

cmd/rainbow/main.go
bnewbold: I added the separate `enable-otel-otlp` flag to enable/disable.

Previously the whole feature was gated by whether this config was enabled at all. The config was buried in code, and not obvious what kind of string to configure. This way running `./rainbow serve --help` displays all the env vars, and comes with an example value, which makes it more accessible. (at least is the theory)

bnewbold:
Updated to address review comment; cleaned up hostname wrangling a bit; and added collectiondir proxying.

splitter/firehose.go
```diff
@@ -0,0 +1,198 @@
...
+	var since *int64
+	if sinceVal := c.QueryParam("cursor"); sinceVal != "" {
+		sval, err := strconv.ParseInt(sinceVal, 10, 64)
+		if err != nil {
+			return err
+		}
+		since = &sval
+	}
+
+	ctx, cancel := context.WithCancel(c.Request().Context())
```
jcalabro, line 29: This might be something to look in to because I haven't use echo in a few years, but does `c.Request().Context()` outlive the HTTP 101 response? I'm guessing the answer is no, but I could easily be wrong.

I would probably suggest replacing it with `context.WithCancel(context.Background())`

jcalabro: Just did a quick experiment and it looks like the `c.Request().Context()` is fine to use. It also aligns with the [docs on http.Hijacker](https://pkg.go.dev/net/http#Hijacker), which gorilla/websocket calls on Upgrade:

```
	// After a call to Hijack, the original Request.Body must not
	// be used. The original Request's Context remains valid and
	// is not canceled until the Request's ServeHTTP method
	// returns.
```

jcalabro: Screenshot 2025-04-15 at 10 27 56

Full code snippet:

```go
package main

import (
        "context"
        "log/slog"
        "time"

        "github.com/gorilla/websocket"
        "github.com/labstack/echo/v4"
)

func handleWebSocket(c echo.Context) error {
        ctx, cancel := context.WithCancel(c.Request().Context())
        defer cancel()

        upgrader := websocket.Upgrader{}
        conn, err := upgrader.Upgrade(c.Response(), c.Request(), nil)
        if err != nil {
                return err
        }
        defer conn.Close()

        _, msg, err := conn.ReadMessage()
        if err != nil {
                panic(err)
        }
        slog.Info("message", "msg", string(msg))

        time.Sleep(time.Second)

        select {
        case <-ctx.Done():
                slog.Info("websocket context canceled")
                return nil
        default:
                slog.Info("ctx not yet canceled")
        }

        err = conn.WriteMessage(websocket.TextMessage, []byte("hello world!\n"))
        if err != nil {
                panic(err)
        }
        slog.Info("sent message")

        return nil
}

func main() {
        e := echo.New()
        e.GET("/ws", handleWebSocket)
        e.Start(":8080")
}
```

splitter/handlers.go
```diff
@@ -0,0 +1,138 @@
...
+		return c.JSON(http.StatusInternalServerError, xrpc.XRPCError{ErrStr: "ProxyRequestFailed", Message: fmt.Sprintf("failed forwarding request: %s", err)})
+	}
+
+	// if that was successful, then forward on to the other upstreams (in goroutines)
+	for _, c := range s.nextCrawlers {
+		go func() {
+			ctx := context.Background()
+			xrpcc := xrpc.Client{
+				Client: s.upstreamClient,
+				Host:   c.String(),
```
jcalabro, line 73: This is a synchronization bug. You need to alias `c` to a variable that's local to this goroutine.

I'd recommend running with `-race` while developing every now and then. Compile times and process shutdown take longer, but it has saved my butt more times than I can count!

splitter/firehose.go
bnewbold: I'll add a comment and link back to this investigation

splitter/handlers.go
bnewbold: good catch!

I can't get the race detector to trigger on this though (I hadn't been running it previously, but tried just now). I ran with a local relay, and configured a bogus "next crawl" server, and ran both:

```
go build -race ./cmd/rainbow/
./rainbow --next-crawler https://pds.example.com

go run -race ./cmd/rainbow/ --next-crawler https://pds.example.com
```

(I have had the race detector trigger on relay code in the past)

jcalabro: This is the thing that was fixed via https://go.dev/blog/loopvar-preview, so as of 1.22, we're good actually!

bnewbold:
@jcalabro is this overall good? I am fine leaving in the fix to the loopvar issue in this PR, even if it is fixed in modern go.

jcalabro (approved):
I'm not 100% sure I have enough context to give meaningful review just yet, but from my understanding of what rainbow's semantics should be nothing jumps out at me. Additionally, code-wise this looks good. Don't want to hold up the train, so approving!

One thing I can think of that would make me feel better is having tests for the `splitter` package.

PR #1024, opened 2025-04-11
some collectiondir fixes

bnewbold:
This isn't a full pass, just fixing some known issues:

- process wasn't shutting down with Ctrl-C because of a WaitGroup. fixed the bug, but didn't clean up the (many) channels+goroutines
- many API response errors were strings, not the expected XRPC JSON message format
- README didn't describe the actual API endpoint

### DIFF
--- a/cmd/collectiondir/serve.go
+++ b/cmd/collectiondir/serve.go
@@ -297,7 +297,6 @@ func (cs *collectionServer) Shutdown() error {
 	}()
 	cs.log.Info("api shutdown start...")
 	err := cs.apiServer.Shutdown(context.Background())
-	//err := cs.esrv.Shutdown(context.Background())
 	cs.log.Info("api shutdown, thread wait...", "err", err)
 	cs.wg.Wait()
 	cs.log.Info("threads done, db close...")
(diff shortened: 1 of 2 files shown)


cmd/collectiondir/serve.go
```diff
@@ -705,6 +704,9 @@ func (cs *collectionServer) ingestReceiver() {
 			} else {
 				errcount = 0
 			}
+		case <-cs.shutdown:
```
jcalabro, line 707: I'm not sure I have a strong preference for either of our approaches to solve this ([here's](https://github.com/bluesky-social/indigo/pull/1029/commits/8c00259a165833ad0e8e8870691b5f55e3725ea8) what I did). Way too many goroutines flying around for managing the server listeners and there's just a lot of mental load needed to read this file from top to bottom.

Happy to take either solution, but I think if we do take mine, we should change `StartApiServer` and `StartMetricsServer` to fatally shut down on any error rather than just return errors which will now be silently eaten. I like that mine simplifies the # of goroutines needed though (made that change [here](https://github.com/bluesky-social/indigo/pull/1029/commits/444a6aa911744316ac491b404a3a793311bfddfe)).

jcalabro (approved):
README and XRPC responses look good, so approving and we can merge this one. I'll close my own [XRPC response PR](https://github.com/bluesky-social/indigo/pull/1030) (I didn't see that this PR existed before creating it!) and I'll clean up the race conditions out of band once this is merged.

cmd/collectiondir/serve.go
bnewbold: I'm in favor of your solution, and let's come back to it after this PR is merged.

PR #1028, opened 2025-04-13
switch to atproto/syntax TID generator

notjuliet:
repo and repomgr packages rely on incorrect TID generation functions which do not enforce lengths properly unlike the ones in the syntax package, producing strings of 12 characters instead of 13, due to not padding the clock id, breaking lexicon validation when used in the context of atproto. This removes the `tid.go` file to use the other implementation instead.


ericvolp12 (approved):

bnewbold (approved):
thanks!

PR #1029, opened 2025-04-14
Collectiondir Improvements

jcalabro:
Perhaps best reviewed by looking at each individual commit.

All of these were found by running this, then sending a ctrl+c:

```bash
go run -race ./cmd/collectiondir serve --pebble ~/local/data/collectiondir --dau-directory ~/local/data/collectiondir
```

This PR improves a few small but important things about the collectiondir service:

1. This service does not have pprof on the metrics router (f16258882a1c6a82c25465547ae824afbeec9486)
    a. `/metrics` and `/debug/pprof/` now live side-by-side
    b. All services should have pprof enabled on private metrics routers
2. Fixes a race condition on startup where we're binding a bunch of fields on `cs` after launching a goroutine that uses some of those fields (82656e88f5ed5ff706c00cfabb98701c3e8e6d0a)
3. Fixes an issue where sending a SIGTERM to the service causes it to start shutting down, but the service never actually dies because there is an infinite hang due to an errant `cs.wg.Wait()`
    a. In general though, the `errchan` is not a good approach. Prefer to let each background goroutine do its logging and only coordinate on shutdown timing through a single variable (`cs.wg`)
4. Fix a race condition on the API server startup/shutdown (5191ddd755928d5763cee92d7510c5999583d006)
5. Fix a race condition on the metrics server startup/shutdown (c90e279b78482bea54451cbf6d0520faa432fa8d)

### DIFF
--- a/cmd/collectiondir/serve.go
+++ b/cmd/collectiondir/serve.go
@@ -217,9 +217,12 @@ func (cs *collectionServer) run(cctx *cli.Context) error {
 	}
 	cs.statsCacheFresh.L = &cs.statsCacheLock
 
-	apiAddr := cctx.String("api-listen")
+	apiServerEcho, err := cs.createApiServer(cctx.Context, cctx.String("api-listen"))
+	if err != nil {
+		return err
+	}
 	cs.wg.Add(1)
-	go func() { cs.StartApiServer(cctx.Context, apiAddr) }()
+	go func() { cs.StartApiServer(cctx.Context, apiServerEcho) }()
 
 	metricsAddr := cctx.String("metrics-listen")
 	cs.wg.Add(1)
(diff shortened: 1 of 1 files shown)


cmd/collectiondir/serve.go
```diff
@@ -385,23 +379,29 @@ func (cs *collectionServer) handleCommit(commit *comatproto.SyncSubscribeRepos_C
 	}
 }
 
-func (cs *collectionServer) StartMetricsServer(ctx context.Context, addr string) error {
-	defer cs.wg.Done()
-	defer cs.log.Info("metrics server exit")
+func (cs *collectionServer) createMetricsServer(addr string) {
+	e := echo.New()
+	e.GET("/metrics", echo.WrapHandler(promhttp.Handler()))
+	e.Any("/debug/pprof/*", echo.WrapHandler(http.DefaultServeMux))
```
bnewbold, line 385: oh, interesting! if this works for pprof we should probably do it everywhere

jcalabro: Yup! This is the pattern I typically follow. Running collectiondir locally, this gives:

```
$ xh localhost:2511/metrics | head -3
# HELP collectiondir_firehose_commits number of #commit events received from upstream firehose
# TYPE collectiondir_firehose_commits counter
collectiondir_firehose_commits 0

$ xh localhost:2511/debug/pprof/goroutine?debug=1 | head -3
goroutine profile: total 82
64 @ 0x4800ee 0x415885 0x415492 0x123785e 0x1054df8 0x1237752 0x488121
# labels: {"pebble":"table-cache"}
```

jcalabro:
Yo @bnewbold, I merged your changes in from https://github.com/bluesky-social/indigo/pull/1024 and I think this is ready for review again. Pretty low priority now that the SIGTERM is handled correctly though (IMO the biggest value add here is pprof)

bnewbold (approved):
nice, thanks for these cleanups!

PR #1030, opened 2025-04-14
Gracefully Handle XRPC Errors in Collectiondir

jcalabro:
Similar to what we do with [other services](https://github.com/bluesky-social/indigo/blob/8500c47d717c0a6004191d5246b37d322210e72b/cmd/bluepages/server.go#L205) that speak XRPC, collectiondir should respond with the standard error response format according to [the docs](https://atproto.com/specs/xrpc).

I'm not 100% that we want this same format for the "admin" handlers on this service. Happy to apply this to all handlers if that's the case, but for now, I've only modified the actual XRPC request (there's only one on the collectiondir service at this time).

Here's an example failed response, and a successful response with these new error handlers:

```
$ xh -v localhost:2510/xrpc/com.atproto.sync.listReposByCollection?collection=bad
GET /xrpc/com.atproto.sync.listReposByCollection?collection=bad HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Host: localhost:2510
User-Agent: xh/0.24.0

HTTP/1.1 400 Bad Request
Content-Length: 84
Content-Type: application/json; charset=UTF-8
Date: Mon, 14 Apr 2025 15:51:04 GMT
Vary: Origin

{
    "error": "InvalidCollectionNSID",
    "message": "NSID syntax didn't validate via regex"
}
```

```
$ xh -v localhost:2510/xrpc/com.atproto.sync.listReposByCollection?collection=a.b.c
GET /xrpc/com.atproto.sync.listReposByCollection?collection=a.b.c HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Host: localhost:2510
User-Agent: xh/0.24.0

HTTP/1.1 200 OK
Content-Length: 13
Content-Type: application/json; charset=UTF-8
Date: Mon, 14 Apr 2025 15:54:52 GMT
Vary: Origin

{
    "repos": []
}
```

### DIFF
--- a/cmd/collectiondir/serve.go
+++ b/cmd/collectiondir/serve.go
@@ -405,6 +405,7 @@ func (cs *collectionServer) StartApiServer(ctx context.Context, addr string) err
 	}
 	e := echo.New()
 	e.HideBanner = true
+	e.HTTPErrorHandler = cs.errorHandler
 
 	e.Use(svcutil.MetricsMiddleware)
 	e.Use(middleware.CORSWithConfig(middleware.CORSConfig{
@@ -455,6 +456,26 @@ func getLimit(c echo.Context, min, defaultLim, max int) int {
 	return lv
 }
 
+type xrpcError struct {
+	Error   string `json:"error"`
+	Message string `json:"message"`
+}
+
+func (srv *collectionServer) errorHandler(err error, c echo.Context) {
+	code := http.StatusInternalServerError
+	var errorMessage string
+	if he, ok := err.(*echo.HTTPError); ok {
+		code = he.Code
+		errorMessage = fmt.Sprintf("%s", he.Message)
+	}
+	if code >= 500 && code <= 599 {
+		slog.With("err", err).Error("collectiondir-http-internal-error")
+	}
+	if !c.Response().Committed {
+		c.JSON(code, xrpcError{Error: "InternalError", Message: errorMessage})
+	}
+}
+
 // /xrpc/com.atproto.sync.listReposByCollection?collection={}&cursor={}&limit={50<=N<=1000}
 // /v1/getDidsForCollection?collection={}&cursor={}&limit={50<=N<=1000}
 //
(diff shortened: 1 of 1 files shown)


cmd/collectiondir/serve.go
```diff
@@ -455,6 +456,26 @@ func getLimit(c echo.Context, min, defaultLim, max int) int {
...
+
+func (srv *collectionServer) errorHandler(err error, c echo.Context) {
+	code := http.StatusInternalServerError
+	var errorMessage string
+	if he, ok := err.(*echo.HTTPError); ok {
+		code = he.Code
+		errorMessage = fmt.Sprintf("%s", he.Message)
+	}
+	if code >= 500 && code <= 599 {
+		slog.With("err", err).Error("collectiondir-http-internal-error")
```
bnewbold, line 472: Is there an advantage to the `With()` syntax, versus:

```
slog.Error("collectiondir-http-internal-error", "err", err)
```

I don't think it matters much either way, just curious

cmd/collectiondir/serve.go
```diff
@@ -455,6 +456,26 @@ func getLimit(c echo.Context, min, defaultLim, max int) int {
 	return lv
 }
 
+type xrpcError struct {
```
bnewbold, line 459: There is an existing type `xrpc.XRPCError` in `indigo:xrpc`. It is such a simple type that i'm fine re-declaring it though!

cmd/collectiondir/serve.go
```diff
@@ -465,14 +486,20 @@ func (cs *collectionServer) getDidsForCollection(c echo.Context) error {
 	collection := c.QueryParam("collection")
 	_, err := syntax.ParseNSID(collection)
 	if err != nil {
-		return c.String(http.StatusBadRequest, fmt.Sprintf("bad collection nsid, %s", err.Error()))
+		return c.JSON(http.StatusBadRequest, xrpcError{
+			Error:   "InvalidCollectionNSID",
+			Message: err.Error(),
```
bnewbold, line 491: FWIW, I often forget to use `err.Error()` and do `fmt.Sprintf("%s", err)`; please do update/correct me if I forget.

cmd/collectiondir/serve.go
```diff
@@ -465,14 +486,20 @@ func (cs *collectionServer) getDidsForCollection(c echo.Context) error {
 	collection := c.QueryParam("collection")
 	_, err := syntax.ParseNSID(collection)
 	if err != nil {
-		return c.String(http.StatusBadRequest, fmt.Sprintf("bad collection nsid, %s", err.Error()))
+		return c.JSON(http.StatusBadRequest, xrpcError{
+			Error:   "InvalidCollectionNSID",
```
bnewbold, line 490: We could declare more error types in either the endpoint-specific lexicon schemas, or as protocol-level "standard errors". But that is a bigger protocol-wide refactor, and i'm :heavy_plus_sign: to declaring new error names on the fly in situations like this (and I have done so myself in a few places).

bnewbold (approved):

bnewbold:
This looks good, and I think is the good pattern for doing error responses for now. Having the overall error handler means that accidentally doing a `return err` results in proper JSON response.

A pattern we use in a few projects is separate "stubs" (which implement the `http` handler function signature, as opposed to `echo`) to do body / query-param parsing, and "handlers" which implement the actual application logic. That comes from having codegen (`lexgen`) output the stubs (though IIRC this is not functioning). A tension I've had is whether the "stubs" should catch and format the errors (as JSON), or have the "handler" code return a specific status code and error name. Basically it feels like the boundary between HTTP-error-response and regular-golang-error-types isn't always super clear, but doing it in "handlers" is probably good. And having an HTTP error handler like this helps catch corner-cases and mistakes.

Admin routes: agree that the actual XRPC endpoints are the priority. For this specific service, I think moving any endpoints to being lexicon schemas (or just removing/deprecating them) might be the future path.

cmd/collectiondir/serve.go
jcalabro: Ah didn't realize I was stepping in to a wider issue here, I just picked a name! Let's go with whatever is standard to the protocol, no need to accommodate my made-up names.

jcalabro:
This was all interesting context, thanks for the review! As I said over here, I'm in favor of closing this one in favor of your [PR here](https://github.com/bluesky-social/indigo/pull/1024#pullrequestreview-2768225894), then we can come back and do further refactors to make better use of echo later if we desire? Closing, but if you disagree we can re-open and take a different course of action.

PR #1031, opened 2025-04-15
batch of golangci-lint fixes in `atproto/*`

bnewbold:
This isn't everything, just a first pass. There were some real issues with errors not getting wired through.

I'm kind of ambivalent about the pattern:

```go
	defer func() {
		_ = resp.Body.Close()
	}()
```

but do want to make any non-defer error returns explicit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ test-coverage.out
 # Don't ignore this file itself, or other specific dotfiles
 !.gitignore
 !.github/
+!.golangci.yaml
 
 # Don't commit your (default location) creds
 bsky.auth
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,17 @@
+version: "2"
+
+issues:
+  max-issues-per-linter: 0
+  max-same-issues: 0
+
+linters:
+  disable:
+    - errcheck
+    - gocritic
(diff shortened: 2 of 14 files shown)


bnewbold:
will need to dig in on the MST test failure; pretty sure it isn't a true failure, that the "prove" function just needs a way to work with partial trees.

jcalabro:
I'd recommend adding a golangci-lint config file and turning on the [errcheck linter's](https://golangci-lint.run/usage/linters/#errcheck) `disable-default-exclusions` rule. I also am not a fan of having to do spurious errchecks on `resp.Body.Close()` and similar cases. [More info here](https://github.com/kisielk/errcheck?tab=readme-ov-file#excluding-functions)

jcalabro:
This isn't running golangci-lint in CI yet, is that intentional? Do you want to get it running locally first, then switch it on in CI?

bnewbold:
- added a golangci.yaml file; did not include formatting and `interface{}` to `any` change for now
- enabled `disable-default-exclusions` as discussed above, and removed from this PR the "defer"-impacting error checks
- tried to narrow the `proveMutation` error check in MST code

jcalabro (approved):
Rad!

PR #1032, opened 2025-04-15
remove invalid JSON 'const=' struct field tags

bnewbold:
I assumed this worked, but it doesn't!

Found via golangci-lint.

### DIFF
--- a/atproto/data/blob.go
+++ b/atproto/data/blob.go
@@ -27,7 +27,7 @@ type LegacyBlobSchema struct {
 }
 
 type BlobSchema struct {
-	LexiconTypeID string  `json:"$type,const=blob" cborgen:"$type,const=blob"`
+	LexiconTypeID string  `json:"$type" cborgen:"$type,const=blob"`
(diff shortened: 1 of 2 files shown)


jcalabro (commented):
Wondering if there was a non-standard json encoder/decoder that was in use that does support these tags?

Should we be actually setting these const values in the places we expect those values to be set? Everything is currently just going through as the zero-value.

For reference with the stdlib's json package: https://go.dev/play/p/e3-SiikP8NA

bnewbold:
`cbor-gen` supports it, and I was probably copying that: https://github.com/whyrusleeping/cbor-gen?tab=readme-ov-file#field-tags

For the `indigo:atproto/blobs` usage, the field does get set in the constructor.

The `indigo:atproto/lexicon` usage is currently used for parsing JSON files (which do include this, used as the discriminator), and not currently for generating schemas (eg, none of these types have "New" methods).

jcalabro (approved):

PR #1039, opened 2025-04-17
Fix Relay Admin Page Auth

jcalabro:
Found this while doing a read-through of the new code and poking around. Without this change, when attempting to log in to the local admin UI, I get `Failed to validate Admin Token: Status 403`.

Either the UI should send basic auth, or the server handler should receive a bearer token. I just went with the server-side change, but if you'd prefer that the UI sends basic auth, happy to do that instead.

Screenshot 2025-04-17 at 17 45 31

### DIFF
--- a/cmd/relay/service.go
+++ b/cmd/relay/service.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"encoding/base64"
+	"fmt"
 	"log/slog"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


bnewbold:
The history here is that the relay (BGS / bigsky) was doing Bearer auth, but most of our atproto services use "atproto Admin Auth" (see https://atproto.com/specs/xrpc).

If you are hitting problems with the admin UI, check if you are using the `cmd/relay/relay-admin-ui` copy, which does `Basic` auth. I will need to update docs and Makefile to use that version.

bnewbold:
Should be working from the docker container builds? here was the commit to switch to Basic auth:
https://github.com/bluesky-social/indigo/commit/c9586bcbc4422e2e66bc0d898c2828439809fb5f

jcalabro:
I see the issue: the [Makefile](https://github.com/bluesky-social/indigo/blob/6b0010a646ae113a07c5a09fb6e9d483376f65bc/Makefile#L101-L105) is pointing at the BGS admin UI which does not have that fix applied.

I see several tasks in there that are pointing at bgs; should I update them all to cmd/relay? Don't want to break existing developer workflows, though I just did a quick grep through the `.github` folder, so it should be safe from a CI/build perspective to rename all those tasks?

jcalabro:
Thanks for fixing! #1048

PR #1044, opened 2025-04-22
support multiple admin passwords

bnewbold:
- relay support for multiple admin passwords at the same time (to make secret rotation easier)
- have a parallel PR going with README updates; will mention in that PR that the env var is comma-separated, so passwords can't contain a comma
- does not change the arg name or env var name, so existing configs work fine
- also switches to constant-time string comparison using go stdlib (more secure)

### DIFF
--- a/cmd/relay/service.go
+++ b/cmd/relay/service.go
@@ -210,13 +210,15 @@ func (svc *Service) checkAdminAuth(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) error {
 		hdr := c.Request().Header.Get("Authorization")
 		if hdr == "" {
-			return echo.ErrForbidden
+			c.Response().Header().Set("WWW-Authenticate", "Basic")
+			return echo.ErrUnauthorized
 		}
 		for _, val := range validAuthHeaders {
 			if subtle.ConstantTimeCompare([]byte(hdr), []byte(val)) == 1 {
 				return next(c)
 			}
 		}
-		return echo.ErrForbidden
+		c.Response().Header().Set("WWW-Authenticate", "Basic")
+		return echo.ErrUnauthorized
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/relay/service.go
```diff
@@ -197,12 +198,25 @@ func (svc *Service) Shutdown() []error {
...
+		if hdr == "" {
 			return echo.ErrForbidden
 		}
-		return next(c)
+		for _, val := range validAuthHeaders {
+			if subtle.ConstantTimeCompare([]byte(hdr), []byte(val)) == 1 {
+				return next(c)
+			}
+		}
+		return echo.ErrForbidden
```
devinivy, line 220: Bit of a nit, but since the user was not able to authenticate I'd call this a 401 Unauthorized.  As opposed to the user authenticated but does not have access (403 Forbidden).  Doing it by the book we'd also send a `WWW-Authenticate: Basic` back (technically a "must").

devinivy: (Another instance of this above when the authorization header is missing.)

cmd/relay/service.go
```diff
@@ -197,12 +198,25 @@ func (svc *Service) Shutdown() []error {
...
+	// for paths that this middleware is applied to, enforce that the auth header must exist and match one of the known passwords
 	return func(c echo.Context) error {
 		hdr := c.Request().Header.Get("Authorization")
-		if hdr != headerVal {
+		if hdr == "" {
 			return echo.ErrForbidden
 		}
-		return next(c)
+		for _, val := range validAuthHeaders {
+			if subtle.ConstantTimeCompare([]byte(hdr), []byte(val)) == 1 {
```
devinivy, line 216: Nice!

devinivy (approved):

cmd/relay/service.go
bnewbold: good catch!

PR #1046, opened 2025-04-23
implement sibling relay admin request forwarding

bnewbold:
This is the sketch of the feature, haven't tested it yet.

In particular, realized we don't have direct access between our multiple prod relay instances across WAN. Might need rainbow proxying support? Or some other hole-punch.


bnewbold:
did this in a separate PR

PR #1047, opened 2025-04-23
relay pull-hosts command (for public bootstrapping)

bnewbold:
This helper command queries an existing relay instance for known hosts, and adds them to the local instance (if they aren't already). The intention is to make bootstrapping a new full-network relay easier (eg, avoid wrangling bash script pipelines and admin API requests).

It does some filtering and checks to try and skip inactive/stale hosts: only "active" (or "idle") hosts which have seen at least one event remotely (aka, `seq` is positive), and does an (optional) HTTP request to `describeServer` on the host to check it is live.

This command should only be run if the overall relay isn't already running as a daemon. It is safe to run it multiple times, or after the relay has been running for some time.

(this code is pretty off-the-cuff; I've run it a few times to test, and it isn't critical to our prod use, but could definitely use eyes)

### DIFF
--- a/cmd/relay/pull.go
+++ b/cmd/relay/pull.go
@@ -43,6 +43,11 @@ var cmdPullHosts = &cli.Command{
 			Value:   cli.NewStringSlice("*.host.bsky.network"),
 			EnvVars: []string{"RELAY_TRUSTED_DOMAINS"},
 		},
+		&cli.BoolFlag{
+			Name:    "skip-host-checks",
+			Usage:   "don't run describeServer requests to see if host is a PDS before adding",
+			EnvVars: []string{"RELAY_SKIP_HOST_CHECKS"},
+		},
 	},
 }
 
@@ -56,6 +61,8 @@ func runPullHosts(cctx *cli.Context) error {
 	client := xrpc.Client{
 		Host: cctx.String("relay-host"),
 	}
+
+	skipHostChecks := cctx.Bool("skip-host-checks")
 
 	dir := identity.DefaultDirectory()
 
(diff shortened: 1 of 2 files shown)


cmd/relay/pull.go
```diff
@@ -0,0 +1,150 @@
...
+				return fmt.Errorf("%w: %s", err, h.Hostname)
+			}
+			accountLimit := r.Config.DefaultRepoLimit
+			trusted := relay.IsTrustedHostname(hostname, r.Config.TrustedDomains)
+			if trusted {
+				accountLimit = r.Config.TrustedRepoLimit
+			}
+
+			if !skipHostChecks {
+				if err := checker.CheckHost(ctx, "https://"+hostname); err != nil {
```
jcalabro, line 126: Does this need to take in to account `noSSL`?

jcalabro (approved):
Only one question (`noSSL`) other than that looks good to me!

Future improvements could be firing up some number of worker routines to check hosts concurrently, but if we're at a small enough volume for now that this is good enough for your needs, then 👍

cmd/relay/pull.go
bnewbold: good point. I added a check to just skip any `noSSL` hosts; this tool is intended for pulling and subscribing over the open net, not internal/private hosts.

cmd/relay/pull.go
```diff
@@ -0,0 +1,150 @@
...
+	// NOTE: setting evtmgr to nil
+	r, err := relay.NewRelay(db, nil, dir, relayConfig)
+	if err != nil {
+		return err
+	}
+
+	checker := relay.NewHostClient(relayConfig.UserAgent)
+
+	cursor := ""
+	var size int64 = 500
```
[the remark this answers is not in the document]
bnewbold: added a config flag

PR #1049, opened 2025-04-23
rainbow: fix getRepo redirects

bnewbold:
Rainbow was accidentially following HTTP redirects and streaming back full CAR files when handling getRepo requests.

This correctly passes through the HTTP 3xx response back to the client, which then does the request to the PDS.

Note that proxying or returning the CAR directly isn't *forbidden*, it just isn't the behavior we want right now. We might eventually do proxying and coalescing/caching in rainbow/relay to reduce network load.

### DIFF
--- a/splitter/handlers.go
+++ b/splitter/handlers.go
@@ -72,7 +72,7 @@ func (s *Splitter) HandleComAtprotoSyncRequestCrawl(c echo.Context) error {
 			// new context to outlive original HTTP request
 			ctx := context.Background()
 			xrpcc := xrpc.Client{
-				Client: s.upstreamClient,
+				Client: s.peerClient,
 				Host:   hostname,
 			}
 			if err := comatproto.SyncRequestCrawl(ctx, &xrpcc, &body); err != nil {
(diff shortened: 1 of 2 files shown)


splitter/handlers.go
```diff
@@ -129,7 +129,13 @@ func (s *Splitter) ProxyRequest(c echo.Context, hostname, scheme string) error {
 	}
 	defer upstreamResp.Body.Close()
 
-	respWriter.Header()["Content-Type"] = []string{upstreamResp.Header.Get("Content-Type")}
+	// copy a subset of headers
+	for _, hdr := range []string{"Content-Type", "Content-Length", "Location"} {
```
jcalabro, line 133: Content-Encoding as well? Any other headers?

jcalabro (approved):
One question, otherwise 👍

splitter/handlers.go
devinivy: Just a headsup that I believe content-encoding and content-length get stripped if the go client handled the body's encoding (e.g. if it were gzipped).  My understanding is that vanilla `transport.RoundTrip()` can be helpful to leave the response untouched.

bnewbold: Yup, we don't want to pass-through any encoding/chunking headers from upstream. In our situation there will also be haproxy involved, which might again de-compress/re-compress any response bodies.

I would defer digging in to lower-level raw stream proxying for now, until we have an actual perf issue on our hands. I'd be a bit worried about complexity and debugging corner-cases.

PR #1051, opened 2025-04-24
yet more relay patches and tweaks

bnewbold:
This is a working branch.

### DIFF
--- a/cmd/relay/README.md
+++ b/cmd/relay/README.md
@@ -135,6 +135,8 @@ We strongly recommend running docker in "host networking" mode when operating a 
 
 ### Bootstrapping Host List
 
+Before bulk-adding hosts, you should probably increase the "new-hosts-per-day" limit, at least temporarily.
+
 The relay comes with a helper command to pull a list of hosts from an existing relay. You should shut the relay down first and run this as a separate command:
 
     ./relay pull-hosts
--- a/cmd/relay/README.md
+++ b/cmd/relay/README.md
@@ -110,6 +110,10 @@ Be sure to double-check bandwidth usage and pricing if running a public relay! B
 
 The relay admin interface has flexibility for many situations, but in some operational incidents it may be necessary to run SQL commands to do cleanups. This should be done when the relay is not actively operating. It is also recommended to run SQL commands in a transaction that can be rolled back in case of a typo or mistake.
 
+On the public web, you should probably run the relay behind a load-balancer or reverse proxy like `haproxy` or `caddy`, which manages TLS and can have various HTTP limits and behaviors configured. Remember that WebSocket support is required.
+
+The relay does not resolve atproto handles, but it does do DNS resolutions for hostnames, and may do a burst of resolutions at startup. Note that the go runtime may have an internal DNS implementation enabled (this is the default for the Dockerfile). The relay *will* do a large number of DID resolutions, particularly calls to the PLC directory, and particularly after a process restart when the in-process identity cache is warming up.
+
 ### PostgreSQL
 
 PostgreSQL is recommended for any non-trival relay deployments. Database configuration is passed via the `DATABASE_URL` environment variable, or the corresponding CLI arg.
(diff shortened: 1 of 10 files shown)


cmd/relay/README.md
```diff
@@ -127,10 +131,21 @@ This service currently uses `gorm` to automatically run database migrations as t
 
 The relay is relatively easy to build and operate as as simple executable, but there is also Dockerfile in this directory. It can be used to build customized/patched versions of the relay as a container, republish them, run locally, deploy to servers, deploy to an orchestrated cluster, etc.
 
-We strongly recommend running docker in "host networking" mode when operating a full-network relay.
+We strongly recommend running docker in "host networking" mode when operating a full-network relay. You may also want to use something other than default docker log management (eg, `svlogd`).
```
devinivy, line 134: May just be worth a note that these are both for performance / resource utilization reasons.

devinivy: (As opposed to, say, correctness or stability reasons.)

cmd/relay/stubs.go
```diff
@@ -71,7 +71,7 @@ func (s *Service) HandleComAtprotoSyncListHosts(c echo.Context) error {
 	}
 
 	out, handleErr := s.handleComAtprotoSyncListHosts(c, cursor, limit)
-	if handleErr != nil {
+	if handleErr != nil || out == nil {
```
devinivy, line 74: When does this case get hit?  I.e. when there was no error but `out == nil`.

devinivy (approved):

cmd/relay/relay/host.go
```diff
@@ -53,13 +53,20 @@ func (r *Relay) GetHostByID(ctx context.Context, hostID uint64) (*models.Host, e
...
+// If `everActive` flag is true, then only hosts which have ever had a message received (seq > 0) are returned.
+func (r *Relay) ListHosts(ctx context.Context, cursor int64, limit int, everActive bool) ([]*models.Host, error) {
 
 	// filters for accounts which have seen at least one event
 	// TODO: also filter based on status?
 	hosts := []*models.Host{}
-	if err := r.db.WithContext(ctx).Model(&models.Host{}).Where("id > ? AND last_seq > 0", cursor).Order("id").Limit(limit).Find(&hosts).Error; err != nil {
-		return nil, err
+	if everActive {
+		if err := r.db.WithContext(ctx).Model(&models.Host{}).Where("id > ? AND last_seq > 0", cursor).Order("id").Limit(limit).Find(&hosts).Error; err != nil {
```
jcalabro, line 63: This is a nit, but you could do:

```go
where := "id > ?"
if everActive {
        where += " AND last_seq > 0"
}

hosts := []*models.Host{}
if err := r.db.WithContext(ctx).Model(&models.Host{}).Where(where, cursor).Order("id").Limit(limit).Find(&hosts).Error; err != nil {
        // ...
```

cmd/relay/service.go
```diff
@@ -101,6 +101,9 @@ func (svc *Service) startWithListener(listen net.Listener) error {
 	e.Use(svcutil.MetricsMiddleware)
 
 	e.HTTPErrorHandler = func(err error, c echo.Context) {
+		if c.Response().Committed {
```
jcalabro, line 104: For my own understanding, when would we be firing a http error handler after already writing the response to the caller?

cmd/relay/stubs.go
bnewbold: This relates to HTTP error responses getting spread around in a not great way. If the the method being called (`s.handleComAtprotoSyncListHosts`) did a `c.JSON()` call to return a specific XRPC error internally, then the HTTP response has already been written, and this stub method should not attempt to return an error (which would result in duplicate HTTP body data), or to try and write the success JSON (which would be a `null` string appended to the HTTP body data).

I think this pattern needs a broader structural fix: for example being rigorous about only writing HTTP bodies in the "stub" functions, not the "handler" functions. But i'm not sure what that pattern should be across all our projects, so I put in this quick fix for now.

cc: @jcalabro on error handling stuff, we had discussed a bit and I think this was the kind of complex situation I was trying to remember.

cmd/relay/service.go
bnewbold: This relates to the "empty response" stuff above: if we rigorously only wrote HTTP bodies in "stub" functions, then it would be easier to reason able this, but I think there are many cases where we do XRPC error responses (eg, with a specific message or name) and they might not all get caught?

I added this as a sanity check / defensive coding; it would probably be good to have a house style / norm about how we do XRPC errors with echo across all of our services.

PR #1057, opened 2025-04-29
basic IP filtering net.Dialer for SSRF protection

bnewbold:
Pulls in code from this 2019 blog post which still seems like the best pattern: https://www.agwa.name/blog/post/preventing_server_side_request_forgery_in_golang

I looked around a bit and there are some packages that do similar things, but this seemed the best, and only a few dozen lines of code so just vendoring it in.

This code has no deps, so I put it in a sub-package of `util`, which will make it painless to depend on from other code (eg, OAuth client, relay) without entangling with any new deps.

### DIFF
--- a/util/ssrf.go
+++ b/util/ssrf.go
@@ -1,0 +1,128 @@
+/*
+ * Written in 2019 by Andrew Ayer.
+ * Patched 2025, Bluesky Social PBC.
+ *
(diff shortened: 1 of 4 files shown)


devinivy (approved):

jcalabro (commented):
Nice!

PR #1058, opened 2025-04-29
basic JWK support in crypto SDK

bnewbold:
Basic JWK serialization of the curves we use for atproto crypto (P-256, K-256), for public keys only. This will get used in OAuth implementation.

The decision to implement this directly instead of using something like https://github.com/lestrrat-go/jwx is entangled with deciding to use `golang-jwt` for service auth implementation (which is distinct from OAuth implementation, and doesn't directly require JWKs). I think the secp256k1 experimental support in `jwx/jwt` would pull in a new dependency/implementation, and I like our current choice, and it doesn't seem as easy to wedge in a new curve type/implementation with `jwx` family of packages.

On the other hand, I sure don't love parsing those uncompressed bytes out in to `(x, y)` coordinates!

I thought about only exposing `[]byte` JSON instead of the `JWK` struct itself, but feels more idiomatic and flexible to just expose the struct. For example, will make it easier for calling code to deal with JWK sets (arrays of keys, used in OAuth client metadata).

### DIFF
--- a/atproto/crypto/jwk.go
+++ b/atproto/crypto/jwk.go
@@ -1,0 +1,124 @@
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"math/big"
+
+	secp256k1 "gitlab.com/yawning/secp256k1-voi"
+	secp256k1secec "gitlab.com/yawning/secp256k1-voi/secec"
+)
+
+// Representation of a JSON Web Key (JWK), as relevant to the keys supported by this package.
+//
+// Expected to be marshalled/unmarshalled as JSON.
+type JWK struct {
+	Algorithm string  `json:"alg"`
+	Curve     string  `json:"crv"`
+	X         string  `json:"x"` // base64url, no padding
+	Y         string  `json:"y"` // base64url, no padding
+	Use       string  `json:"use,omitempty"`
+	KeyID     *string `json:"kid,omitempty"`
+}
+
+// Loads a [PublicKey] from JWK (serialized as JSON bytes)
+func ParsePublicJWKBytes(jwkBytes []byte) (PublicKey, error) {
+	var jwk JWK
+	if err := json.Unmarshal(jwkBytes, &jwk); err != nil {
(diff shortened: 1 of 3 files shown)


bnewbold:
Found and fixed a bug with encoding. Have this working with a proof-of-concept OAuth implementation (and thus interop with typescript codebase), and thus have more confidence in it.

I think this PR is ready for final review and merge.

atproto/crypto/jwk.go
```diff
@@ -0,0 +1,124 @@
...
+		X:       base64.RawURLEncoding.EncodeToString(k.pubP256.X.Bytes()),
+		Y:       base64.RawURLEncoding.EncodeToString(k.pubP256.Y.Bytes()),
+	}
+	return &jwk, nil
+}
+
+func (k *PublicKeyK256) JWK() (*JWK, error) {
+	raw := k.UncompressedBytes()
+	if len(raw) != 65 {
+		return nil, fmt.Errorf("unexpected K-256 bytes size")
```
jcalabro, line 113: This code path is unreachable. `k.UncompressedBytes()` always returns 65 bytes. We could consider removing the `error` in the return value from the `JWK() (*JWK, error)` interface method, but could see a case for leaving it around to future-proof

atproto/crypto/jwk_test.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Fatal(err)
+	}
+
+	pk, ok := pub.(*PublicKeyK256)
+	if !ok {
+		t.Fatal()
+	}
+	jwk, err := pk.JWK()
+	if err != nil {
+		t.Fatal(err)
```
jcalabro, line 93: One nice small thing to note is you can use require just like assert, but require stops the test now instead of allowing it to run to the end. i.e.

```go
jwk, err := pk.JWK()
require.NoError(err)
```

jcalabro (approved):

atproto/crypto/jwk_test.go
bnewbold: I think we should probably use `testify/require` consistently across the SDK tests, or stick with `t.Fatal`. Right now it would be a new dep, and need to be initialized separately in every test function; would be nice if it was part of `testify/assert` package.

I could go either way on it!

PR #1059, opened 2025-04-29
admin auth and service helpers in SDK

bnewbold:
- `net/http` server middleware for "atproto admin auth" (which is just Basic auth with user "admin), supporting multiple passwords for operational flexibility (eg, new relay could use this)
- inter-service auth validation, including `net/http` middleware. fairly complete and correct, though might have some performance issues (purging/retrying identity resolution; and parsing keys on every request)

The service auth code is loosely inspired by code in Discover (which is not open source), which has been running in prod a long time now. For example the test cases. But that code caches parsed keys much more aggressively (with an in-process LRU).

There are some TODOs in here about XRPC error responses (aka, reply with JSON). I'm not sure if we want to manage that here in middleware library code, or leave it to service error handler to fill in.

The logic around LXM validation/enforcement could also use review, both of the code and against the atproto XRPC specs.

Note that this package intentionally does not use "XRPC" in method/variable names, as we are slowly deprecating that terminology.


atproto/auth/jwt.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("%w: resolving DID (%s): %w", jwt.ErrTokenInvalidIssuer, did, err)
+		}
+		return ident.PublicKey()
+	}
+}
+
+func randomNonce() string {
+	buf := make([]byte, 16)
+	rand.Read(buf)
```
jcalabro, line 115: This is importing `math/rand`, did you mean to import `crypto/rand`?

atproto/auth/jwt.go
```diff
@@ -0,0 +1,147 @@
...
+	var sm *signingMethodAtproto
+
+	// NOTE: could also have a crypto.PrivateKey.Alg() method which returns a string
+	switch priv.(type) {
+	case *crypto.PrivateKeyP256:
+		sm = signingMethodES256
+	case *crypto.PrivateKeyK256:
+		sm = signingMethodES256K
+	default:
+		return "", fmt.Errorf("unknown signing key type")
```
jcalabro, line 142: One thing that might make this error a bit more meaningful is to return the type name itself in the error value, something like:

```go
return "", fmt.Errorf("unsupported signing private key: %T")
```

atproto/auth/jwt.go
bnewbold: It's really more of a "unique nonce" than needing strong un-predictability, I think, so I used `math/rand`.

But it is security-related, and probably won't be a performance / blocking issue on modern kernels/runtimes, so i'll just update to `crypto/rand`.

atproto/auth/http.go
```diff
@@ -0,0 +1,79 @@
...
+			for _, pw := range adminPasswords {
+				if subtle.ConstantTimeCompare([]byte(pw), []byte(password)) == 1 {
+					handler(w, r)
+					return
+				}
+			}
+		}
+		w.Header().Set("WWW-Authenticate", `Basic realm="admin", charset="UTF-8"`)
+		// TODO: XRPC error body?
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
```
jcalabro: I'm finding myself agreeing with these comments - let's use the json error+message xrpc shape?

atproto/auth/http.go
```diff
@@ -0,0 +1,97 @@
...
+				if nil == err {
+					lxm = &nsid
+				}
+			}
+
+			did, err := v.Validate(r.Context(), parts[1], lxm)
+			if err != nil {
+				w.Header().Set("WWW-Authenticate", "Bearer")
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusUnauthorized)
```
jcalabro, line 73: `http.StatusForbidden`?

jcalabro (approved):
One more question, otherwise 👍

atproto/auth/http.go
bnewbold: I think 401 Unauthorized is the correct status here, though I had to look it up a couple times.

Skimming both [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401) and [RFC 7235](https://www.rfc-editor.org/rfc/rfc7235#section-3.1), 401 should be used if auth was attempted but the credentials are not valid. For example, if a password is provided but wrong, or a token has expired.

The semantics are a bit funky here because the JWT mixes authn (who) and authz (what). But the validation which is being checked in this specific code path is just validating the credential itself against the request, not doing application-layer permission checks to a backend or anything like that, so I think 401 is best.

jcalabro: Ah true!

PR #1060, opened 2025-04-30
Small cleanup for relay host code

James-Gilbert-:
Minor changes: 

1. Removes duplicate WithContext() call
2. Uses `First()` instead of `Find()` where only one struct is expected


bnewbold (approved):
good catches, thanks!

PR #1061, opened 2025-04-30
relay state synchronization: requestCrawl, admin endpoints

bnewbold:
- more complete rainbow forwarding to the relay backing it ("upstream relay")
- do requestCrawl forwarding in relay. code in rainbow is still there, but will remove config for rainbows deployed on new relay instances ("relay1")
- have relay also forward (some) admin endpoints to the same set of "sibling" relays that requestCrawl gets forwarded
- HTTP header hygiene to prevent self-slurps: relay and rainbow set `Server` and `Via` headers when appropriate, and relay checks this when slurping. include the exact string `atproto-relay` in user-agents.
- integrate basic SSRF checks: when doing CheckHost (deciding whether to add hostname to database), and at slurp time (when actually opening websocket)

Expectation is that this results in the important info getting synchronized between our multiple new prod relays, while preventing "self-slurps" and forwarding loops:

- requestCrawl
- admin add-host
- domain bans
- account takedowns
- PDS takedowns
- PDS account limit changes

~This is marked Draft until I can self-review and test a bit more, but I think all the pieces are here~. Don't have a test framework for this part of relay/rainbow interaction :cry:.


cmd/relay/forward.go
```diff
@@ -0,0 +1,91 @@
...
+		// copy some headers from inbound request
+		for _, hdr := range []string{"Accept", "User-Agent", "Authorization", "Content-Type"} {
+			val := req.Header.Get(hdr)
+			if val != "" {
+				upstreamReq.Header.Set(hdr, val)
+			}
+		}
+
+		// add Via header (critical to prevent forwarding loops)
+		upstreamReq.Header.Add("Via", s.relay.Config.UserAgent)
```
devinivy, line 76: Two little things, not critical in any way:
 - we might append to the `Via` from the incoming request.
 - each item in the `Via` header has two parts: first the http version, e.g. `1.1`, and then second a name not containing whitespace.  An example might be `1.1 bsky-relay1-us-east, 1.1 bsky-relay2-us-west`.

cmd/relay/forward.go
```diff
@@ -0,0 +1,91 @@
...
+
+		upstreamResp, err := client.Do(upstreamReq)
+		if err != nil {
+			s.logger.Warn("forwarded admin HTTP request failed", "method", req.Method, "sibling", hostname, "url", u.String(), "err", err)
+			continue
+		}
+		if upstreamResp.StatusCode != http.StatusOK {
```
devinivy, line 83: Might add a little give to loosen this to the 2xx range.

cmd/relay/forward.go
bnewbold: Good catch on the `Via` header, I had looked it up but forgot: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Via

I went with the full proto, followed by the pseudoname "atproto-relay", for example "HTTP/1.1 atproto-relay". To prevent loops it is critical for this to reliably have a detectable sub-string (like "atproto-relay"), so I went with that over config or trying to auto-detect.

The stdlib `Header.Add()` method does append to any existing values (unlike `Header.Set()`); and the code above which checks headers uses `Header.Values()` to iterate over all values.

devinivy: Nice!  

Re: adding vs setting the header, my understanding was that it was adding it but that there were no values set from the original incoming `req`.  It is probably alright either way for our purposes 👍

devinivy (approved):

cmd/relay/forward.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	for _, ua := range req.Header.Values("User-Agent") {
+		if strings.Contains(ua, "atproto-relay") {
+			s.logger.Info("not re-forwarding request to sibling relay", "header", "User-Agent", "value", ua)
+			return
+		}
+	}
+
+	// TODO: could turn this in to a shared/persistent client
+	client := http.Client{
```
jcalabro, line 40: This could be a global to enable connection reuse where possible (your comment agrees 😄)

cmd/relay/forward.go
```diff
@@ -0,0 +1,91 @@
...
+
+		upstreamResp, err := client.Do(upstreamReq)
+		if err != nil {
+			s.logger.Warn("forwarded admin HTTP request failed", "method", req.Method, "sibling", hostname, "url", u.String(), "err", err)
+			continue
+		}
+		if !(upstreamResp.StatusCode >= 200 && upstreamResp.StatusCode < 300) {
+			respBytes, _ := io.ReadAll(upstreamResp.Body)
+			s.logger.Warn("forwarded admin HTTP request failed", "method", req.Method, "sibling", hostname, "url", u.String(), "statusCode", upstreamResp.StatusCode, "body", string(respBytes))
+			continue
```
jcalabro, line 86: Need an `upstreamResp.Body.Close()` before this `continue`

cmd/relay/handlers_admin.go
```diff
@@ -17,14 +18,13 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
-// this is the same as the regular com.atproto.sync.requestCrawl endpoint, except it sets a flag to bypass configuration checks
+// This endpint is basically the same as the regular com.atproto.sync.requestCrawl endpoint, except it sets a flag to bypass configuration checks.
```
jcalabro, line 21: Typo (endpint)

cmd/relay/forward.go
bnewbold: realized we actually already had such a client set up but wasn't using it! wire it up

PR #1069, opened 2025-05-04
lexgen against atproto 7049117ea28105666b762a7a0ca7c1246ad8b98b

bnewbold:
Just keeping things in sync. Ozone types only.


foysalit (approved):
Thanks!

PR #1070, opened 2025-05-04
lexgen: use lexutil.LexClient interface for codegen

bnewbold:
Instead of referring directly to `xrpc.Client` (and the `Do()` method on that type) in lexgen output, define a new client interface with roughly the same function signature. The way this is done, the `xrpc.Client` implementation is basically untouched, so any existing code depending on, including in other git repos, it should not be impacted.

The motivation for this is to allow alternative client implementations. For example, @haileyok has an OAuth client (https://github.com/haileyok/atproto-oauth-golang), and i'm working on a flexible SDK client.

This PR doesn't include an actual lexgen run. There are a bunch of unrelated changes when I do lexgen right now (see https://github.com/bluesky-social/indigo/pull/1069), and it is easier to review without all the (many!) changes). However, everything "Just Works" with both old and new lexgen output. This PR should be mergable as-is.

I did try a lexgen run and everything worked as expected. I also manually updated the "record type agnostic" helper code, which gives a feel for what the full lexgen diff will look like.

### DIFF
--- a/api/agnostic/actorgetPreferences.go
+++ b/api/agnostic/actorgetPreferences.go
@@ -7,7 +7,7 @@ package agnostic
 import (
 	"context"
 
-	"github.com/bluesky-social/indigo/xrpc"
+	"github.com/bluesky-social/indigo/lex/util"
 )
 
 // ActorGetPreferences_Output is the output of a app.bsky.actor.getPreferences call.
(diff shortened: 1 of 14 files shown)


bnewbold:
FWIW discussed this with @whyrusleeping yesterday and he is good with the general direction.

jcalabro (approved):

PR #1073, opened 2025-05-05
Changed error handling during blob export command in goat to help with account migration

orual:
Prints the error and continues rather than bailing entirely during blob export. Slightly worse in the "i have a connection problem" scenario, but means repository exports where some blobs are missing actually work now rather than getting stuck.

Was trying to migrate my main to my own PDS and ran into it, and figured addressing it would be both simple enough and useful to others trying to do the same.

### DIFF
--- a/cmd/goat/blob.go
+++ b/cmd/goat/blob.go
@@ -113,7 +113,8 @@ func runBlobExport(cctx *cli.Context) error {
 			}
 			blobBytes, err := comatproto.SyncGetBlob(ctx, &xrpcc, cidStr, ident.DID.String())
 			if err != nil {
-				return err
+				fmt.Printf("%s\tnot found\n", blobPath)
+
 			}
 			if err := os.WriteFile(blobPath, blobBytes, 0666); err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


cmd/goat/blob.go
```diff
@@ -113,7 +113,8 @@ func runBlobExport(cctx *cli.Context) error {
 			}
 			blobBytes, err := comatproto.SyncGetBlob(ctx, &xrpcc, cidStr, ident.DID.String())
 			if err != nil {
-				return err
+				fmt.Printf("%s\tnot found\n", blobPath)
```
bnewbold, line 116: can you update the text here from "not found" to "failed"?

bnewbold (approved):
this is definitely an improvement! makes it possible to run-to-completion, then just re-run to retry any blobs which failed.

I left a small note about how to log (the error might be a network error or something other than "not found"), but with that change can merge.

cmd/goat/blob.go
orual: yeah, sure. can also print the actual error in addition or instead of, if that's better. thought i actually pushed the latter (appending the error to the line), but apparently not.

PR #1074, opened 2025-05-05
subscribeRepos Client Reconnects

jcalabro:
We've been seeing periodic cases where firehose subscribers get booted off because they [fail to keep up](https://github.com/bluesky-social/indigo/blob/9c19d0c438ef38657dff9f2f55caa167271287c4/cmd/relay/stream/eventmgr/event_manager.go#L70-L94). Thus far, we've seen this during network turbulence events with one of our ISPs.

The candidate solution here is to add some number of retries on websocket closures now that the firehose will properly clean up network conns if the consumer is too slow (#1066).

I was looking at splitter.go's [subscribeWithRedialer](https://github.com/bluesky-social/indigo/blob/9c19d0c438ef38657dff9f2f55caa167271287c4/splitter/splitter.go#L354) as prior art, but I wanted to specifically check for certain retry-able errors and not infinite loop.

I hemmed and hawed about the look of this API as well as the function name for a while as well since there are a lot of options being passed in, and I eventually opted for a struct with a default function. IMO it's the simplest API to use, and it provides reasonable defaults. If there's a better pattern, I'm 100% game for that though.

### DIFF
--- a/events/consumer.go
+++ b/events/consumer.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
+	"math/rand"
 	"net"
 	"net/http"
 	"time"
@@ -186,6 +188,7 @@ func HandleRepoStreamRobust(ctx context.Context, opts *HandleRepoStreamRobustOpt
 				return fmt.Errorf("failed to dial host (max retries exceeded): %w", err)
 			}
 
+			time.Sleep(backoffDuration(failures))
 			failures++
 			continue
 		}
@@ -201,6 +204,7 @@ func HandleRepoStreamRobust(ctx context.Context, opts *HandleRepoStreamRobustOpt
 				return fmt.Errorf("failed to : %w", err)
 			}
 
+			time.Sleep(backoffDuration(failures))
 			failures++
 			continue
 		}
@@ -211,6 +215,21 @@ func HandleRepoStreamRobust(ctx context.Context, opts *HandleRepoStreamRobustOpt
 	}
 
 	return nil
+}
+
+func backoffDuration(retry int) time.Duration {
+	baseMS := float64(50)
+
+	// Exponential backoff: 2^retry * base
+	durationMS := math.Pow(2, float64(retry)) * baseMS
+
+	// Clamp
+	durationMS = min(1000, durationMS)
+
+	// Add jitter
+	jitter := durationMS * (0.25 + rand.Float64())
+
+	return time.Duration(jitter) * time.Millisecond
 }
 
 // HandleRepoStream
--- a/events/consumer.go
+++ b/events/consumer.go
@@ -6,13 +6,14 @@ import (
 	"io"
 	"log/slog"
 	"net"
+	"net/http"
 	"time"
 
 	"github.com/RussellLuo/slidingwindow"
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
-	"github.com/prometheus/client_golang/prometheus"
-
+	"github.com/carlmjohnson/versioninfo"
 	"github.com/gorilla/websocket"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 type RepoStreamCallbacks struct {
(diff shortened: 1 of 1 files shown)


events/consumer.go
```diff
@@ -5,14 +5,18 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
+	"math/rand"
```
ericvolp12, line 9: Do you want to use the `math/rand/v2` package?

jcalabro: Yup, thanks!

events/consumer.go
```diff
@@ -112,11 +116,137 @@ func (sr *instrumentedReader) Read(p []byte) (int, error) {
...
+	}
+	if opts.Header.Get("User-Agent") == "" {
+		opts.Header.Set("User-Agent", "indigo/"+versioninfo.Short())
+	}
+
+	upstream, err := url.Parse(opts.Upstream)
+	if err != nil {
+		return fmt.Errorf("invalid upstream url %q: %w", opts.Upstream, err)
+	}
+	upstream = upstream.JoinPath("/xrpc/com.atproto.sync.subscribeRepos")
```
ericvolp12, line 177: Do we want this path to always be hardcoded in here or should this be overridable?

events/consumer.go
```diff
@@ -112,11 +116,137 @@ func (sr *instrumentedReader) Read(p []byte) (int, error) {
...
+
+	retryableCloses := []int{
+		websocket.CloseNormalClosure,
+		websocket.CloseGoingAway,
+		websocket.CloseAbnormalClosure,
+	}
+
+	failures := 0
+	for failures <= opts.MaxReconnectAttempts {
+		con, _, err := opts.Dialer.DialContext(ctx, upstream.String(), opts.Header)
```
ericvolp12, line 191: How do we make sure reconnects use the correct cursor? When we reconnect we probably want to reconnect with a cursor of the last "committed" work we've done for the stream.

This may be a harder problem than we initially thought.

events/consumer.go
jcalabro: Decided to just allow the user to set the entire URL as an argument.

events/consumer.go
jcalabro: True, good catch! Take a look at 83b64976bde7d488d8f0a26582422f79c9aec9d7, I think this should do it, and it's not too hard?

ericvolp12: I think that ought to work for reconnects given the workers themselves are still alive during a connection drop and so there _should_ be a contiguous series of events "added" to the work queue. I wonder what happens to the context in these cases and would want to make sure connection drops don't cancel a context that makes workers fail to process work items until the reconnect (which would effectively drop work we've "added" and then that work would get skipped on reconnect because we advance the cursor when "adding" work).

jcalabro: When relay boots the consumer off, the top-level parent context isn't cancelled, so we're safe to that. However, I did need to move the `sched.Shutdown()` call out of the inner function and instead defer it in each of the callers.

jcalabro:
Alright, I'm marking this for review now and would be very appreciative of some eyes on this, particularly:

1. Design of the public-facing API function and its options. Is there a more idiomatic way to expose this without being overly cumbersome?
2. How do we feel about the really aggressive retry loop (it retries almost all errors without trying to disambiguate why)
    - I think this is what we'd want for our own consumers, so I assume that others will want this. Is that not the case?

My testing steps included the following: I ran this new code on the system that's been dropping connections on my local machine and added some `time.Sleep` calls to simulate being a slow consumer. I was able to observe dropping the connection after a while then successfully auto-reconnecting with the new cursor.

events/consumer.go
```diff
@@ -112,17 +117,165 @@ func (sr *instrumentedReader) Read(p []byte) (int, error) {
...
+		URL:                  u,
+		Dialer:               websocket.DefaultDialer,
+		Header:               http.Header{},
+		MaxReconnectAttempts: 10,
+		Log:                  slog.Default().With("system", "events"),
+	}
+}
+
+// The same as `HandleRepoStream`, but with auto-reconnects in the case of upstream disconnects
+func HandleRepoStreamRobust(ctx context.Context, opts *HandleRepoStreamRobustOptions) error {
```
ericvolp12, line 158: I think it might make sense since we've only got 2 required params to do something like this?
```
func HandleRepoStreamRobust(ctx context.Context, sched Scheduler, url string, opts *HandleRepoStreamRobustOptions) error {
```

PR #1075, opened 2025-05-06
Add ValidateInput method to atproto/lexicon

wdantuma:
This PR adds a ValidateInput ( counterpart of ValidateRecord ) to validate the request body and parameters against a lexicon for use in a pds implementation.

This PR also changes the term field to property in the error messages ( to align it with the terms used in the lexicon )


atproto/lexicon/lexicon.go
```diff
@@ -55,6 +56,46 @@ func validateRecordConfig(cat Catalog, recordData any, ref string, flags Validat
 	return validateObject(cat, s.Record, d, flags)
 }
 
+// Checks Lexicon schema (fetched from the catalog) for the given input and parameters, with optional flags tweaking default validation rules.
+//
+// 'inputData' is typed as 'any', but is expected to be 'map[string]any'
+// 'parameters' are the HTTP request parameters to be validated
+// 'ref' is a reference to the schema type, as an NSID with optional fragment. For records, the '$type' must match 'ref'
+// 'flags' are parameters tweaking Lexicon validation rules. Zero value is default.
+func ValidateInput(cat Catalog, inputData any, parameters url.Values, ref string, flags ValidateFlags) error {
```
bnewbold, line 65: I don't love "Input" as the term here, but need to think more about what I would use instead. "Request"? Should this be broken out by "ValidateQuery" / "ValidateProcedure" for the two API request types?

bnewbold:
Thanks for sending this over! We do need this to help implement services of many types. Relatedly, we have some early work on SDK support for server-side auth validation coming along.

I probably will not be able to give this a real review in the next two weeks. Will also want to have some test coverage.

atproto/lexicon/lexicon.go
wdantuma: I already renamed it to validateRequest, but the more i am  thinking about it  i think you're right, separate calls for ValidateQuery and ValidateProcedure are more aligned with the lexicon structure ( input is not a main type ). I shall make the changes and add test coverage.

wdantuma:
Moved this code out of indigo

PR #1080, opened 2025-05-15
bypass subscribe limits in /admin/pds/requestCrawl

mackuba:
`handleComAtprotoSyncRequestCrawl` should pass its boolean `admin` flag to `SubscribeToHost`, to let it bypass the daily new host limit when run as admin. Currently it just always passes `false`. The result is that when calling `goat relay admin host add` or the `/admin/pds/requestCrawl` endpoint, you get a "500 new host subscriptions temporarily disabled" error in response.


bnewbold (approved):
good catch, thanks!

PR #1082, opened 2025-05-20
update codegen

haileyok:


cmd/relay/stream/consumer.go
```diff
@@ -248,24 +239,6 @@ func HandleRepoStream(ctx context.Context, con *websocket.Conn, sched Scheduler,
 				}); err != nil {
 					return err
 				}
-			case "#handle":
```
bnewbold, line 251: I think should probably log a warning here for network observability. AKA, to keep track if other software is emitting these events, even if we don't process them or emit them ourselves.

Don't need to hold up this PR on that; I created a tracking issue here: https://github.com/bluesky-social/indigo/issues/1083

bnewbold (approved):
yup, this looks right to me! sheds a lot of lines of code, great.

the key thing is to not remove and re-number the evtKind enum in persister.

bnewbold:
Remembered that we should also get this branch merged: https://github.com/bluesky-social/indigo/pull/1020

PR #1090, opened 2025-06-05
add a 1ms delay between re-subscribing to PDS instances

bnewbold:
Assuming 5k active PDS instances, this spreads things out over 5 seconds. had been seeing burst of errors when restarting relay, which I suspect is related to too many sockets (and DNS lookups) at the same time.


ericvolp12 (approved):
Sure

PR #1100, opened 2025-06-16
OAuth client SDK

bnewbold:
Design goals:

- specific to atproto OAuth (not a general-purpose OAuth framework)
- implementation is correct and interoperable with atproto specification
- reasonably complete and flexible, though may make some opinionated implementation decisions to avoid footguns
- compatible with client SDK (aka, implement `AuthMethod`)
- oriented towards server-side (eg, BFF and integrations)
- supports "just authn" use-cases

Progress:

- [x] basic confidential client demo web interface
- [x] public client mode
- [x] localhost dev client mode
- [x] refactor core types and method attachments (eg, a session-agnositc OAuthClient struct with http.Client)
- [x] persist token callback (wired to ClientApp)
- [x] make PAR DPoP retries more specific (parse error response)
- [x] resolve XXX and TODO
- [x] multiple session support (and document this pattern)
- [x] document authn-only usecase
- [x] update `randomNonce`
- [x] add doc comments to major functions/types
- [x] more consistent DPoP capitalization in variables (?)
- [ ] fix DID/handle display in demo app
- [ ] ability to embed JWKs in client metadata directly (blocked on needing `key_ops`?)
- [ ] proactive detection and update (persist) when DPoP nonce changes
- [ ] remember token deletion endpoint as part of session; and add logout helper which calls it (if defined)
- [ ] mock tests (like service auth has)

### DIFF
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -417,6 +417,7 @@ func (app *ClientApp) SendAuthRequest(ctx context.Context, authMeta *AuthServerM
 		Scope:                   scope,
 		PKCEVerifier:            pkceVerifier,
 		RequestURI:              parResp.RequestURI,
+		AuthServerTokenEndpoint: authMeta.TokenEndpoint,
 		DPoPAuthServerNonce:     dpopServerNonce,
 		DPoPPrivateKeyMultibase: dpopPrivKey.Multibase(),
 	}


bnewbold:
This is finally ready for review! Recommend starting with `doc.go` and `HACKING.md` as an overview of the package.

Some self-review notes:

- there are still some smaller TODOs and tasks sprinkled around, but I think the overall structure is pretty set
- could probably use a pass over adding more comments and maybe renaming some things
- I don't love the code duplication to handle DPoP nonces for auth server endpoints: PAR, initial token request, and token refreshes. auth requests to the PDS have a different shape, but there are 3x implementations which are almost the same

atproto/auth/oauth/doc.go
```diff
@@ -0,0 +1,131 @@
...
+	doc := oauthApp.Config.ClientMetadata(oauthScope)
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(doc); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+```
+
+The login auth flow starts with a user identifier, which could be an atproto handle, DID, or a host URL. The high-level [StartAuthFlow()] method will resolve the identifier, send an auth request (PAR) to the server, persist request metadata in the [OAuthStore], and return a redirect URL for the user to visit:
```
DavidBuchanan314: "a host URL" -> "an authorization server URL"

atproto/auth/oauth/jwt_signing.go
```diff
@@ -0,0 +1,87 @@
...
+	if !sm.hash.Available() {
+		return jwt.ErrHashUnavailable
+	}
+
+	if len(sig) != sm.sigLen {
+		return jwt.ErrTokenSignatureInvalid
+	}
+
+	// NOTE: important to use using "lenient" variant here
+	return pub.HashAndVerifyLenient([]byte(signingString), sig)
```
DavidBuchanan314, line 59: Might be worth spelling out what "lenient" actually means here, for people reading this as a reference impl without broader atproto background. i.e. 'jwt spec does not require "low-s" ecdsa signatures, so we need to be able to accept them'

atproto/auth/oauth/doc.go
```diff
@@ -0,0 +1,131 @@
...
+	http.Redirect(w, r, redirectURL, http.StatusFound)
+}
+```
+
+The service then waits for a callback request on the configured endpoint. The [ProcessCallback()] method will load the earlier request metadata from the [OAuthStore], send an initial token request to the auth server, and validate that the session is consistent with the identifier from the begining of the login flow.
+
+```
+http.HandleFunc("GET /client-metadata.json", HandleClientMetadata)
+
+func HandleOAuthCallback(w http.ResponseWriter, r *http.Request) {
```
DavidBuchanan314: copy-paste error I assume, HandleFunc args are wrong

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	if info.State != state || info.AuthServerURL != authserverURL {
+		return nil, fmt.Errorf("callback params don't match request info")
+	}
+
+	tokenResp, err := app.SendInitialTokenRequest(ctx, authCode, *info)
+	if err != nil {
+		return nil, fmt.Errorf("initial token request: %w", err)
+	}
+
+	// verify against account/server from start of login
```
DavidBuchanan314, line 604: two thumbs up, failing to do this is a big footgun and it's great that we're handling it inside the sdk

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+
+	var tokenResp TokenResponse
+	if err := json.NewDecoder(resp.Body).Decode(&tokenResp); err != nil {
+		return nil, fmt.Errorf("token response failed to decode: %w", err)
+	}
+
+	return &tokenResp, nil
+}
+
+func (app *ClientApp) StartAuthFlow(ctx context.Context, username string) (string, error) {
```
DavidBuchanan314: s/username/identifier/ ?

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+		}
+	}
+
+	authserverMeta, err := app.Resolver.ResolveAuthServerMetadata(ctx, authserverURL)
+	if err != nil {
+		return "", fmt.Errorf("fetching auth server metadata: %w", err)
+	}
+
+	scope := scopeStr(app.Config.Scopes)
+	info, err := app.SendAuthRequest(ctx, authserverMeta, scope, username)
```
DavidBuchanan314: if a DID was passed initially, it'd still be a DID here - probably want to be passing the resolved handle?

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+
+	// persist auth request info
+	app.Store.SaveAuthRequestInfo(ctx, *info)
+
+	params := url.Values{}
+	params.Set("client_id", app.Config.ClientID)
+	params.Set("request_uri", info.RequestURI)
+
+	// TODO: check that 'authorization_endpoint' is "safe" (?)
```
DavidBuchanan314: https-only seems like the right thing to do here, perhaps with an exception for http://localhost

atproto/auth/oauth/store.go
```diff
@@ -0,0 +1,20 @@
...
+// Implementations should allow for concurrent access.
+type ClientAuthStore interface {
+	GetSession(ctx context.Context, did syntax.DID) (*ClientSessionData, error)
+	SaveSession(ctx context.Context, sess ClientSessionData) error
+	DeleteSession(ctx context.Context, did syntax.DID) error
+
+	GetAuthRequestInfo(ctx context.Context, state string) (*AuthRequestData, error)
+	SaveAuthRequestInfo(ctx context.Context, info AuthRequestData) error
+	DeleteAuthRequestInfo(ctx context.Context, state string) error
+}
```
DavidBuchanan314, line 24: This data model assumes you have a single active session per DID, right?

Maybe a BFF backend, for example, would want to have multiple active sessions per DID? (e.g. user signed in on both mobile and desktop) - or would they be required to multiplex multiple "frontend sessions" over a single "backend session"?

atproto/auth/oauth/session.go
```diff
@@ -0,0 +1,361 @@
...
+	// WWW-Authenticate: DPoP error="use_dpop_nonce", error_description="Resource server requires nonce in DPoP proof"
+	return strings.Contains(hdr, "error=\"use_dpop_nonce\"")
+}
+
+// Parses a WWW-Authenticate response header to see if access token has expired (needs refresh)
+func isExpiredAccessTokenHeader(hdr string) bool {
+	// Example from OAuth 2.1 draft:
+	// WWW-Authenticate: Bearer error="invalid_token" error_description="The access token expired"
+	// TODO: should this also look for "expired"?
+	return strings.Contains(hdr, "error=\"invalid_token\"")
```
DavidBuchanan314, line 239: I went down a small rabbit hole to find out whether the double-quotes are actually required and afaict they are, so these strings.Contains checks should be reasonably robust :+1:

atproto/auth/oauth/util.go
```diff
@@ -0,0 +1,24 @@
...
+	"crypto/sha256"
+	"encoding/base64"
+	"math/rand"
+)
+
+// this generates pseudo-unique nonces to prevent token (JWT) replay. these do not need to be cryptographically resilient
+func randomNonce() string {
+	// TODO: make this longer?
+	buf := make([]byte, 16)
+	rand.Read(buf)
```
DavidBuchanan314, line 12: Length looks good to me. Even if you get a collision the only impact is that the server gives you an error.

staticcheck flags `rand.Read` as deprecated recommending `crypto/rand.Read` instead. It is not security-relevant but I'd recommend following it if only to silence the linter.

atproto/auth/oauth/cmd/oauth-web-demo/home.html
```diff
@@ -0,0 +1,3 @@
+{{ define "content" }}
+This is home!
+{{ end }}
```
haileyok, line 3: sickos yes on the template engine (and i am so sorry)

atproto/auth/oauth/cmd/oauth-web-demo/main.go
```diff
@@ -0,0 +1,340 @@
...
+
+func main() {
+	app := cli.App{
+		Name:   "oauth-web-demo",
+		Usage:  "atproto OAuth web server demo",
+		Action: runServer,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:     "session-secret",
+				Usage:    "random string/token used for session cookie security",
```
haileyok, line 31: may be nice to include a tidbit on how to generate one of these e.g. `openssl rand -hex 16`

atproto/auth/oauth/cmd/oauth-web-demo/main.go
```diff
@@ -0,0 +1,340 @@
...
+				EnvVars:  []string{"SESSION_SECRET"},
+			},
+			&cli.StringFlag{
+				Name:    "hostname",
+				Usage:   "public host name for this client (if not localhost dev mode)",
+				EnvVars: []string{"CLIENT_HOSTNAME"},
+			},
+			&cli.StringFlag{
+				Name:    "client-secret-key",
+				Usage:   "confidential client secret key. should be P-256 private key in multibase encoding",
```
haileyok, line 42: do we have helpers, e.g. in goat to make these? again, getting into a "nice to have" thing but would be really nice to make all of this setup as easy as possible for folks

atproto/auth/oauth/doc.go
haileyok: may also add in parentheses something like `(usually a PDS)`

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	// set of scope strings; should not include "atproto"
+	Scopes []string
+
+	UserAgent string
+
+	// For confidential clients, the private client assertion key. Note that while an interface is used here, only P-256 is allowed by the current specification.
+	PrivateKey crypto.PrivateKey
+
+	// ID for current client assertion key (should be provided if PrivateKey is)
+	KeyID *string
```
haileyok, line 49: i'm not the best with idomatic go, but do we really want this to be a pointer? (i've started straying away from doing this for various reasons, including that it's a little bit annoying to create them). could just check `KeyID != ""` when `PrivateKey` is not nil

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	app := &ClientApp{
+		Client:   http.DefaultClient,
+		Resolver: NewResolver(),
+		Dir:      identity.DefaultDirectory(),
+		Config:   config,
+		Store:    store,
+	}
+	if config.UserAgent != "" {
+		app.Resolver.UserAgent = config.UserAgent
+		// TODO: some way to wire UserAgent through to identity directory
```
haileyok: probably worth us getting this going ahead of time, i assume not too difficult (could be like `identity.DefaultDirectory(...opts option.Option)` maybe. feels pretty important here as a lot of calls get made to the directory

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	return jwks
+}
+
+// helper to turn a list of scope strings in to a single space-separated scope string
+func scopeStr(scopes []string) string {
+	if len(scopes) == 0 {
+		return "atproto"
+	}
+	return "atproto " + strings.Join(scopes, " ")
+}
```
haileyok: while we do require `atproto` to be in the scopes, are we sure that we want to be adding it here? i'm fairly sure (though not 100% sure) that we don't do this in the typescript implementation, e.g. we expect the input for instantiating a client to be the exact same as the metadata JSON https://github.com/bluesky-social/atproto/blob/82e3e7bf6bd16ddee6bf684ad10160593b1b0671/packages/oauth/oauth-client-browser/README.md?plain=1#L67

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	// TODO: refactor this in to ClientAuthStore layer?
+	priv, err := crypto.ParsePrivateMultibase(sd.DpopPrivateKeyMultibase)
+	if err != nil {
+		return nil, err
+	}
+	sess.DpopPrivateKey = priv
+	return &sess, nil
+}
+
+type clientAssertionClaims struct {
```
haileyok, line 220: Maybe we just make this (and others like `dpopClaims`) public? Could be useful to devs.

atproto/auth/oauth/types.go
```diff
@@ -0,0 +1,394 @@
...
+	// must include code
+	ResponseTypesSupported []string `json:"response_types_supported"`
+
+	// must include authorization_code and refresh_token (refresh tokens must be supported)
+	GrantTypesSupported []string `json:"grant_types_supported"`
+
+	/// must include S256
+	CodeChallengeMethodsSupported []string `json:"code_challenge_methods_supported"`
+
+	// must include both none (public client s) and private_key_jwt (confidential clients)
```
haileyok:
```suggestion
	// must include both none (public clients) and private_key_jwt (confidential clients)
```

atproto/auth/oauth/types.go
```diff
@@ -0,0 +1,394 @@
...
+	// must include atproto. If supporting the transitional grants, they should be included here as well.
+	ScopesSupported []string `json:"scopes_supported"`
+
+	// must be true
+	AuthorizationReponseISSParameterSupported bool `json:"authorization_response_iss_parameter_supported"`
+
+	// must be true
+	RequirePushedAuthorizationRequests bool `json:"require_pushed_authorization_requests"`
+
+	// correspnds be the PAR endpoint URL
```
haileyok: typo, i think you meant this?

```suggestion
	// correspnds to the PAR endpoint URL
```

atproto/auth/oauth/types.go
```diff
@@ -0,0 +1,394 @@
...
+	// endpoint URL for token requests
+	TokenEndpoint string `json:"token_endpoint"`
+
+	// must include code
+	ResponseTypesSupported []string `json:"response_types_supported"`
+
+	// must include authorization_code and refresh_token (refresh tokens must be supported)
+	GrantTypesSupported []string `json:"grant_types_supported"`
+
+	/// must include S256
```
haileyok:
```suggestion
	// must include S256
```

atproto/auth/oauth/types.go
```diff
@@ -0,0 +1,394 @@
...
+
+	// `code` must be included
+	ResponseTypes []string `json:"response_types"`
+
+	// At least one redirect URI is required.
+	RedirectURIs []string `json:"redirect_uris"`
+
+	// confidential clients must set this to `private_key_jwt`; public must be `none`
+	// TODO: should this be string not *string?
+	TokenEndpointAuthMethod *string `json:"token_endpoint_auth_method,omitempty"`
```
haileyok: seems it can be omitted and we give it a default value

https://github.com/bluesky-social/atproto/blob/82e3e7bf6bd16ddee6bf684ad10160593b1b0671/packages/oauth/oauth-types/src/oauth-client-metadata.ts#L38-L40

atproto/auth/oauth/session.go
```diff
@@ -0,0 +1,361 @@
...
+			// parse the error body to confirm the error type
+			var errResp map[string]any
+			if err := json.NewDecoder(resp.Body).Decode(&errResp); err != nil {
+				slog.Warn("token refresh failed, and could not parse response body", "authServer", tokenURL, "err", err, "statusCode", resp.StatusCode)
+				resp.Body.Close()
+				return "", fmt.Errorf("token refresh failed: HTTP %d", resp.StatusCode)
+			} else if errResp["error"] != "use_dpop_nonce" {
+				slog.Warn("token refresh failed", "authServer", tokenURL, "body", errResp, "statusCode", resp.StatusCode)
+				return "", fmt.Errorf("token refresh failed: %s", errResp["error"])
+			}
```
haileyok: body doesn't get closed here

haileyok: what i've done for this in the past is wrap the `for` loop's code in a `func`, so you can just use a `defer` to handle this and avoid missing a case where you need to close.

atproto/auth/oauth/resolver.go
```diff
@@ -0,0 +1,177 @@
...
+
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	var body ProtectedResourceMetadata
+	if err := json.Unmarshal(respBytes, &body); err != nil {
+		return "", fmt.Errorf("invalid protected resource document: %w", err)
+	}
```
haileyok: can we just use `json.NewDecoder().Decode()` for this?

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+			// parse the error body to confirm the error type
+			var errResp map[string]any
+			if err := json.NewDecoder(resp.Body).Decode(&errResp); err != nil {
+				slog.Warn("PAR request failed, and could not parse response body", "authServer", parURL, "err", err, "statusCode", resp.StatusCode)
+				resp.Body.Close()
+				return nil, fmt.Errorf("token refresh failed: HTTP %d", resp.StatusCode)
+			} else if errResp["error"] != "use_dpop_nonce" {
+				slog.Warn("PAR request failed", "authServer", parURL, "body", errResp, "statusCode", resp.StatusCode)
+				return nil, fmt.Errorf("PAR request failed: %s", errResp["error"])
+			}
```
haileyok: same as other comment, missing `resp.Body.Close()`. could wrap this all in a `func` if you want

haileyok: in fact, since this logic is getting reused can we just extract it to its own function anyway?

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+			// parse the error body to confirm the error type
+			var errResp map[string]any
+			if err := json.NewDecoder(resp.Body).Decode(&errResp); err != nil {
+				slog.Warn("initial token request failed, and could not parse response body", "authServer", authServerMeta.TokenEndpoint, "err", err, "statusCode", resp.StatusCode)
+				resp.Body.Close()
+				return nil, fmt.Errorf("initial token request failed: HTTP %d", resp.StatusCode)
+			} else if errResp["error"] != "use_dpop_nonce" {
+				slog.Warn("initial token request failed", "authServer", authServerMeta.TokenEndpoint, "body", errResp, "statusCode", resp.StatusCode)
+				return nil, fmt.Errorf("initial token request failed: %s", errResp["error"])
+			}
```
haileyok: also missing here

haileyok (commented):
this overall looks really nice! great stuff. left some nits as well as some stuff around the http requests to fix and maybe improve if you want

atproto/auth/oauth/oauth.go
bnewbold: Do you think we should use the resolved handle because it would be more legible/recognizable?

I think we should pass through what the user provided: if they are thinking handle they would use handle. Also things should keep working if handle resolution breaks, and adding extra logic to handle `invalid.handle` feels brittle.

atproto/auth/oauth/oauth.go
bnewbold: Checking for HTTPS makes sense. What I was probably thinking of was trying to do some of the SSRF checks from `indigo:util/ssrf/ssrf.go` here (even if the end user device might see different resolution than the server, if they are distinct).

atproto/auth/oauth/doc.go
bnewbold: At some point I kind of want to try using "host" as a generic term that could encompass "PDS and/or entryway" and "resource server and/or auth server". But here in the OAuth SDK probably best to be more precise.

atproto/auth/oauth/store.go
bnewbold: Yup, that is the assumption here. And now that you mention it, this needs to be revisited.

What I had been thinking was that in a server/BFF situation, there could be multiple client end devices/sessions connected to the server via whatever authentication, but the server would only need one active session with the user's auth server. However, all those end devices/sessions are probably going to authenticate to the BFF using atproto OAuth bouncing to the auth server, which could clobber existing sessions.

Will need to think this through a bit.

atproto/auth/oauth/session.go
bnewbold: I hadn't dug in to confirm this would work, thanks for checking! This still isn't perfect, in that it isn't splitting/parsing out the actual fields, but feels good enough to me.

atproto/auth/oauth/util.go
bnewbold: `crypto/rand` is fine, I made that change in `indigo:atproto/auth` and will be consistent here.

I do wonder if small-CPU high-request-throughput servers (raspberry pi?) could end up exhausting string cryptographic randomness... I don't really know what the state of all that is in various kernels and the go runtime is these days.

atproto/auth/oauth/cmd/oauth-web-demo/main.go
bnewbold: added a README, covering this and the client-secret-key

atproto/auth/oauth/oauth.go
bnewbold: It's a great question, and I have been doing `string` with `""` a nil-like a fair amount, because the pointer ergonomics are so poor. It rubs the wrong way from a correctness standpoint (empty string is different from no string!), but whatever, it's Go.

In this specific case, the `indigo:crypto` library has `KeyID` as a `*string`, so I passed it all through consistently that way. It ends up getting used as part of JWTs, and I think the server needs to remember it.

If it is a huge itch we could maybe update the crypto library. Or, I guess, this part of the library could just be a bit inconsistent with the crypto library (be `string` and convert when constructing the JWK).

atproto/auth/oauth/oauth.go
bnewbold: I can type-unwrap the directory and insert it in to the `BaseDirectory` (as a TODO before merging, will need to test).

atproto/auth/oauth/oauth.go
bnewbold: I kind of like being able to pass `nil` or empty string and have the SDK handle it. But I guess that might be unexpected to devs? Hrm.

atproto/auth/oauth/oauth.go
bnewbold: My thinking was to not expose the JWT library as part of the public API for this package, so we could swap out dependencies in the future if needed. Eg, i'm not 100% confident we won't want to switch to the `jwx` family in the future.

I guess we could make this code public in a sub-package?

atproto/auth/oauth/session.go
bnewbold: mmm yeah I think a func/defer could clean up this error handling a bunch. and could probably actually be a shared helper function between all these auth server request cases.

atproto/auth/oauth/oauth.go
haileyok: that feels like a good reason! wasn't quite sure, but if it's what we use throughout the rest of the repo probably makes sense to just stick with that for this particular case.

atproto/auth/oauth/oauth.go
haileyok: definitely a bit of an implementation detail...at the very least, i think right now this could lead to accidentally adding `atproto` _twice_ to the scopes since - in most usage examples (and looking at other people's client metadata) that exist today - they will see `atproto` and may think they need it too

atproto/auth/oauth/oauth.go
haileyok: i do like the sub-package approach i think...given that these types are really quite large and a bit of a PITA to re-implement everywhere (kinda a minus on go, struct tags are cool but annoying to write...) would be nice to just have confidence they are available wherever

(fwiw though, maybe less important given that people will probably need to write their own struct tags for e.g. gorm or other dbs...... hrm...)

atproto/auth/oauth/oauth.go
DavidBuchanan314: I was thinking an application might want to "remember" a user by DID, and it'd be gross for regular users to see that in the UI upon (re)logging in - but maybe that's not common/realistic as a scenario. Having to think about `invalid.handle` is also a bit of a headache. Maybe the auth server should be responsible for smoothing that UX corner over instead.

atproto/auth/oauth/util.go
DavidBuchanan314: `crypto/rand` draws from `getrandom` on Linux, which uses `urandom` by default. This is cheap/fast especially on modern Linux where it's based on chacha20.

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	token.Header["jwk"] = pubJWK
+	return token.SignedString(privKey)
+}
+
+// Sends PAR request to auth server
+func (app *ClientApp) SendAuthRequest(ctx context.Context, authMeta *AuthServerMetadata, scope, loginHint string) (*AuthRequestData, error) {
+
+	parURL := authMeta.PushedAuthorizationRequestEndpoint
+	state := randomNonce()
+	pkceVerifier := fmt.Sprintf("%s%s%s", randomNonce(), randomNonce(), randomNonce())
```
DavidBuchanan314: ah, didn't notice `randomNonce()` was used here too. This *is* security sensitive so using crypto/rand is essential. (which we do now, so all good)

atproto/auth/oauth/util.go
```diff
@@ -0,0 +1,23 @@
+package oauth
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+)
+
+// this generates pseudo-unique nonces to prevent token (JWT) replay. these do not need to be cryptographically resilient
```
DavidBuchanan314: > these do not need to be cryptographically resilient

This is no longer correct since `randomNonce()` is also used for state+pkce (see my other comment)

atproto/auth/oauth/oauth.go
DavidBuchanan314: Might be worth renaming it to something like `secureRandomBase64(n)` where n is the number of bytes to ask for.

rfc7636 recommends at least 32-bytes (256 bits) worth of entropy for PKCE so we're ok here but would be good to make what we're doing more explicit.

atproto/auth/oauth/oauth.go
```diff
@@ -0,0 +1,651 @@
...
+	switch priv.(type) {
+	case *crypto.PrivateKeyP256:
+		// pass
+	case *crypto.PrivateKeyK256:
+		return fmt.Errorf("only P-256 (ES256) private keys supported for atproto OAuth")
+	default:
+		return fmt.Errorf("unknown private key type: %T", priv)
+	}
+	config.PrivateKey = priv
+	config.KeyID = &keyID
```
DavidBuchanan314, line 125: This impl only supports having one private key at a time, which seems reasonable, so `AddClientSecret` -> `SetClientSecret`?

(There are perhaps use cases for having multiple active *public* keys declared in the client metadata, but adding public keys would have to be a separate API)

atproto/auth/oauth/oauth.go
bnewbold: I added a URL syntax check in AuthServerMetadata.Validate() to check this earlier in the flow.

I'm not sure about allowing localhost redirects at all; it opens up some weird SSRF-y concerns. I think if folks are doing local OAuth dev work, they should probably have a docker-compose or something set up? Though then TLS is tricky. Generally I don't love adding local network carveouts in the default SDK, unless we get clear feedback that it is needed, and even then it should probably be an explicit config thing.

For now i've left a TODO note, I don't think SSRF checks are a "MUST".

atproto/auth/oauth/session.go
bnewbold: added a helper function and much cleaner now!

atproto/auth/oauth/oauth.go
bnewbold: (used same function noted above)

atproto/auth/oauth/oauth.go
bnewbold: (used same function noted above)

atproto/auth/oauth/oauth.go
bnewbold: added this un-wrapping; and added logging if it fails so that we notice this if DefaultDirectory type changes

atproto/auth/oauth/oauth.go
bnewbold: yeah, I guess it is less un-expected/magic to require devs to always inlude "atproto" in the array. updated to work that way.

I'm almost tempted to have it return an error if it isn't included, but that would add a bunch of required error handling.

atproto/auth/oauth/oauth.go
bnewbold: I think i'm going to just leave these private here for now, and will revist when we add server validation. I think folks that want it sooner can just copy this code; I am not confident enough to want to want to maintain these quite yet.

haileyok: sgtm

atproto/auth/oauth/types.go
bnewbold: Yup, but that default value is invalid for an atproto OAuth client; if the default is used, validation fails later.

Created a specs issue: https://github.com/bluesky-social/atproto-website/issues/434

and updated this type to be `string` not `*string`.

atproto/auth/oauth/oauth.go
bnewbold: We might need to support additional public keys in `PublicJWKS`, in addition to the public key corresponding to the configured secret key. But yeah, probably only ever a single secret key at a time, so this rename makes sense (and I updated).

bnewbold:
@DavidBuchanan314 yeah we went around a bit on whether to delete the auth flow info in `GetAuthRequestInfo` itself back when writing/reviewing the python cookbook example. It isn't the auth server which would retry, it is the user's browser.

When developing, it is helpful to be able to retry the callbacks with a simple browser refresh. Once the app is all set, I guess I can see that it could prevent rare scenarios to delete the session in the `Get`? But I feel like this is pretty corner-case, and from the security aspect can be left to the auth server to prevent replays (for example).

atproto/auth/oauth/store.go
bnewbold: Updated to support multiple auth sessions; and added docs about when that is needed, vs a single session per account.

atproto/auth/oauth/oauth.go
bnewbold: haha, forgot I had that `fmt.Sprintf("%s%s%s", randomNonce(), randomNonce(), randomNonce())` in there.

Updated to `secureRandomBase64(n)`.

bnewbold:
Resolved a few more things (see commit names), and I should manually test all the various modes.

There is a list of TODOs at the top of this PR, which I will turn in to issues if they don't get resolved before merging. I don't think they are blocking.

I need to re-test with these changes, and should do another pass over doc comments. But in general I think this is ready for final review.

PR #1102, opened 2025-06-20
repomgr: implement PutRecord

iameli:
Uses the same logic as CreateRecord. Works. Not sure if there's other gotchas for this other than that.

### DIFF
--- a/repomgr/repomgr.go
+++ b/repomgr/repomgr.go
@@ -224,6 +224,72 @@ func (rm *RepoManager) CreateRecord(ctx context.Context, user models.Uid, collec
 	return collection + "/" + tid, cc, nil
 }
 
+func (rm *RepoManager) PutRecord(ctx context.Context, user models.Uid, collection string, rkey string, rec cbg.CBORMarshaler) (string, cid.Cid, error) {
+	ctx, span := otel.Tracer("repoman").Start(ctx, "PutRecord")
+	defer span.End()
+
+	unlock := rm.lockUser(ctx, user)
+	defer unlock()
+
+	rev, err := rm.cs.GetUserRepoRev(ctx, user)
+	if err != nil {
+		return "", cid.Undef, err
+	}
+
+	ds, err := rm.cs.NewDeltaSession(ctx, user, &rev)
+	if err != nil {
+		return "", cid.Undef, err
+	}
+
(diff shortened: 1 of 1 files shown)


bnewbold:
Could you describe your use case for this package? We have not been actively maintaining it; i'd consider it soft-deprecated right now.

iameli:
> Could you describe your use case for this package? We have not been actively maintaining it; i'd consider it soft-deprecated right now.

Certainly. The big picture thing I'm looking to do is teach Streamplace how to self-host its `com.atproto.lexicon.schema` records at `did:web:stream.place`. It'll be much cleaner and easier to have that as a "static" PDS repo that hosts the records rather than needing to remember to push them to an external repository on every change. (I also have other fun ideas for "Streamplace as a PDS", so it seemed like a good investment to build the foundation.) So I'm implementing whatever subset of the PDS features I need to make that happen; so far that's the following XRPC queries:

```
com.atproto.repo.describeRepo
com.atproto.repo.getRecord
com.atproto.repo.listRecords
com.atproto.server.describeServer
com.atproto.sync.getRecord
com.atproto.sync.listRepos
```

`repomgr` came in when I wanted to implement `com.atproto.sync.getRecord`; `repomgr`'s `GetRecordProof` seemed like the existing way to do that in the codebase. So I used `repomgr` as the foundation for Streamplace's built-in.

If it's not actively maintained that's totally fine, but are the underlying `repo` and `carstore` packages maintained? Like I'm happy just to do all of these things myself, though it seems like potentially something that'd be nice to have in the `pds` package:

https://github.com/bluesky-social/indigo/blob/336ebe49427bb069072a819f664ef92b6cf0463e/repomgr/repomgr.go#L467-L491

iameli:
> The direction we are moving in is the `atproto/repo` package. Hailey used that code/interface to implement her Go PDS, which you might want to check out: https://github.com/haileyok/cocoon
> 
> The other thing you would need is `subscribeRepos` and the repo event stream (aka, PDS firehose). Without that, downstream services will not discover or update any records posted in repos you host. You'd be sort-atproto-ish, but in my opinion without a firehose, data isn't really in the atproto network.

Right on, yeah, `subscribeRepos` was definitely next on my list. I'll switch over to using `repo` and `carstore` directly :+1:

PR #1103, opened 2025-06-20
carstore: don't ignore the first record in sqlite

iameli:
Small fix to the SQLite carstore -- the first record in the cache will have ID 0, there's no reason to ignore it. Naturally, this case happens a lot in testing and stuff 'cause the first attempt to get a revision on an empty database will fail

### DIFF
--- a/carstore/sqlite_store.go
+++ b/carstore/sqlite_store.go
@@ -6,11 +6,12 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"go.opentelemetry.io/otel/attribute"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
+
+	"go.opentelemetry.io/otel/attribute"
 
 	"github.com/bluesky-social/indigo/models"
 	blockformat "github.com/ipfs/go-block-format"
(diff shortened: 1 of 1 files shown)


iameli:
Force-pushed to ignore the same check in a different spot

haileyok:
bumping to @whyrusleeping

iameli:
Rebased, and added another bugfix. Previously we were calling `blocks.NewBlock` when retrieving blocks from the database, which causes the returned block to have a CID v0 and break for stuff like `com.atproto.sync.getRecord`: https://github.com/ipfs/boxo/blob/v0.7.0/blocks/blocks.go#L37

I assume the Scylla implementation has the same problem but I haven't booted that up locally so I didn't want to blindly touch that code: https://github.com/bluesky-social/indigo/blob/8be102876fb7e638aa4c9ed6c9d4991ca19a0973/carstore/scylla.go#L494

bnewbold:
As a heads up, this `carstore` code is a scary forest to most of us except for Why (the original author), and it will probably be a while before he can look at this. We do have some production services using the `carstore` code, but it isn't as actively maintained.

whyrusleeping:
Oh people should really ping me on bluesky or slack when theres PRs i need to look at, my github notifications are a hellish quagmire

carstore/sqlite_store.go
```diff
@@ -215,9 +216,6 @@ func (sqs *SQLiteStore) GetUserRepoHead(ctx context.Context, user models.Uid) (c
 	if lastShard == nil {
 		return cid.Undef, nil
 	}
-	if lastShard.ID == 0 {
```
whyrusleeping, line 218: I see how this is unnecessary, but also I feel like its cleaner to be explicit here. Was this causing problems?

iameli: Yeah, the first shard has ID == 0 — it's the first record in the database. So it was getting ignored

PR #1118, opened 2025-07-14
Make MockDirectory Thread Safe

jcalabro:
I've wanted this when writing tests that use the mock directory from multiple goroutines.


bnewbold (approved):
thanks!

PR #1120, opened 2025-07-16
Goat PLC identity management commands

DavidBuchanan314:
Adds the following new subcommands to `goat plc`, for PLC identity management without the involvement of a PDS:

```
   genesis   produce an unsigned genesis operation
   calc-did  calculate the DID corresponding to a signed PLC operation
   sign      sign an operation, ready to be submitted
   submit    submit a signed operation to the PLC directory
   update    apply updates to a previous operation produce a new one (but don't sign or submit it, yet)
```


cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+	return nil
+}
+
+// fetch logs from /log/audit, select according to base_cid ("" means use latest), and
+// prepare it for updates:
+//   - convert from legacy op format if needed (and reject tombstone ops)
+//   - strip signature
+//   - set `prev` to appropriate value
+func fetchOpForUpdate(ctx context.Context, c didplc.Client, did string, base_cid string) (*didplc.RegularOp, error) {
+	auditlog, err := c.OpLog(ctx, did, true) // NB: this API changes in a pending go-didplc PR
```
DavidBuchanan314: Will get the `go-didplc` changes merged before landing this PR, maybe a v0.0.1 tag too

DavidBuchanan314: (did not do a v0.0.1 in the end)

DavidBuchanan314:
cc @bnewbold (I don't have the "request review" button for whatever reason)

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/bluesky-social/indigo
 
-go 1.24
+go 1.24.0
```
bnewbold: this seems to have pulled in a lot more dependency updates, and tweaks to these toolchain versions, than I would have expected from just adding `go-didplc`. did you do a general update or some command other than `go mod tidy`?

If possible I think it would be better to just import `go-didplc` and not touch anything else. if we've touched a lot of dependencies, we'll want to check with other backend folks about this impacting deploys of other services (like the dataplane).

cmd/goat/plc.go
```diff
@@ -71,6 +77,97 @@ var cmdPLC = &cli.Command{
...
+			Flags:     []cli.Flag{},
+			Action:    runPLCCalcDID,
+		},
+		&cli.Command{
+			Name:      "sign",
+			Usage:     "sign an operation, ready to be submitted",
+			ArgsUsage: `<operation.json>`,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:    "plc-private-rotation-key",
```
bnewbold: maybe this could be a bit shorted as `plc-signing-key` / `PLC_SIGNING_KEY`? I could go either way

cmd/goat/plc.go
```diff
@@ -71,6 +77,97 @@ var cmdPLC = &cli.Command{
...
+					Usage:   "private key used as a rotation key, if operation is not signed (multibase syntax)",
```
bnewbold: I think the description is out of date, should probably be something like "private key used to sign operation (multibase syntax)"

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+	if atprotoKey != "" {
+		if _, err := crypto.ParsePublicDIDKey(atprotoKey); err != nil {
+			return err
+		}
+		op.VerificationMethods["atproto"] = atprotoKey
+	}
+
+	pds := cctx.String("pds")
+	if pds != "" {
+		// TODO: check pds is valid URI?
```
bnewbold: parsing with `net/url` and confirming the scheme is https is probably a good start? IIRC `net/url` parsing works with relative URLs as well as full URLs; we'd want to check that it is the later

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+	}
+	op := enum.AsOperation()
+
+	if !op.IsGenesis() {
+		return fmt.Errorf("not a genesis op")
+	}
+
+	if !op.IsSigned() {
+		return fmt.Errorf("genesis op must be signed")
+	}
```
bnewbold: ok with me to double-check here, but does `go-didplc` also now do these checks when calling `op.DID()`?

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+
+	didplc, err := op.DID()
```
bnewbold: this is quietly clobbering the `go-didplc` package import name. this should be avoided even if it isn't causing a problem here (and some form of linting should probably catch this).

I'd just use `did`

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+	c := didplc.Client{
+		DirectoryURL: cctx.String("plc-host"),
+		UserAgent:    GOAT_PLC_USER_AGENT,
+	}
+
+	if err = c.Submit(ctx, didString, op); err != nil {
+		return err
+	}
+
+	// TODO: print confirmation?
```
bnewbold: yeah, i'd probably print the string `success` (but flexible on this)

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+
+	return nil
+}
+
+// fetch logs from /log/audit, select according to base_cid ("" means use latest), and
+// prepare it for updates:
+//   - convert from legacy op format if needed (and reject tombstone ops)
```
bnewbold, line 629: oh nice I didn't even think of that part!

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+			return nil, fmt.Errorf("no operation found matching CID %s", base_cid)
+		}
+	}
+	var op didplc.RegularOp
+	switch baseOp := baseLogEntry.Operation.AsOperation().(type) {
+	case *didplc.RegularOp:
+		op = *baseOp
+		op.Sig = nil
+	case *didplc.LegacyOp:
+		op = baseOp.RegularOp()
```
bnewbold: do we not need to strip `op.Sig = nil` on this conversion path? I'd add a comment if we don't

cmd/goat/plc.go
```diff
@@ -9,13 +9,19 @@ import (
...
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/bluesky-social/indigo/util"
 
+	"github.com/did-method-plc/go-didplc"
+
 	"github.com/urfave/cli/v2"
 )
 
+// TODO: use this consistently for all requests
+const GOAT_PLC_USER_AGENT = "goat-cli"
```
bnewbold: there is a helper `userAgent()` method in this goat `main` package, i'd use that inline instead of defining this constant

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+		if _, err := crypto.ParsePublicDIDKey(rotationKey); err != nil {
+			return err
+		}
+		// prepend (Note: if adding multiple rotation keys at once, they'll end up in reverse order)
+		op.RotationKeys = append([]string{rotationKey}, op.RotationKeys...)
+	}
+
+	handle := cctx.String("handle")
+	if handle != "" {
+		parsedHandle, err := syntax.ParseHandle(strings.TrimPrefix(handle, "at://"))
```
bnewbold, line 716: out of habit i'd consider adding:

```
parsedHandle = parsedHandle.Normalize()
```

that basically lower-cases it. does the handle *need* to be lower-cased in the DID doc? I suppose not.

cmd/goat/plc.go
```diff
@@ -320,3 +417,344 @@ func fetchPLCData(ctx context.Context, plcHost string, did syntax.DID) (*PLCData
...
+	case *didplc.LegacyOp:
+		op = baseOp.RegularOp()
+	case *didplc.TombstoneOp:
+		return nil, fmt.Errorf("cannot update from a tombstone op")
+	}
+	op.Prev = &baseLogEntry.CID
+	return &op, nil
+}
+
+func runPLCUpdate(cctx *cli.Context) error {
```
bnewbold, line 672: this function is right at the edge of needing test coverage, especially for the likely less-frequently used stuff.

probably the best way to handle that will be to add helpers in `go-didplc` for doing these atproto-specific operations; `op.SetHandle(hdl)`, `op.AddRotationKey(key)`, etc.

I don't think we need to tackle that in this PR (let's get this shipped).

bnewbold (approved):
left notes, but overall looks good. excited to get this functionality out!

I left more review notes than I maybe would normally, as style feedback

go.mod
DavidBuchanan314: oh, yes there was something broken and I had to run a `go mod tidy` before it would let me `go get`. I forget what the specific issue was but iirc it was just one package, I will see if I can roll that back

cmd/goat/plc.go
DavidBuchanan314: Ah yeah good catch, that was a copy-paste from plcli

cmd/goat/plc.go
DavidBuchanan314: It does one of the checks but not the other. Now you mention it it's probably best to fix that up in go-didplc first, then both checks can be elided here.

cmd/goat/plc.go
DavidBuchanan314: We do not (it's handled as part of the conversion), I'll add the comment

cmd/goat/plc.go
DavidBuchanan314: Ah I assumed it would do that automatically for some reason, good call.

go.mod
DavidBuchanan314:
```
go: github.com/ipfs/go-bs-sqlite3@v0.0.0-20221122195556-bfcee1be620d: invalid version: unknown revision bfcee1be620d
```

Will see if I can manually splice the updated version into go.mod/go.sum without touching anything else (is that a terrible idea?)

DavidBuchanan314: manually splicing was indeed a terrible idea - Is there a good path forward without updating a bunch of deps?

cmd/goat/plc.go
```diff
@@ -494,20 +494,12 @@ func runPLCCalcDID(cctx *cli.Context) error {
...
-	if !op.IsGenesis() {
-		return fmt.Errorf("not a genesis op")
-	}
-
-	if !op.IsSigned() {
-		return fmt.Errorf("genesis op must be signed")
-	}
-
-	didplc, err := op.DID()
+	did, err := op.DID() // errors if op is not a signed genesis op
```
DavidBuchanan314, line 502: NOTE: go-didplc needs to be updated before this is valid, pending merge of https://github.com/did-method-plc/go-didplc/pull/9

go.mod
bnewbold: we hacked this in https://github.com/bluesky-social/indigo/pull/1121

PR #1122, opened 2025-07-17
goat: Create service auth token via locally-held signing key

DavidBuchanan314:
This is needed as part of the "adversarial PDS migration" flow, because the existing `goat account service-auth` command requires the origin PDS to create the token.

```
NAME:
   goat account service-auth-offline - create service auth token via locally-held signing key

USAGE:
   goat account service-auth-offline [command options] [arguments...]

OPTIONS:
   --atproto-signing-key value    private key used to sign the token (multibase syntax) [$ATPROTO_SIGNING_KEY]
   --iss value                    the DID of the account issuing the token
   --endpoint value, --lxm value  restrict token to API endpoint (NSID, optional)
   --audience value, --aud value  DID of service that will receive and validate token
   --duration-sec value           validity time window of token (seconds) (default: 0)
   --help, -h                     show help
```

### DIFF
--- a/cmd/goat/account.go
+++ b/cmd/goat/account.go
@@ -416,6 +416,7 @@ func runAccountServiceAuthOffline(cctx *cli.Context) error {
 	}
 
 	issString := cctx.String("iss")
+	// TODO: support fragment identifiers
 	iss, err := syntax.ParseDID(issString)
 	if err != nil {
 		return fmt.Errorf("iss argument must be a valid DID: %w", err)


cmd/goat/account.go
```diff
@@ -110,6 +112,40 @@ var cmdAccount = &cli.Command{
...
+				},
+				&cli.StringFlag{
+					Name:     "audience",
+					Aliases:  []string{"aud"},
+					Required: true,
+					Usage:    "DID of service that will receive and validate token",
+				},
+				&cli.IntFlag{
+					Name:  "duration-sec",
+					Value: 60,
```
bnewbold, line 143: My intuition is to bump this up to something between 120 to 300 seconds for folks doing this manually/locally? Though I guess this is basically the same with `service-auth` (fetched from PDS, but output locally) and consistency is good. And too-long a duration might be rejected by the audience. I'd probably bump to 120 but also fine leaving as-is, up to you.

cmd/goat/account.go
```diff
@@ -110,6 +112,40 @@ var cmdAccount = &cli.Command{
 			},
 			Action: runAccountServiceAuth,
 		},
+		&cli.Command{
+			Name:  "service-auth-local",
```
bnewbold: I don't love this name, but also don't have a great alternative without renaming `service-auth` (which I don't want to do now b/c we'd need to update existing docs).

`service-auth-offline`? fine leaving as `service-auth-local` as well.

Agree that here in `account` is a bit weird but best right now.

cmd/goat/account.go
```diff
@@ -369,6 +405,52 @@ func runAccountServiceAuth(cctx *cli.Context) error {
...
+	if lxmString != "" {
+		lxmTmp, err := syntax.ParseNSID(lxmString)
+		if err != nil {
+			return fmt.Errorf("lxm argument must be a valid NSID: %w", err)
+		}
+		lxm = &lxmTmp
+	}
+
+	aud := cctx.String("audience")
+	// TODO: can aud DID have a fragment?
```
bnewbold, line 436: Good question; it kind of feels like it might, but the spec doesn't mention and I don't remember off the top of my head.

On the other hand, the `iss` *does* support a fragement/reference: https://atproto.com/specs/xrpc#inter-service-authentication-jwt

We should add parsing support to the syntax package for "DID and optional fragment" and "NSID or NSID plus fragment or just fragment" for this and similar use cases.

I don't think we need to tackle that right now though. I would just add a TODO above on "iss".

cmd/goat/account.go
```diff
@@ -369,6 +405,52 @@ func runAccountServiceAuth(cctx *cli.Context) error {
...
+	_, err = syntax.ParseDID(aud)
+	if err != nil {
+		return fmt.Errorf("aud argument must be a valid DID: %w", err)
+	}
+
+	durSec := cctx.Int("duration-sec")
+	duration := time.Duration(durSec * int(time.Second))
+
+	token, err := auth.SignServiceAuth(iss, aud, duration, lxm, privkey)
```
bnewbold, line 445: yay, glad you found this and that it can be used here!

cmd/goat/account.go
DavidBuchanan314: fwiw that comment was copy-pasted from the original `service-auth` codepath, so I will add the `iss` TODO there too

cmd/goat/account.go
DavidBuchanan314: Do you think it's likely for a too-long duration to be rejected by the audience outright, or just treated as if it were a shorter one?

I think I'll leave this as-is since it's nice to have a "safer" default (and enforces the convention that service auth tokens are intended to be short-lived in general), but in my writeup I set the duration to 3600

cmd/goat/account.go
DavidBuchanan314: Ah, there's no `iss` in the `service-auth` codepath anyway, but added in the TODO for this one

cmd/goat/account.go
DavidBuchanan314: I did the rename to `service-auth-offline`

PR #1125, opened 2025-07-29
atproto/label: fix missing `neg` in lexicon functions

iameli:
Previously we forgot about the `neg` field in between these conversions. Now we don't. With tests!


bnewbold (approved):
oof! thanks

PR #1126, opened 2025-07-30
cmd/goat: concurrent blob export

geesawra:
This commit adds concurrency to the blob export function, downloading each blob listing in an errgroup.

Errors encountered during the download phase are fatal: added a bit of error message to point out blob export may be incomplete.


geesawra:
One thing this PR doesn't take care of is handling PDS blob download rate-limits. 

Does such thing exists?

bnewbold:
NOTE: we are moving development of goat to a new repo: https://github.com/bluesky-social/goat

PR #1129, opened 2025-08-07
feat(cmd/goat): Prompt to request a PLC token and enter the token if orginal PLC token fails

fatfingers23:
Hey, I am making this PR in relation to bluesky-social/goat#8 .

This should hopefully resolve issues for bigger accounts have with using goat to migrate their account to a new PDS. I've moved PDSes twice now, and each time the PLC token has timed out due to the lengthy upload time of blobs. This gives the user an option to request a new PLC token and enter it, so they do not have to restart or manually finish the migration.

- Added a prompt where, if the PLC sign operation fails, it asks the user if they would like to request a new token
- If they choose yes, it calls their old PDS to send a new PLC token
- Prompts for input for the PLC token and tries again

I also took the suggestion from David in the hopes that users who find [bnewbold's blog post](https://whtwnd.com/bnewbold.net/3l5ii332pf32u) (or others) that use goat for migrations will not have any issues and will be prompted to request a new PLC token if their token times out without changing any of the command inputs.

I wasn't sure about the verbiage I used in the prompts, but please let me know if there are any changes needed, and I will gladly make them.


cmd/goat/account_migrate.go
```diff
@@ -5,16 +5,15 @@ import (
...
 	"github.com/bluesky-social/indigo/api/agnostic"
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/bluesky-social/indigo/xrpc"
-
+	"github.com/manifoldco/promptui"
 	"github.com/urfave/cli/v2"
+	"log/slog"
+	"strings"
+	"time"
```
bnewbold, line 16: as a general style we try keep stdlib imports in the first/top set of imports

cmd/goat/account_migrate.go
```diff
@@ -235,7 +234,37 @@ func runAccountMigrate(cctx *cli.Context) error {
...
-		return fmt.Errorf("failed requesting PLC operation signature: %w", err)
+		slog.Warn("failed signing PLC operation", "err", err)
+		confirmPrompt := promptui.Prompt{
+			Label:     "There was an error signing the PLC operation. Do you want to request a new token",
+			IsConfirm: true,
+		}
+
+		_, err = confirmPrompt.Run()
+		if err != nil {
+			return fmt.Errorf("the PLC operation has been cancelled")
```
bnewbold, line 245: this should return the `err` from the API request itself

cmd/goat/account_migrate.go
```diff
@@ -235,7 +234,37 @@ func runAccountMigrate(cctx *cli.Context) error {
 
 	signedPlcOpResp, err := agnostic.IdentitySignPlcOperation(ctx, oldClient, &unsignedOp)
 	if err != nil {
-		return fmt.Errorf("failed requesting PLC operation signature: %w", err)
+		slog.Warn("failed signing PLC operation", "err", err)
+		confirmPrompt := promptui.Prompt{
```
bnewbold, line 238: this should be conditional on a check of whether the process is connected to a live terminal or not

cmd/goat/account_migrate.go
```diff
@@ -235,7 +234,37 @@ func runAccountMigrate(cctx *cli.Context) error {
...
+			Label:     "There was an error signing the PLC operation. Do you want to request a new token",
```
bnewbold, line 239: needs a trailing `?` I think

cmd/goat/account_migrate.go
```diff
@@ -5,16 +5,15 @@ import (
...
-	"log/slog"
-	"strings"
-	"time"
-
 	"github.com/bluesky-social/indigo/api/agnostic"
 	comatproto "github.com/bluesky-social/indigo/api/atproto"
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/bluesky-social/indigo/xrpc"
-
+	"github.com/manifoldco/promptui"
```
bnewbold, line 12: do you think we really need this new dependency, or could we just use the `bufio.NewReader(os.Stdin)` pattern?

fatfingers23: Thank you for the feedback! I agree with everything else. I had used this dependency since I saw it used in another cli, and gave some nice quality of life things (auto y/N, styling, one liner, etc), and I am not as familiar with Go's apis. But I will take some time this week to give it a try with `bufio.NewReader(os.Stdin)` and see what that looks like.

bnewbold:
NOTE: we are moving development of `goat` to a new repo: https://github.com/bluesky-social/goat

PR #1131, opened 2025-08-08
add missing valid rkey char

uniphil:
by [the rkey spec](https://atproto.com/specs/record-key#record-key-syntax):

> restricted to a subset of ASCII characters — the allowed characters are alphanumeric (A-Za-z0-9), period, dash, underscore, colon, or tilde (.-_:~)

tilde was missing from this check, which leads to at least [jetstream becoming unhappy](https://github.com/bluesky-social/jetstream/issues/54) over valid-but-uncommon rkeys.

this change aligns the check with the spec.

### DIFF
--- a/mst/mst_util.go
+++ b/mst/mst_util.go
@@ -197,7 +197,7 @@ func cidForEntries(ctx context.Context, entries []nodeEntry, cst cbor.IpldStore)
 }
 
 // keyHasAllValidChars reports whether s matches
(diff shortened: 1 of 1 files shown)


uniphil:
a few hits come up for other repos that may have copied this bug:

https://github.com/search?q=%20%22%5Ba-zA-Z0-9_%3A.-%5D%22%20atproto&type=code

devinivy (approved):

Issue #1128, opened 2025-08-06
goat: improve `goat account plc submit` error messages


DavidBuchanan314:
The PDS (and PLC directory itself) gives fairly opaque errors if the operation is malformed in some way.

We should be able to use go-didplc to validate the syntax before submission, like `goat plc submit` already does (along with checking that the op is signed)

PR #1132, opened 2025-08-12
goat: improve PLC-related error messages

DavidBuchanan314:
Addresses #1128 and also adds some "hints" to error messages regarding when to use `goat plc` vs `goat account plc`. There's perhaps still room for improvement on the command help texts though.

Note: I haven't fully tested a PLC update via `goat account plc submit` with these changes, yet


bnewbold (approved):
nice

PR #1142, opened 2025-08-28
fix nil pointer deref for ErrAccountRepoNotFound

uniphil:
this change just makes sure `repo` is not nil before we try to pull a `Rev` out of it.

assigning the Rev from the repo follows what's *almost* a boilerplate golang error handler, but is sneakily not handling `relay.ErrAccountRepoNotFound` -- seems the intent is to provide a happy-path response for a common error case.

unfortunately the Rev assignment tried to dereference that from `repo` unconditionally, so it panicked in the allowed-error path.

(still need to test and will verify on my relays)

### DIFF
--- a/cmd/relay/handlers.go
+++ b/cmd/relay/handlers.go
@@ -197,8 +197,11 @@ func (s *Service) handleComAtprotoSyncGetRepoStatus(c echo.Context, did syntax.D
 	if err != nil && !errors.Is(err, relay.ErrAccountRepoNotFound) {
 		return nil, err
 	}
+	// ^^ only returns for non-ErrAccountRepoNotFound: repo can be nil after this!
 
-	out.Rev = &repo.Rev
+	if repo != nil {
+		out.Rev = &repo.Rev
+	}
 
 	return out, nil
 }


bnewbold (approved):
thanks! sloppy on my part, though feels like static analysis could have caught this one :thinking: 

I checked and the `rev` field is optional under the `getRepoStatus` lexicon; the text says:

> Optional field, the current rev of the repo, if active=true

So in theory if active=false we maybe shouldn't return the rev. I'm not sure if we should update the schema/description to have have that behavior, or update the code to not return the rev in that case (could also skip looking up repo status if active=false).

`listRepos` is different because `rev` is required in the response. that should probably be changed in the spec... it breaks the lexicon evolution rules but is the kind of change I think we should still consider as the overall protocol is not 100% baked/stabilized yet.

Regardless, I think this PR is a strict improvement on the current situation so going to merge.

bnewbold:
as a note, hands are full right now and it will probably be a while before this gets deployed to our relay instances.

created an issue with follow-ups: https://github.com/bluesky-social/indigo/issues/1143

uniphil:
cool!

note that in this case, the affected account _is_ active, according to the PDS at least

https://matsutake.us-west.host.bsky.network/xrpc/com.atproto.sync.getRepoStatus?did=did:plc:r47ixofywwpyonnqd5d4pmaz

Issue #1143, opened 2025-08-28
relay: follow-up on missing-rev accounts


bnewbold:
See: https://github.com/bluesky-social/indigo/pull/1142

- [ ] confirm understanding of how all this happened (eg, with DID `did:plc:r47ixofywwpyonnqd5d4pmaz`)
- [ ] either update schema description to clarify that `rev` might be undefined even if `active` is true; or update relay implementation to treat that condition as "repo not found"
- [ ] either don't return `rev` if `active` is false, or update schema description to clarify that it might be returned in that situation (less important?)
- [ ] update `listRepos` to only return accounts with a rev defined (matching the current schema), instead of erroring during pagination
- [ ] consider updating the `listRepos` response schema to make `rev` optional (this would break schema evolution rules, and not obvious this is the right approach)

uniphil:
wait unconfirmed but i think this might be really simple?

- the `account_repo` extension table [was added 4 months ago](https://github.com/bluesky-social/indigo/commit/eaa0f9e3a5a59b74393d50542a96105bc3bd54cb#diff-73c12ff5798f21ddb85081ffae5f58ad268dee51f2c54355de9b0c429d83acfaR82-R89)
- the table gets upserted when processing commits
- accounts without any activity since >4months have not entry?

both accounts i repro'd with were >=4mo since last activity

(and possibly: new accounts are also missing an entry until they emit a commit?)

uniphil:
this actually feels not unrelated to the current connection bootstrapping problem for third party relays, where they have to observe an event come through the firehose from a quiet account in order to sync, and may drop that connection and never try to reconnect if not.

uniphil:
re

> consider updating the listRepos response schema to make rev optional (this would break schema evolution rules, and not obvious this is the right approach)

i'm running into this in two ways now with the `hubble-sync` sync1.1 library:

- `rev` might not be known if we've no successfully synchronized the repo yet. in `relay` we empty-stringed it [for a bit](https://github.com/bluesky-social/indigo/commit/0e2f2596dafda8e3206ce937d51dace1af7917c7), until [this refactor](https://github.com/bluesky-social/indigo/commit/aff436701d9f05dd9a36854a457a49eafbdcc267) (cc @dholms) filtered them out entirely (inner join on `account_repo`). i can match this behaviour in `hubble-sync`, but it feels a bit unfortunate to lose.

[...]

PR #1144, opened 2025-08-28
prevent HTTP 500 responses for listRepos when one account has no account_repo

uniphil:
Accounts sometimes don't have an entry in their account_repo extension table. This change handles that case for listRepos by including them in the response but with empty strings for the "Rev" and "Head" fields.

...which probably isn't ultimately correct, but is at least more usable for clients than missing an entire page of repos because one of them was affected by this.

placeholder improvement directly related to https://github.com/bluesky-social/indigo/issues/1143

basically the same as #1142 but that PR only addressed getRepoStatus

---

for what it's worth, the sql join `TODO` suggested in the code here would have hidden (fixed?) this problem by omitting the affected accounts, assuming inner-join


bnewbold:
this finally got deployed to the relay1 instances just now

PR #1146, opened 2025-08-30
oauth demo: log if returned scopes don't match

bnewbold:


atproto/auth/oauth/cmd/oauth-web-demo/main.go
```diff
@@ -78,7 +79,7 @@ var tmplPost = template.Must(template.Must(template.New("post.html").Parse(tmplB
 
 func runServer(cctx *cli.Context) error {
 
-	scopes := []string{"atproto", "transition:generic"}
+	scopes := []string{"atproto", "account:email"}
```
DavidBuchanan314: Probably want to keep the old scope here too? Or better, `repo:app.bsky.feed.post?action=create` (since that's what the rest of the demo uses)

bnewbold: took me a minute to remember why I even had `account:email` in there (it is to test not granting that permission during auth flow).

I don't think we need `transition:generic`, so just added the record-creation-specific one.

PR #1147, opened 2025-08-30
Client assertion jwt missing exp ( expiresat ) claim

wdantuma:
According to [rfc7523](https://datatracker.ietf.org/doc/html/rfc7523#page-6) a client assertion jwt MUST have an exp ( expiresat ) claim. This could also be a issue with the assertion verification in the reference PDS code ( seems to accepts assertions without the exp claim)  and [python-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/python-oauth-web-app) not sending the exp claim either

### DIFF
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -247,7 +247,7 @@ func (cfg *ClientConfig) NewClientAssertion(authURL string) (string, error) {
 			Audience:  []string{authURL},
 			ID:        secureRandomBase64(16),
 			IssuedAt:  jwt.NewNumericDate(time.Now()),
-			ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Minute * 5)),
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Minute * 1)),
 		},
 	}
 
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -242,11 +242,12 @@ func (cfg *ClientConfig) NewClientAssertion(authURL string) (string, error) {
 	}
 	claims := clientAssertionClaims{
 		RegisteredClaims: jwt.RegisteredClaims{
-			Issuer:   cfg.ClientID,
-			Subject:  cfg.ClientID,
-			Audience: []string{authURL},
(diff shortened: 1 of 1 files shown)


DavidBuchanan314:
Good catch, and thank you for the PR(s)

I just checked our TypeScript client implementation too, and we're setting `exp` there correctly: https://github.com/bluesky-social/atproto/blob/a9ea0d5629c43cb8c563a692891170c6d270f337/packages/oauth/oauth-client/src/oauth-client-auth.ts#L147

5 minute expiry seems totally reasonable, but for consistency with the TS impl I think 1 minute would be even better.

I'll take a look at the PDS-side too, ideally we should tighten that up, but the client changes should probably roll out first.

atproto/auth/oauth/oauth.go
```diff
@@ -242,11 +242,12 @@ func (cfg *ClientConfig) NewClientAssertion(authURL string) (string, error) {
...
-			Subject:  cfg.ClientID,
-			Audience: []string{authURL},
-			ID:       secureRandomBase64(16),
-			IssuedAt: jwt.NewNumericDate(time.Now()),
+			Issuer:    cfg.ClientID,
+			Subject:   cfg.ClientID,
+			Audience:  []string{authURL},
+			ID:        secureRandomBase64(16),
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Minute * 1)),
```
bnewbold: The duration here should definitely be a defined constant, not in-line.

There is an existing constant, `jwtExpirationDuration`, which is set to 30 seconds. If really necessary we could define a second constant, but I would just re-use that. I think the same security logic applies for both (aka, if we need more than 30 seconds for the client attestation, we probably do for request request tokens as well).

wdantuma: Your're right, i use the same const now

wdantuma: @bnewbold Is it possible to merge this one and https://github.com/bluesky-social/cookbook/pull/33 ?

wdantuma: thank you @DavidBuchanan314

PR #1150, opened 2025-09-02
basic permission-set support in lexicon package

bnewbold:
I should probably add parsing helpers to the atproto/syntax package for MIME type glob patterns, and for "service refs".

This isn't super rigorous yet, and it doesn't support pass-through of extra fields on `permission` types. But it would catch early/initial syntax issues, and could be used with `goat` for publishing and resolving lexicons (eg, basic devex for permission-sets feature).


wdantuma:
catalog.LoadDirectory gives an error after this merge ( expected 'params' ) i load the standard lexicon set (com/atproto/* ) from the atproto repo

Issue #1152, opened 2025-09-02
OAuth client SDK: Improve callback URL error handling


DavidBuchanan314:
If the user cancels the authorization (or something else goes wrong), the AS may pass an error message back to the client in the callback URL, in the `error` and (optionally) `error_description` paramaters, per https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

Right now `ProcessCallback()` will generate a nondescript `fmt.Errorf("missing required query param")`

PR #1155, opened 2025-09-08
OAuth client: Surface callback error responses

DavidBuchanan314:
Addresses #1152

This can be tested by initiating a login with the demo client app, but hitting "cancel" on the AS login page (or by refusing to grant the requested scopes).

### DIFF
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -21,6 +21,26 @@ import (
 
 var jwtExpirationDuration = 30 * time.Second
 
+// Returned by [ClientApp.ProcessCallback] if the AS signals an error in the redirect URL parameters, per rfc6749 section 4.1.2.1
+//
+// NOTE: This is untrusted data and should not be e.g. rendered to HTML without appropriate escaping
+type CallbackError struct {
+	code        string
+	description string
+	uri         *syntax.URI
+}
+
+func (e *CallbackError) Error() string {
+	res := "callbackError: " + e.code
+	if e.description != "" {
+		res += ": " + e.description
+	}
+	if e.uri != nil {
+		res += " (" + e.uri.String() + ")"
+	}
+	return res
+}
+
 // Service-level client. Used to establish and refrsh OAuth sessions, but is not itself account or session specific, and can not be used directly to make API calls on behalf of a user.
 type ClientApp struct {
 	Client   *http.Client
@@ -579,6 +599,20 @@ func (app *ClientApp) StartAuthFlow(ctx context.Context, identifier string) (str
 
 // High-level helper for completing auth flow: verifies callback query parameters against persisted auth request info, makes initial token request to the auth server, validates account identifier, and persists session data.
 func (app *ClientApp) ProcessCallback(ctx context.Context, params url.Values) (*ClientSessionData, error) {
+
+	errorCode := params.Get("error")
+	if errorCode != "" {
+		var errorUri *syntax.URI
+		parsedUri, err := syntax.ParseURI(params.Get("error_uri"))
+		if err == nil {
+			errorUri = &parsedUri
+		}
+		return nil, &CallbackError{
+			code:        errorCode,
+			description: params.Get("error_description"),
+			uri:         errorUri,
+		}
+	}
 
 	state := params.Get("state")
 	authserverURL := params.Get("iss")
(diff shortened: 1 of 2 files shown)


atproto/auth/oauth/oauth.go
```diff
@@ -600,6 +600,21 @@ func (app *ClientApp) StartAuthFlow(ctx context.Context, identifier string) (str
...
+		return nil, fmt.Errorf("missing state query param")
+	}
+
+	info, err := app.Store.GetAuthRequestInfo(ctx, state)
+	if err != nil {
+		return nil, fmt.Errorf("loading auth request info: %w", err)
+	}
+	if info.State != state {
+		return nil, fmt.Errorf("callback state doesn't match request info")
+	}
```
DavidBuchanan314, line 600: Is this check redundant? (given the state was just used to look up the info, I don't think it can ever not match)

atproto/auth/oauth/oauth.go
```diff
@@ -581,19 +581,42 @@ func (app *ClientApp) StartAuthFlow(ctx context.Context, identifier string) (str
...
+		return nil, &ErrCallback{
+			code:        errorCode,
+			description: params.Get("error_description"),
+			uri:         errorUri,
+		}
+	}
+
+	authserverURL := params.Get("iss")
+	authCode := params.Get("code")
+	if authserverURL == "" || authCode == "" {
```
bnewbold, line 621: any reason to move these checks down? I would keep the "validating HTTP request" sort of checks up at the top of the method, before doing any database requests.

it would be more friendly to do separate error strings for each missing param as well (not needed for this PR, but could hit it while touching this PR)

atproto/auth/oauth/oauth.go
bnewbold: depends on how much we trust the auth store implementation!

but yeah it is effectively redundant. I'd be fine keeping it, maybe with a comment?

atproto/auth/oauth/types.go
```diff
@@ -407,3 +407,23 @@ type TokenResponse struct {
 	// Refresh token, for doing additional token requests to the auth server.
 	RefreshToken string `json:"refresh_token"`
 }
+
+// Returned by [ClientApp.ProcessCallback] if the AS signals an error in the redirect URL parameters, per rfc6749 section 4.1.2.1
+//
+// NOTE: This is untrusted data and should not be e.g. rendered to HTML without appropriate escaping
+type ErrCallback struct {
+	code        string
```
bnewbold: I think these fields could be public (Capitalized), to make it easier for calling code to unpack this struct

atproto/auth/oauth/oauth.go
bnewbold: oh, I see now, an error response won't include any of these. hrm! maybe worth a comment up top that `state` query param is always set, but `error_code`, `iss`, and `code` all depend on type of response?

atproto/auth/oauth/types.go
bnewbold: A bit pedantic, but to distinguish "error code" from "HTTP status code", and "error URI" from "HTTP request URI" or "OAuth Redirect URI", I would name these fields:

- `ErrorCode`
- `ErrorDescription`
- `ErrorURI`

atproto/auth/oauth/types.go
```diff
@@ -407,3 +407,23 @@ type TokenResponse struct {
...
+	description string
+	uri         *syntax.URI
+}
+
+func (e *ErrCallback) Error() string {
+	res := "callbackError: " + e.code
```
bnewbold: "OAuth request callback error:"

atproto/auth/oauth/types.go
```diff
@@ -407,3 +407,23 @@ type TokenResponse struct {
 	// Refresh token, for doing additional token requests to the auth server.
 	RefreshToken string `json:"refresh_token"`
 }
+
+// Returned by [ClientApp.ProcessCallback] if the AS signals an error in the redirect URL parameters, per rfc6749 section 4.1.2.1
+//
+// NOTE: This is untrusted data and should not be e.g. rendered to HTML without appropriate escaping
+type ErrCallback struct {
```
bnewbold: I think the idiom is to use `ErrSomething` when doing simple:

    var ErrSomething = errors.New("some thing")

but if defining a type (struct with `Error()` method), it should be `SomethingError`.

I'd call this something like `AuthRequestCallbackError`

bnewbold (approved):
overall looks good! thanks for improving this, definitely should catch explicit error callbacks, and return something helpful with all the relevant metadata.

left some smallish notes

Issue #1135, opened 2025-08-12
OAuth client SDK: logout using token deletion endpoint


bnewbold:
Token deletion endpoint is optional, but if the auth server declares it the client should remember and use it as part of logout.

This probably means adding a new high-level "Logout" helper to `ClientApp` which takes an account DID and session ID.

PR #1156, opened 2025-09-12
OAuth client SDK: Support token revocation via Logout helper

DavidBuchanan314:
Addresses #1135

Revocation has similar properties to session refresh, so I implemented it in `ClientSession` alongside `RefreshTokens`.

A high-level `Logout` helper has been added to `ClientApp` which both revokes the tokens against the AS and deletes the session from the app's session store.

Due to aforementioned similarity between refresh and revocation, I created the `ClientSession.postToAuthServer` helper to reduce duplicated logic. (Note that "DPoP retry" logic is still duplicated elsewhere, I couldn't think of any elegant way to eliminate it)

### DIFF
--- a/atproto/auth/oauth/cmd/oauth-web-demo/main.go
+++ b/atproto/auth/oauth/cmd/oauth-web-demo/main.go
@@ -288,10 +288,10 @@ func (s *Server) OAuthRefresh(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) OAuthLogout(w http.ResponseWriter, r *http.Request) {
 
-	// delete session from auth store
+	// revoke tokens and delete session from auth store
 	did, sessionID := s.currentSessionDID(r)
 	if did != nil {
-		if err := s.OAuth.Store.DeleteSession(r.Context(), *did, sessionID); err != nil {
+		if err := s.OAuth.Logout(r.Context(), *did, sessionID); err != nil {
 			slog.Error("failed to delete session", "did", did, "err", err)
 		}
 	}
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -668,8 +668,15 @@ func (app *ClientApp) Logout(ctx context.Context, did syntax.DID, sessionID stri
 		return err
 	}
 
-	// Tell the AS to revoke the tokens
-	sess.RevokeSession(ctx)
+	// Tell the AS to revoke the tokens, if supported
+	if sess.Data.AuthServerRevocationEndpoint == "" {
+		slog.Info("AS does not support token revocation, skipping RevokeSession")
+	} else {
+		err = sess.RevokeSession(ctx)
+		if err != nil {
+			slog.Warn("error during session revocation", "err", err)
+		}
+	}
 
 	// Delete from our own session store
 	err = app.Store.DeleteSession(ctx, did, sessionID)
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -664,16 +664,12 @@ func (app *ClientApp) ProcessCallback(ctx context.Context, params url.Values) (*
 // High-level helper to delete a session, including revoking access/refresh tokens if supported by the AS
 func (app *ClientApp) Logout(ctx context.Context, did syntax.DID, sessionID string) error {
 	sess, err := app.ResumeSession(ctx, did, sessionID)
-	// TODO: Should this be idempotent? i.e. logging out of a session that does not exist does nothing and succeeds?
 	if err != nil {
 		return err
 	}
 
 	// Tell the AS to revoke the tokens
-	err = sess.RevokeSession(ctx)
-	if err != nil {
-		return err
-	}
+	sess.RevokeSession(ctx)
 
 	// Delete from our own session store
 	err = app.Store.DeleteSession(ctx, did, sessionID)
--- a/atproto/auth/oauth/oauth.go
+++ b/atproto/auth/oauth/oauth.go
@@ -412,14 +412,15 @@ func (app *ClientApp) SendAuthRequest(ctx context.Context, authMeta *AuthServerM
 	}
 
 	parInfo := AuthRequestData{
-		State:                   state,
-		AuthServerURL:           authMeta.Issuer,
-		Scopes:                  scopes,
-		PKCEVerifier:            pkceVerifier,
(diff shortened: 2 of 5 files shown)


atproto/auth/oauth/oauth.go
```diff
@@ -658,3 +660,26 @@ func (app *ClientApp) ProcessCallback(ctx context.Context, params url.Values) (*
 	}
 	return &sessData, nil
 }
+
+// High-level helper to delete a session, including revoking access/refresh tokens if supported by the AS
+func (app *ClientApp) Logout(ctx context.Context, did syntax.DID, sessionID string) error {
+	sess, err := app.ResumeSession(ctx, did, sessionID)
+	// TODO: Should this be idempotent? i.e. logging out of a session that does not exist does nothing and succeeds?
```
DavidBuchanan314: Not sure why I thought this. Returning an error the second time doesn't make it not-idempotent (since internal state is not updated). The behaviour here is fine.

atproto/auth/oauth/session.go
```diff
@@ -77,52 +80,42 @@ type ClientSession struct {
...
 		}
-		body.ClientAssertionType = &ClientAssertionJWTBearer
-		body.ClientAssertion = &clientAssertion
-	}
-
-	vals, err := query.Values(body)
-	if err != nil {
-		return "", err
+		vals.Set("client_assertion_type", ClientAssertionJWTBearer)
+		vals.Set("client_assertion", clientAssertion)
```
DavidBuchanan314, line 99: Note: Because I'm setting the client assertion fields on the query string directly rather than using the struct fields, the client assertion struct fields aren't actually needed - but I figured they might be worth keeping around if we ever implement the server-side of oauth in go, too.

atproto/auth/oauth/session.go
```diff
@@ -77,52 +80,42 @@ type ClientSession struct {
...
-
-	body := RefreshTokenRequest{
-		ClientID:     sess.Config.ClientID,
-		GrantType:    "refresh_token",
-		RefreshToken: sess.Data.RefreshToken,
+// Helper method to handle DPoP retries and client assertions (if the client is confidential)
+// body object will be url-encoded (expected to be either RefreshTokenRequest or RevocationRequest)
+// expects sess.lk to be held by caller
+// if a non-nil *http.Response is returned, the caller is responsible for closing the response body
+func (sess *ClientSession) postToAuthServer(ctx context.Context, url string, body interface{}) (*http.Response, error) {
```
DavidBuchanan314, line 88: I looked into using something a little more type-safe than `interface{}` here, but couldn't come up with anything simple.

atproto/auth/oauth/cmd/oauth-web-demo/main.go
```diff
@@ -288,10 +288,10 @@ func (s *Server) OAuthRefresh(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) OAuthLogout(w http.ResponseWriter, r *http.Request) {
 
-	// delete session from auth store
+	// revoke tokens and delete session from auth store
 	did, sessionID := s.currentSessionDID(r)
 	if did != nil {
-		if err := s.OAuth.Store.DeleteSession(r.Context(), *did, sessionID); err != nil {
+		if err := s.OAuth.Logout(r.Context(), *did, sessionID); err != nil {
```
bnewbold, line 300: Should also update the equivalent section in `doc.go`

atproto/auth/oauth/session.go
```diff
@@ -172,6 +186,46 @@ func (sess *ClientSession) RefreshTokens(ctx context.Context) (string, error) {
 	return sess.Data.AccessToken, nil
 }
 
+// If supported by the AS, use the revocation endpoint to revoke both the access token and the refresh token.
+// This method always succeeds - any errors during revocation are logged but not returned.
+func (sess *ClientSession) RevokeSession(ctx context.Context) {
```
bnewbold: I think the error handling and codeflow here is acceptable, but i'm not sure it is the best pattern.

Having the entire method be fire-and-forget (no errors returned) matches the common "best effort" use-case, but I feel like that behavior should be left to calling code. Which as-is can not tell whether the operation was successful or even attempted.

We should try revoking the second token even if the first revocation failed (as the current code does). But maybe only if there was a bad HTTP response, as opposed to a non-nil `err` from `postToAuthServer`? 

There is a style argument that the `sess.Data.AuthServerRevocationEndpoint == ""` should be checked in calling code, and that this function should throw an error instead of logging if called when that endpoint isn't defined.

(sorry that this is a bit vague feedback)

atproto/auth/oauth/session.go
bnewbold: A minor nit, but can use `any` as a simpler alias for `interface{}` in type declarations.

atproto/auth/oauth/session.go
```diff
@@ -140,13 +133,34 @@ func (sess *ClientSession) RefreshTokens(ctx context.Context) (string, error) {
 				// already updated nonce value above; loop around and try again
 				continue
 			}
-			return "", fmt.Errorf("token refresh failed (HTTP %d): %s", resp.StatusCode, reason)
+			return nil, fmt.Errorf("request failed (HTTP %d): %s", resp.StatusCode, reason)
```
bnewbold: "auth server request failed"? (I could go either way)

atproto/auth/oauth/types.go
```diff
@@ -407,3 +413,21 @@ type TokenResponse struct {
 	// Refresh token, for doing additional token requests to the auth server.
 	RefreshToken string `json:"refresh_token"`
 }
+
+// The fields which are included in a token revocation request. These HTTP POST bodies are form-encoded, so use URL encoding syntax, not JSON.
```
bnewbold, line 417: we haven't been particularly consistent about it, but we could toss in a link to RFC7009 here

bnewbold (approved):
Left a bit of feedback, but overall this looks good.

I tested locally (public client mode) and it worked against my self-hosted PDS. Didn't try against entryway.

atproto/auth/oauth/session.go
DavidBuchanan314: I moved the AuthServerRevocationEndpoint check up into the logout helper, and made RevokeSession error if it's unset. I also used `errors.Join()` to return one or both HTTP errors if they exist, so the caller can decide what to do.

`Logout()` handles those errors by logging and continuing, and someone could write their own Logout method if they want to customise that further.

PR #1164, opened 2025-10-07
OAuth: create StartAuthFlowWithUserData

DavidBuchanan314:
API/functionality not final.

When _not_ using an OAuth SDK, developers have the option to stash additional data in the `state` string (either directly, or by using the state string as a lookup key). The indigo SDK handles `state` strings internally and so does not expose this option to the developer.

This PR adds a "userdata" field to bring back some of that flexibility. It can be set initially via the newly-added `StartAuthFlowWithUserData` method, which will persist it via the session store. The field is later accessible via the `AuthRequestData` or `ClientSessionData` structs (e.g. when handling the callback).

This is a non-breaking change - the original `StartAuthFlow` method can be used unchanged.


bnewbold:
We could have `StartAuthFlow` return the full `info` struct, in addition to the redirect URL. That would let calling code extract the `state` and do whatever persistence it wants. This would be a breaking change but not too bad, and there might be other situations where calling code wants that info. I'd be in favor of doing this regardless of storing metadata as well.

I'm not totally against the idea of persisting extra metadata fields, especially in the request info, but it feels like a bit of a hack and adds complexity to the API. Having initial metadata stick around over the full lifetime of the session (aka, include in the session as well) feels a bit off. May want to think on it a bit more.

If we did go with attaching metadata, I'd maybe not call it "UserData". "User data" or "user function" are fairly common terms for this sort of thing, but are still a bit confusing (who is the "user"? the account, or the developer?). I'd probably say "ExtraMetadata" or something. And maybe should be `map[string]string` to cover common cases; folks can still always stick JSON in a string value if they want.

PR #1165, opened 2025-10-08
add a few helpers for working with posts in a not ugly way

whyrusleeping:


api/bsky/feedpost_util.go
```diff
@@ -0,0 +1,37 @@
...
+	return nil, false
+}
+
+func (fp *FeedPost) GetReplyParentUri() (string, bool) {
+	if fp.Reply != nil && fp.Reply.Parent != nil {
+		return fp.Reply.Parent.Uri, true
+	}
+
+	return "", false
+}
```
haileyok, line 37: may as well add one of these for the root too, i find myself needing that a lot
```suggestion
func (fp *FeedPost) GetReplyRootUri() (string, bool) {
	if fp.Reply != nil && fp.Reply.Root != nil {
		return fp.Reply.Root.Uri, true
	}

	return "", false
}

func (fp *FeedPost) GetReplyParentUri() (string, bool) {
	if fp.Reply != nil && fp.Reply.Parent != nil {
		return fp.Reply.Parent.Uri, true
	}

	return "", false
}
```

haileyok (approved):

PR #1170, opened 2025-10-09
Tap (sync tool)

dholms:
Sync tool for backfilling and consuming the firehose in order to hand off a fully authenticated stream of operations to an application.

More details in the included README.

TS client lib: https://github.com/bluesky-social/atproto/pull/4290/files

### DIFF
--- a/cmd/tap/README.md
+++ b/cmd/tap/README.md
@@ -14,6 +14,8 @@ Features and design decisions:
 - SQLite or Postgres backend
 - designed for moderate scale (millions of repos, 30k+ events/sec)
 
+There is a convenient client library in Typescript for working with Tap at [@atproto/tap](https://github.com/bluesky-social/atproto/blob/main/packages/tap/README.md)
+
 ⚠️ Tap is still in beta and may have bugs. Please report any issues you encounter.
 
 ## Quick Start
@@ -80,7 +82,7 @@ Environment variables or CLI flags:
 
 Tap supports three delivery modes:
 
-**WebSocket with acks** (default): Client sends acks each event once it has been processed/persisted. Ensures that no data is lost and client does not need to handle cursors. It's recommended to use a client library such as [@atproto/tap](https://github.com/bluesky-social/atproto/tree/main/packages/tap) when using this mode.
+**WebSocket with acks** (default): Client sends acks each event once it has been processed/persisted. Ensures that no data is lost and client does not need to handle cursors. It's recommended to use a client library such as [@atproto/tap](https://github.com/bluesky-social/atproto/tree/main/packages/tap/README.md) when using this mode.
 
 **Fire-and-forget**: Set `TAP_DISABLE_ACKS=true`. Events are sent and considered "acked" once the client receives them. Simpler but may result in data loss. Recommended for testing purposes or when data integrity is not critical.
 


nexus/backfill.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	comatproto "github.com/bluesky-social/indigo/api/atproto"
+	"github.com/bluesky-social/indigo/atproto/data"
```
bnewbold: just renamed some of these packages to `atdata`, `atcrypto`, etc. sorry!

nexus/nexus.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if err := db.AutoMigrate(&models.Did{}, &models.RepoRecord{}, &models.OutboxBuffer{}, &models.BackfillBuffer{}, &models.Cursor{}); err != nil {
+		return nil, err
+	}
+
+	e := echo.New()
+	e.HideBanner = true
+
+	bdir := identity.BaseDirectory{
+		SkipHandleVerification: true,
```
bnewbold: :+1:

nexus/nexus.go
```diff
@@ -0,0 +1,146 @@
...
+		TryAuthoritativeDNS:    false,
+		SkipDNSDomainSuffixes:  []string{".bsky.social"},
+	}
+	cdir := identity.NewCacheDirectory(&bdir, 1_000_000, time.Hour*24, time.Minute*2, time.Minute*5)
```
bnewbold: can be network DAU or even more (eg, 2 million is a good value right now)

cmd/nexus/nexus.go
```diff
@@ -0,0 +1,146 @@
...
+		Outbox:    n.outbox,
+	}
+
+	cursor, err := n.EventProcessor.ReadLastCursor(context.Background(), config.RelayHost)
+	if err != nil {
+		return nil, err
+	}
+
+	rsc := &events.RepoStreamCallbacks{
+		RepoCommit: func(evt *comatproto.SyncSubscribeRepos_Commit) error {
```
bnewbold: you'll need to handle `#sync`, `#account`, and `#identity` events as well.

for identity, call Purge method on caching directory. this ensures signing keys are fresh after account migration (for example).

for account, might need to change account status (eg, if deleted/deactivated)

nexus/types.go
```diff
@@ -0,0 +1,41 @@
...
+			Rkey:       op.Rkey,
+			Action:     op.Action,
+			Record:     op.Record,
+			Cid:        op.Cid,
+		})
+	}
+	return ops
+}
+
+type Op struct {
```
bnewbold: I suspect we'll need an account status op as well. Eg, jetstream has this. for account deletion, deactivation.

nexus/outbox.go
```diff
@@ -0,0 +1,88 @@
...
+type Outbox struct {
+	db     *gorm.DB
+	outCh  chan *Op
+	logger *slog.Logger
+}
+
+func NewOutbox(db *gorm.DB) *Outbox {
+	return &Outbox{
+		db:     db,
+		outCh:  make(chan *Op, 100),
```
bnewbold: my intuition is to use a large chan depth, like thousands

nexus/backfill.go
dholms: all good 👍

cmd/nexus/nexus.go
dholms: yup totally 👍 doing sync events next them will handle the other two

cmd/tap/main.go
```diff
@@ -0,0 +1,278 @@
...
+	"context"
+	"log/slog"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	_ "github.com/joho/godotenv/autoload"
+
+	"github.com/carlmjohnson/versioninfo"
```
bnewbold: we updated all projects to `github.com/earthboundkid/versioninfo/v2` (which is a drop-in maintained fork

cmd/tap/main.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/urfave/cli/v2"
```
bnewbold: we updated everything to `v3`. I can do the upgrade after this is merged, or you can hit the changes:

- function signature becomes `(ctx context.Context, cmd *cli.Command)`
- `cli.App` becomes `cli.Command`
- env vars become `Sources: cli.EnvVars("RELAY_PLC_HOST", "ATP_PLC_HOST"),`
- invoke the app like `app.Run(context.Background(), args)`

atproto/repo/sync.go
```diff
@@ -164,31 +164,41 @@ func parseCommitOps(ops []*comatproto.SyncSubscribeRepos_RepoOp) ([]Operation, e
...
+}
+
+func VerifyCommitSignatureFromCar(ctx context.Context, dir identity.Directory, car []byte) (*Commit, error) {
+	commit, _, err := LoadCommitFromCAR(ctx, bytes.NewReader(car))
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	if err := commit.VerifyStructure(); err != nil {
```
bnewbold, line 176: was pulling out this verification intentional? should check that we do equivalent verification everywhere else this code is used if we do.

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
+`tap`: atproto sync utility
+========================================
+
+Tap is a single-tenant service that subscribes to an atproto relay and outputs filtered, verified events for a subset of repos.
+
+Tap simplifies firehose consumption by handling verification, backfill, and filtering. Your application connects to tap and receives simple JSON events for only the repos and collections you care about. Historical data for configured repos is automatically fetched from PDSs and delivered before live events begin.
```
bnewbold: might be worth noting that client specifically do not need to deal with binary formats (CBOR) or cryptographic signatures.

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+Features and design decisions:
+
+- verifies repo structure, MST integrity, and identity signatures
+- automatic backfill: fetches full repo history from PDS when adding new repos
+- filtered output: by DID list, by collection, or full network mode
+- ordering guarantees: live events wait for historical backfill to complete
+- delivery modes: WebSocket with acks, fire-and-forget, or webhook
+- single golang binary, SQLite backend
```
bnewbold: or postgresql (if we officially support that)

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+Each repo will be backfilled from its PDS, then live events will stream as they arrive from the relay.
+
+## HTTP API
+
+- `GET /health`: returns `{"status":"ok"}`
+- `POST /add-repos`: add DIDs to track (triggers backfill of added repos)
+- `POST /remove-repos`: remove DIDs (stops sync, deletes data. does not delete buffered events in outbox)
+- `GET /channel`: WebSocket endpoint to receive events
+
+If more than one client connects to the WebSocket, events will be transparently sharded across all connected clients.
```
bnewbold: are there any guarantees about this sharding? eg, will the same all messages for the same DID go to the same connection? gets weird with re-connections I suppose.

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+  -d '{"dids": ["did:plc:z72i7hdynmk6r22z27h6tvur"]}' # @bsky.app repo
+```
+
+Each repo will be backfilled from its PDS, then live events will stream as they arrive from the relay.
+
+## HTTP API
+
+- `GET /health`: returns `{"status":"ok"}`
+- `POST /add-repos`: add DIDs to track (triggers backfill of added repos)
+- `POST /remove-repos`: remove DIDs (stops sync, deletes data. does not delete buffered events in outbox)
```
bnewbold: what data gets deleted, just record state? does it emit deletion ops for each record?

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+- `GET /channel`: WebSocket endpoint to receive events
+
+If more than one client connects to the WebSocket, events will be transparently sharded across all connected clients.
+
+## Configuration
+
+Environment variables or CLI flags:
+
+- `TAP_DB_PATH`: path to SQLite database file (default: `./tap.db`)
```
bnewbold: elsewhere we use `DATABASE_URL` (which is something of a generic norm/standard), and use `sqlite://` or `postgres://` prefix to distinguish.

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+- `TAP_RELAY_URL`: atproto relay URL (default: `https://relay1.us-east.bsky.network`)
+- `TAP_BIND`: HTTP server address (default: `:8080`)
+- `TAP_FIREHOSE_PARALLELISM`: concurrent firehose event processors (default: `10`)
+- `TAP_RESYNC_PARALLELISM`: concurrent resync workers (default: `5`)
+- `TAP_OUTBOX_PARALLELISM`: concurrent outbox workers (default: `5`)
+- `TAP_CURSOR_SAVE_INTERVAL`: how often to save cursor (default: `5s`, set to `0` to disable)
```
bnewbold: I assume this is the interval to persist upstream cursor, right? worth clarifying

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+Tap supports three delivery modes:
+
+**WebSocket with acks** (default): Client sends acks each event once it has been processed/persisted. Ensures that no data is lost and client does not need to handle cursors. It's recommended to use a client library such as (@TODO) when using this mode.
+
+**Fire-and-forget**: Set `TAP_DISABLE_ACKS=true`. Events are sent and considered "acked" once the client receives them. Simpler but may result in data loss. Recommended for testing purposes or when data integrity is not critical.
+
+**Webhook**: Set `TAP_WEBHOOK_URL=http://...`. Events are POSTed as JSON. Events considered "acked" once the webhook responds with a 200. Recommended for lower throughput serverless environments.
+
+
+## Network Boundary Modes
```
bnewbold, line 92: I think `TAP_COLLECTION_FILTERS` should also be mentioned in this section

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+  }
+}
+```
+
+**User events** (handle or status changes):
+
+```json
+{
+  "id": 12346,
+  "type": "user",
```
bnewbold: maybe `account` instead of `user`? I don't have a super strong opinion, but we generally use "account" instead of "user" elsewhere in the protocol and SDKs

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+  "user": {
+    "did": "did:plc:abc123",
+    "handle": "alice.bsky.social",
```
bnewbold, line 148: can handles be trusted? I assume `tap` is doing resolution to confirm, and they can be trusted

cmd/tap/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+Filters apply to record events only. User events are always delivered for tracked repos.
+
+## Operations
+
+Tap logs to stdout in JSON format. The firehose consumer automatically reconnects with exponential backoff on relay failures. Cursor position is saved periodically (default 5 seconds) and restored on restart.
+
+SQLite is tuned for high write throughput: WAL mode, 10-second busy timeout, `synchronous=NORMAL`, 64MB cache, batched deletes. The outbox buffers up to 1M pending events in memory.
+
+Resync is automatic: if a commit does not validate according to [Sync v1.1](https://github.com/bluesky-social/proposals/tree/main/0006-sync-iteration) semantics, the repo is marked `desynced` until it can be refetched from the authoritative PDS. Live events during resync are buffered and replayed after completion. Failures trigger exponential backoff (1 minute → 1 hour max).
```
bnewbold: The actual status string is `desynchronized`

cmd/tap/main.go
```diff
@@ -0,0 +1,278 @@
...
+	tap, err := NewTap(config)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(cctx.Context)
+
+	if !config.OutboxOnly {
+		go func() {
+			if err := tap.Crawler.Run(ctx); err != nil {
```
bnewbold: if the crawler crashes, the process just keeps chugging along? I guess that isn't the worst behavior, but i'd sort of expect it to retry or fail more clearly

cmd/tap/server.go
```diff
@@ -0,0 +1,262 @@
...
+	ts.echo.HideBanner = true
+
+	// Apply admin token middleware if configured
+	if ts.adminToken != "" {
+		ts.echo.Use(ts.adminAuthMiddleware)
+	}
+
+	ts.echo.GET("/health", ts.handleHealthcheck)
+	ts.echo.GET("/channel", ts.handleChannelWebsocket)
+	ts.echo.POST("/add-repos", ts.handleAddRepos)
```
bnewbold: could be `/repos/add` and `/repos/remove`

cmd/tap/crawler.go
```diff
@@ -0,0 +1,199 @@
...
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"time"
+
+	comatproto "github.com/bluesky-social/indigo/api/atproto"
+	"github.com/bluesky-social/indigo/atproto/atclient"
+	"github.com/bluesky-social/indigo/cmd/tap/models"
```
bnewbold, line 12: as a house style, we usually have multiple sections of imports separated by newlines):

- stdlib imports (like you have)
- any underscore imports (like pprof)
- anything `bluesky-social`
- all others

not a huge deal, and I can cleanup after this merges if needed

cmd/tap/util.go
```diff
@@ -0,0 +1,77 @@
...
+			continue
+		}
+		if collection.String() == signalColl {
+			return true
+		}
+	}
+	return false
+}
+
+func isRateLimitError(err error) bool {
```
bnewbold, line 62: `atclient.APIClient` returns `atclient.APIError`, not `xrpc.Error`.

You can do the same `errors.As`, and then check if the error `StatusCode` is `http.StatusTooManyRequests` (HTTP 451).

cmd/tap/db_helpers.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func deleteRepo(tx *gorm.DB, did string) error {
+	if err := tx.Delete(&models.RepoRecord{}, "did = ?", did).Error; err != nil {
+		return err
+	}
+	if err := tx.Delete(&models.ResyncBuffer{}, "did = ?", did).Error; err != nil {
+		return err
+	}
+	return tx.Delete(&models.Repo{}, "did = ?", did).Error
```
bnewbold, line 15: will the account just get re-initialized if it is seen again? I kind of assume this should result in the status getting set to `deleted`, as a sort of tombstone row.

cmd/tap/crawler.go
```diff
@@ -0,0 +1,199 @@
...
+				State:  models.RepoStatePending,
+				Status: models.AccountStatusActive,
+			})
+		}
+
+		if len(repos) == 0 {
+			break
+		}
+
+		if err := c.db.Save(&repos).Error; err != nil {
```
bnewbold: If feels like the 'ON UPDATE DO' behavior here should be more explicit, either with a code comment or clause.

If repos already existed in the table, should the state and status be set to pending/active? Or 'DO NOTHING'?

cmd/tap/crawler.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		repoList, err := comatproto.SyncListRepos(ctx, client, cursor, 1000)
```
bnewbold, line 83: I don't know if it is worth mentioning in the docs or something, but in the new relay implementation this isn't very complete: it only lists accounts which the relay has "seen" since it started (so any accounts which have not been active are not included).

This is distinct from collectiondir, which goes to PDS instances and enumerates all accounts in a backfill stage. An archive service might do that and have a more complete listRepos endpoint in the future.

cmd/tap/main.go
```diff
@@ -0,0 +1,278 @@
...
+		&cli.IntFlag{
+			Name:    "db-max-conns",
+			Usage:   "maximum number of database connections",
+			Value:   32,
+			EnvVars: []string{"TAP_DB_MAX_CONNS"},
+		},
+		&cli.StringFlag{
+			Name:    "relay-url",
+			Usage:   "AT Protocol relay URL",
+			Value:   "https://relay1.us-east.bsky.network",
```
bnewbold: confusion around whether these URLs should be `https://` or `wss://` has been common across our go projects. I'd recommend either being flexible and normalizing consistently, or being strict and bail out early (eg, when parsing args) if it is wrong.

cmd/tap/README.md
dholms: No no guarantees around what goes to which shard. Except for the general deliverability guarantees that live updates for a repo will not be delivered until the previous update for that repo has been acked.

cmd/tap/README.md
dholms: yeah record state. I should say "deletes repo metadata"

no deletion ops for records as the records aren't actually being deleted, we're just no longer tracking/syncing that repo

cmd/tap/README.md
dholms: ah good looks. artifact of starting with just sqlite

cmd/tap/README.md
dholms: yup 👍

cmd/tap/README.md
dholms: I used `user` specifically because this combines both `identity` and `account` semantics.

But maybe it's better to just use an existing term. I think i'd prefer `identity` over `account`

cmd/tap/main.go
dholms: yeah hmm it's probably worth doing this on a loop anyway. even in the case where the crawler doesn't crash just because you could finish enumeration but then some new repositories pop up that you want to start crawling

cmd/tap/db_helpers.go
dholms: hmm well it depends what mode you're running it in. In full network/signal collection mode, then yes. If you're manually adding/removing repos then no.

I would say this method doesn't make much sense to call unless you're manually adding/removing repos

cmd/tap/crawler.go
dholms: yeah that's a great point. if it already exists then this should DO NOTHING

cmd/tap/crawler.go
dholms: hmm i didn't realize that. that's a thinker 🤔 

do we have anywhere to go to `listRepos` for the full network (after we deprecate the old relay)?

cmd/tap/README.md
dholms: yup 👍

atproto/repo/sync.go
dholms: I coulda sworn there was some reason why I removed this but looking back over it I can't figure out why. Good catch, I'll add it back in

cmd/tap/outbox.go
```diff
@@ -0,0 +1,336 @@
...
+	"time"
+)
+
+// Ordering guarantees for events belonging to the same DID:
+//
+// Live events are synchronization barriers - all prior events must complete
+// before a live event can be sent, and the live event must complete before
+// any subsequent events can be sent.
+//
+// Historical events can be sent concurrently with each other (no ordering
```
jcalabro, line 18: This is interesting, mind digging in a bit about why it's ok to send historical events concurrently?

What if there are two historical events, one that creates a record, then one that deletes that same record (and they were submitted in that order at some point in the past). What if the sync tool processes the delete before the create?

cmd/tap/outbox.go
```diff
@@ -0,0 +1,336 @@
...
+
+type Outbox struct {
+	logger      *slog.Logger
+	mode        OutboxMode
+	parallelism int
+	webhook     *WebhookClient
+
+	events *EventManager
+
+	didWorkers sync.Map // map[string]*DIDWorker
```
jcalabro: Small go thing, but if you're looking for a strongly typed thread safe map that's quite good in a lot of scenarios, [xsync.Map](https://pkg.go.dev/github.com/puzpuzpuz/xsync/v4#Map) might be worth looking at. We use that in lots of places in the data plane

cmd/tap/event_manager.go
```diff
@@ -0,0 +1,345 @@
...
+)
+
+type EventManager struct {
+	logger *slog.Logger
+	db     *gorm.DB
+
+	nextID atomic.Uint64
+
+	cacheSize       int
+	finishedLoading bool
```
jcalabro: This should be an [atomic.Bool](https://pkg.go.dev/sync/atomic#Bool). It's currently a race condition

cmd/tap/README.md
```diff
@@ -0,0 +1,185 @@
...
+```bash
+# Run tap
+go run ./cmd/tap --disable-acks=true
+# By default, the service uses SQLite at `./tap.db` and binds to port `:8080`.
+
+# In a separate terminal, connect to receive events:
+websocat ws://localhost:8080/channel
+
+#  Add a repo to track
+curl -X POST http://localhost:8080/repos/add \
```
jcalabro: port 2480 rather than 8080?

cmd/tap/event_manager.go
```diff
@@ -0,0 +1,345 @@
...
+	cache   map[uint]*OutboxEvt
+	cacheLk sync.RWMutex
+
+	pendingIDs chan uint
+}
+
+type DBCallback = func(tx *gorm.DB) error
+
+func (em *EventManager) IsFull() bool {
+	return len(em.cache) >= em.cacheSize
```
jcalabro, line 37: This needs a read lock

cmd/tap/event_manager.go
```diff
@@ -0,0 +1,345 @@
...
+			if err := tx.Clauses(clause.OnConflict{
+				Columns:   []clause.Column{{Name: "did"}, {Name: "collection"}, {Name: "rkey"}},
+				DoUpdates: clause.AssignmentColumns([]string{"cid"}),
+			}).CreateInBatches(toPut, 100).Error; err != nil {
+				return err
+			}
+		}
+
+		if len(dbEvts) > 0 {
+			if err := tx.CreateInBatches(dbEvts, 100).Error; err != nil {
```
jcalabro, line 251: One thing to note with gorm is that pretty much all their functions have a `WithContext` function that allows you to pass cancellation tokens through nicely. Important to ensure things don't hang forever unexpectedly

cmd/tap/main.go
```diff
@@ -0,0 +1,293 @@
...
+			&cli.DurationFlag{
+				Name:    "cursor-save-interval",
+				Usage:   "how often to save firehose cursor",
+				Value:   1 * time.Second,
+				Sources: cli.EnvVars("TAP_CURSOR_SAVE_INTERVAL"),
+			},
+			&cli.DurationFlag{
+				Name:    "repo-fetch-timeout",
+				Usage:   "timeout when fetching repo CARs from PDS (e.g. 180s for slow hosts)",
+				Value:   30 * time.Second,
```
jcalabro: I was actually running in to some timeout issues on some larger accounts (I was testing with Jerry's repo). 5m might be a better value for real-world use cases?

cmd/tap/outbox.go
jcalabro: Thought about this a bit more - is this valid because these deleted messages from the repo are actually fully gone? Like, in the actual repo, we don't keep storing `create` and `delete`, and "historical" messages are just what remains after the all-time list of operations had been applied? AKA there's just nothing in the repo pertaining to message X after X has been deleted, so tap will never get that message?

dholms: Yes exactly 👍 

"Historical events" are done by doing a full sync of the repo which we then translate into a list of ops by comparing to the metadata we have stored about the repo. All the ops in that delta can be thought of as happening at the same in some sense

cmd/tap/event_manager.go
dholms: ah nice good call out 👍

PR #1174, opened 2025-10-10
Relay:Fix flag type initial-seq-number

### LINKED ISSUE #1176: Relay:initial-seq-number, default-account-limit and new-hosts-per-day-limit parameters always 0
The parameters initial-seq-number, default-account-limit and new-hosts-per-day-limit are always 0 because of the wrong flag type IntFlag instead of Int64Flag

wdantuma:
Relay is not starting because initial-seq-number is always 0, resulting from typecast from int to int64 fails. Changed flag type of initial-seq-number fom int to Int64.

default-account-limit and new-hosts-per-day-limit had the same issue

fixes #1176

### DIFF
--- a/cmd/relay/main.go
+++ b/cmd/relay/main.go
@@ -103,13 +103,13 @@ func run(args []string) error {
 					Sources: cli.EnvVars("RELAY_HOST_CONCURRENCY", "RELAY_CONCURRENCY_PER_PDS"),
 					Value:   40,
 				},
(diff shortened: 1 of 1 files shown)


verdverm:
Encountered this issue as well and **can confirm the changes here do fix the issue**.

I would note there is still intermixing of `Int` vs `Int64` between the flags and it may be helpful to make usage consistent between all of them.

PR #1187, opened 2025-11-10
Add warning counter for non-active accounts in event processing

erikologic:
I was trying to follow dropped events, and realised this metric is not being written.

### DIFF
--- a/cmd/relay/relay/ingest.go
+++ b/cmd/relay/relay/ingest.go
@@ -99,7 +99,7 @@ func (r *Relay) processCommitEvent(ctx context.Context, evt *comatproto.SyncSubs
 	}
 
 	if !acc.IsActive() {
-		logger.Info("dropping commit message for non-active account", "status", acc.Status, "upstreamStatus", acc.UpstreamStatus)
+		logger.Info("dropping commit message for inactive-active account", "status", acc.Status, "upstreamStatus", acc.UpstreamStatus)
 		eventsWarningsCounter.WithLabelValues(hostname, "inactive-account").Add(1)
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


cmd/relay/relay/metrics.go
```diff
@@ -11,7 +11,7 @@ var EventsReceivedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
 	Help: "The total number of events received",
 }, []string{"pds"})
 
-var eventsWarningsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+var EventsWarningsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
```
bnewbold: I don't think that this needs to be exported (capitalized)?

The `EventsReceivedCounter` above gets exposed via the admin web UI, which is why that one counter is exported.

cmd/relay/relay/ingest.go
```diff
@@ -159,6 +160,7 @@ func (r *Relay) processSyncEvent(ctx context.Context, evt *comatproto.SyncSubscr
 
 	if !acc.IsActive() {
 		logger.Info("dropping sync message for non-active account", "status", acc.Status, "upstreamStatus", acc.UpstreamStatus)
+		EventsWarningsCounter.WithLabelValues(hostname, "non-active").Add(1)
```
bnewbold: I think the "warn" label value should be "inactive-account", not "non-active" (the log line above could also be changed from "non-active" to "inactive").

This also applies to line 103 in the same file.

bnewbold (commented):
Thanks for the submission, left some small notes.

cmd/relay/relay/metrics.go
erikologic: fixed

cmd/relay/relay/ingest.go
erikologic: fixed

bnewbold (approved):

bnewbold:
The full relay test suite does have a flakey test, which also fails in CI sometimes. I have not had a chance to track it down yet.

PR #1192, opened 2025-11-12
Make repo fetch timeout configurable in `sync-tool`

aliceisjustplaying:
Some big repos take more than 30 seconds to fetch as I have learned the hard way.


dholms (approved):
looks good thanks!

PR #1197, opened 2025-11-16
lexicon SDK: stricter schema parsing

bnewbold:
These are mostly to align with recent changes documented here: https://github.com/bluesky-social/atproto/discussions/4343

A couple other nits and corner-cases also turned up.

### DIFF
--- a/atproto/lexicon/language.go
+++ b/atproto/lexicon/language.go
@@ -459,6 +459,7 @@ func (s *SchemaSubscription) CheckSchema() error {
 
 type SchemaPermissionSet struct {
 	Type        string             `json:"type"` // "permission-set"
+	Description *string            `json:"description,omitempty"`
 	Title       *string            `json:"title,omitempty"`
 	TitleLangs  map[string]string  `json:"title:langs,omitempty"`
 	Detail      *string            `json:"detail,omitempty"`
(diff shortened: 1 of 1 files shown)


atproto/lexicon/language.go
```diff
@@ -65,6 +65,44 @@ func (s *SchemaDef) CheckSchema() error {
...
+	return true
+}
+
+// Checks if the type is allowed to be included in object fields.
+func (s *SchemaDef) IsFieldType() bool {
+	if s.IsConcrete() {
+		return true
+	}
+	switch s.Inner.(type) {
+	case SchemaArray, SchemaObject, SchemaRef, SchemaUnion, SchemaUnknown:
```
mary-ext, line 100: object schema isn't allowed to be nested is it not?

bnewbold: objects can be nested! at least in my understanding, I can't find an example from a quick grep

mary-ext: interesting! I know that the TS implementation currently forbids them

https://github.com/bluesky-social/atproto/blob/b7bc95d6abe1136d3ad23b2ff23f6a9991bb373a/packages/lexicon/src/types.ts#L171-L197

mary-ext: d'oh perhaps the language on the current lexicon specification isn't helping either with this confusion

objects and tokens are specified under ["field type definitions"](https://atproto.com/specs/lexicon#field-type-definitions) which would suggest that they can be inlined but it's only when you get to their descriptions that they specify explicitly that it must be ["by reference"](https://atproto.com/specs/lexicon#field-type-definitions:~:text=inside%20other%20definitions-,by%20reference,-.)

bnewbold: this turns out to be an open debate and difference between implementations, tracking it here: https://github.com/bluesky-social/atproto-website/issues/470

PR #1198, opened 2025-11-17
lexicon: more granular account and identity parsing

bnewbold:
There is maybe an open question here around how strict the validation should be when parsing a schema file, versus parsing permissions.

Eg, should `CheckSchema()` on an overall schemafile actually fail for unsupported permission resource types? What about additional account/identity `attr`? We definitely want linting to flag this sort of thing, this comes down to parsing `permission-set` declarations.

I think being strict to start here is fine. Auth scope string parsing will be separate, and we can revisit when adding specific SDK support for permissions.

### DIFF
--- a/atproto/lexicon/language.go
+++ b/atproto/lexicon/language.go
@@ -432,6 +432,7 @@ func (s *SchemaPermissionSet) CheckSchema() error {
 	return nil
 }
 
+// Because the 'action' field could be a string or array of strings, this schema type has custom JSON marshalling behavior.
 type SchemaPermission struct {
 	Type        string  `json:"type"` // "permission"
 	Description *string `json:"description,omitempty"`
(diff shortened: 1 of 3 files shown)


atproto/lexicon/testdata/catalog/permission-set.json
```diff
@@ -88,6 +88,28 @@
...
+        {
+          "type": "permission",
+          "resource": "identity",
+          "attr": "*"
+        },
+        {
+          "type": "permission",
+          "resource": "identity",
+          "attr": "handle",
+          "action": "submit"
```
matthieusieben, line 112: These are allowed (by the syntax) but should be ignored when interpreted from permission sets.

atproto/lexicon/testdata/lexicon-valid.json
```diff
@@ -10,5 +10,60 @@
...
+              "type": "permission",
+              "resource": "identity",
+              "attr": "*"
+            },
+            {
+              "type": "permission",
+              "resource": "identity",
+              "attr": "handle",
+              "action": "submit"
+            }
```
matthieusieben, line 63: These are allowed (by the syntax) but should be ignored when interpreted from permission sets.

atproto/lexicon/language.go
```diff
@@ -510,6 +592,40 @@ func (s *SchemaPermission) CheckSchema() error {
...
+			return fmt.Errorf("unsupported identity permission 'attr': %s", s.Attribute)
+		}
+		if len(s.Action) > 0 {
+			switch s.Action[0] {
+			case "submit", "manage":
+				// pass
+			default:
+				return fmt.Errorf("unsupported identity permission 'action': %s", s.Action[0])
+			}
+		}
```
matthieusieben, line 628: Don't think we need this

>

PR #1199, opened 2025-11-20
a few perf tweaks

whyrusleeping:
Theres a few tweaks here that should help perf, i was getting ~200k/s before and see up to ~260k after this, but the biggest perf thing that actually is getting in the way is json marshaling and unmarshaling, basically the fact that we are loading from the DB, unmarshaling the thing, then marshaling it all back to send it down the websocket is creating a lot of excess work and memory pressure that is leading to a good deal of slowness. Ideally we can set it up so that we don't need to unmarshal from the database, instead using something like json.RawMessage to 'pre-cache' the marshaled json. Thats a larger refactor than I wanted to try out without chatting with @dholms first.

### DIFF
--- a/cmd/nexus/outbox.go
+++ b/cmd/nexus/outbox.go
@@ -542,7 +542,6 @@ func (ec *EventCache) loadEventsParallel(startID int) (int, int, error) {
 	resultsSize := batchesSize(results)
 	// didn't load any events
 	if resultsSize == 0 {
-		time.Sleep(500 * time.Millisecond)
 		return startID, 0, nil
 	}
 


cmd/nexus/outbox.go
```diff
@@ -535,7 +549,8 @@ func (ec *EventCache) loadEventsParallel(startID int) (int, int, error) {
 	ec.cacheMu.Lock()
 	for _, result := range results {
 		for _, evt := range result.events {
-			ec.eventCache[evt.ID] = &evt
+			evtCopy := evt // Create heap copy to avoid storing address of loop variable
```
whyrusleeping, line 551: This was a bug that was probably causing data inconsistency issues, taking the address of a loop variable results in unexpected outcomes because the loop variables are reused

cmd/nexus/outbox.go
```diff
@@ -549,10 +564,8 @@ func (ec *EventCache) loadEventsParallel(startID int) (int, int, error) {
 }
 
 func (ec *EventCache) loadEventBatch(startID int, endID int) batchResult {
-	var dbEvts []models.OutboxBuffer
-	err := ec.db.Where("id > ? AND id <= ?", startID, endID).
-		Order("id ASC").
-		Find(&dbEvts).Error
+	dbEvts := make([]models.OutboxBuffer, 0, (endID-startID)+1)
+	err := ec.db.Raw("SELECT * FROM outbox_buffers WHERE id > ? AND id <= ? ORDER BY id ASC", startID, endID).Scan(&dbEvts).Error
```
whyrusleeping, line 567: Gorm is great, but its fancy clause stuff gets expensive in tight loop code.

cmd/nexus/processor.go
```diff
@@ -32,9 +34,16 @@ type EventProcessor struct {
 	SignalCollection  string
 	CollectionFilters []string
 
+	repoStateCache *lru.TwoQueueCache[string, *cachedRepoState]
```
whyrusleeping, line 37: we were spending a good chunk of time just calling GetRepoState and pulling stuff out of the DB for it, this change isnt super necessary (and does touch a lot of stuff unfortunately) but it feels safe enough to me

PR #1204, opened 2025-11-25
Satisfy sync.listRepos with JOIN

dholms:
This fixes a bug where `sync.listRepos` returns a 500 if we fail to read the current commit from the DB.

Also fixes a long-standing TODO around satisfying this method with a JOIN rather than repeated SELECTs.


bnewbold (approved):
nice!

bnewbold:
this fix has now been deployed

PR #1207, opened 2025-11-30
auth: add permission scope string support

bnewbold:
This is copied over from cobalt, which got review earlier: https://tangled.org/bnewbold.net/cobalt/tree/main/permissions

I did a pass over implementation to check against written specs.

### DIFF
--- a/atproto/auth/permission.go
+++ b/atproto/auth/permission.go
@@ -133,7 +133,7 @@ func ParsePermissionString(scope string) (*Permission, error) {
 		if p.Attribute != "" && p.Attribute != "email" && p.Attribute != "repo" {
 			return nil, ErrInvalidPermissionParams
 		}
-		// TODO: maybe this should not be limited to a single "action" string?
+		// NOTE: currently actions are `read` and `manage`, and it makes no sense to use more than one as `manage` implies `read`. This might change in the future, if more action are added.
 		if len(g.Params["action"]) > 1 {
 			return nil, ErrInvalidPermissionParams
 		}


atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+		if p.Attribute != "*" && p.Attribute != "handle" {
+			return nil, ErrInvalidPermissionParams
+		}
+	case "include":
+		for k, _ := range g.Params {
+			if !(k == "nsid" || k == "aud") {
+				return nil, fmt.Errorf("%w: unsupported 'include' param: %s", ErrInvalidPermissionParams, k)
+			}
+		}
+		if g.Params.Has("nsid") {
```
matthieusieben, line 193: I'm not fluent in go: do we check that the positional and name attribute are not set together?

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+	case "blob":
+		if len(p.Accept) == 1 {
+			positional = p.Accept[0]
+		} else if len(p.Accept) > 1 {
+			params["accept"] = p.Accept
+		}
+	case "identity":
+		if p.Attribute != "" {
+			positional = p.Attribute
+		}
```
matthieusieben, line 60: Do we need to handle possibly invalid input values (permission) here ?

If yes: this should return an error if `p.Attribute` is empty
If no: then the `if` here could be omitted.

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+	params := make(url.Values)
+
+	switch p.Resource {
+	case "account":
+		if p.Attribute != "" {
+			positional = p.Attribute
+		}
+		if len(p.Action) != 0 {
+			params["action"] = p.Action
+		}
```
matthieusieben, line 50: Should always be true if the `permission` is a valid one (`action` is required to be non-empty).

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+		if p.Attribute != "" {
+			positional = p.Attribute
+		}
+	case "include":
+		if p.NSID != "" {
+			positional = p.NSID
+		}
+		if p.Audience != "" {
+			params.Set("aud", p.Audience)
+		}
```
matthieusieben, line 67: Strictly speaking, `include` is *not* a "permission", only a "scope" that is used to fetch extra permissions.

In the TS we represent this distinction in the class name (`IncludeScope` vs. `AccountPermission`), interface (`ResourcePermission`) and usage (`IncludeScope` has methods that allow to generate permissions from it, by combining it with a permission set document).

Should this distinction also exist in here ?

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+	default:
+		return ""
+	}
+
+	scope := p.Resource
+	if positional != "" {
+		scope = scope + ":" + positional
```
matthieusieben, line 92: TS uses URL encoding on the `positional` in order to ensure that no `?` can be present (as it would break parsing).

atproto/auth/permission_util.go
```diff
@@ -0,0 +1,95 @@
...
+func ParseGenericScope(scope string) (*GenericPermission, error) {
+
+	if !isASCII(scope) {
+		return nil, ErrInvalidPermissionSyntax
+	}
+
+	front, query, _ := strings.Cut(scope, "?")
+	resource, positional, _ := strings.Cut(front, ":")
+
+	// XXX: more charset restrictions
```
matthieusieben, line 28: TS performs URL decoding on the positional

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+		if g.Positional != "" {
+			p.Attribute = g.Positional
+		}
+		if p.Attribute == "" {
+			return nil, ErrInvalidPermissionParams
+		}
+		if p.Attribute != "" && p.Attribute != "email" && p.Attribute != "repo" {
+			return nil, ErrInvalidPermissionParams
+		}
+		// TODO: maybe this should not be limited to a single "action" string?
```
matthieusieben:
```suggestion
		// Currently actions are `read` and `manage` so it makes no sense to use more than one as `manage` implies `read`. This might change in the future, if more action are added.
```

atproto/auth/permission.go
```diff
@@ -0,0 +1,288 @@
...
+			}
+			p.Attribute = g.Params.Get("attr")
+		}
+		if g.Positional != "" {
+			p.Attribute = g.Positional
+		}
+		if p.Attribute != "*" && p.Attribute != "handle" {
+			return nil, ErrInvalidPermissionParams
+		}
+	case "include":
```
matthieusieben, line 187: Strictly speaking `include` is not a permission, only a scope.

atproto/auth/permission_util.go
```diff
@@ -0,0 +1,95 @@
...
+		Positional: positional,
+		Params:     params,
+	}
+	return &p, nil
+}
+
+func (p *GenericPermission) ScopeString() string {
+	scope := p.Resource
+	if p.Positional != "" {
+		scope = scope + ":" + p.Positional
```
matthieusieben, line 46: Pseudo code:
```suggestion
		scope = scope + ":" + urlencodethis(p.Positional)
```

atproto/auth/permission_util.go
```diff
@@ -0,0 +1,95 @@
...
+	parts := strings.SplitN(accept, "/", 3)
+	if len(parts) != 2 {
+		return false
+	}
+	if parts[0] == "*" {
+		return false
+	}
+	if parts[1] == "**" {
+		return false
+	}
```
matthieusieben: - This recognizes `foo/***` as a valid accept: `parts[1]` should either be exactly `*` or contain no `*` char at all.
- Empty parts should be rejected (`foo/` and `/foo`)

atproto/auth/permission.go
bnewbold: yep, the line right below this checks that; if `g.Params.Has("nsid")` is true and `g.Positional != ""`, then an error is returned

atproto/auth/permission.go
bnewbold: I felt it was idiomatic Go to leave it to calling code to assume that the `Permission` struct has reasonable values, and to not return an `error` when formatting as a scope string.

as to whether the `if` check is needed, I agree it is redundant, but at least to me feels like it improves readability? but I could go either way

atproto/auth/permission.go
bnewbold: I think that distinction would be fiddly to express in the Go type system; eg it is nice for `ParseOAuthScope` to return an array of a single struct type.

atproto/auth/permission.go
bnewbold: (discussed above)

atproto/auth/permission.go
bnewbold: doing this in the general case means `blob:image/*` gets round-tripped as `blob:image%2F%2A`, which seems undesirable.

the specs don't talk about encoding the "positional" part, only the "parameters" part. they say that the "positional" can include any characters except `?`: https://atproto.com/specs/permission#scope-string-syntax

the idiomatic Go thing to me here would be to check when parsing (when an error can be returned), but to assume that the struct was valid when creating the permission string. in this case, an extra `?` in the positional isn't really possible to detect in the general case: it will look like a query param separator.

atproto/auth/permission_util.go
bnewbold: (discussed above)

atproto/auth/permission_util.go
bnewbold: is the "no wildcard character ever allowed" part true in the general case? I don't remember how things work for the full mimetype/charset syntax.

bnewbold: added checks and tests for "empty" parts

Issue #1084, opened 2025-05-22
Walking `com.atproto.sync.listReposByCollection` with cursor fails


ZachOrr:
Hey y'all - a bit of an odd one.

I'm attempting to walk the `com.atproto.sync.listReposByCollection` for `app.bsky.actor.profile` (using the `relay1.us-east.bsky.network` instance) and I'm getting [an error response back](https://github.com/bluesky-social/indigo/blob/236dd575c91e4c7173e4cd23aff360180d88e42b/cmd/collectiondir/serve.go#L474) -

```
{"error":"DatabaseError","message":"failed to read DIDs for collection"}
```

The oddity here is I can repro this with Swift code, but I cannot repro this with Python code. I've got no idea what the difference is.

---

```python
import json
from urllib import parse, request

BASE_URL = "https://relay1.us-east.bsky.network/xrpc/com.atproto.sync.listReposByCollection"

cursor = None
while True:
    print(f"Fetching cursor: {cursor}")
    params = {
        "collection": "app.bsky.actor.profile",
        "limit": 1,
        "cursor": cursor
    }
    # Remove None values from params
    params = {k: v for k, v in params.items() if v is not None}
    url = f"{BASE_URL}?{parse.urlencode(params)}"
    with request.urlopen(url) as response:
        response_json = json.loads(response.read().decode())
    if len(response_json["repos"]) == 0:
        break
    cursor = response_json["cursor"]
```

```bash
$ uv run runme.py
Fetching cursor: None
Fetching cursor: QQAAAAYAAAGVsC0YgWRpZDpwbGM6NTV3dmJ4M2llNWYyZXRlcnhtM2dubnZuAA==
Fetching cursor: QQAAAAYAAAGVsC0Y5mRpZDpwbGM6cHZ1aGxzYnM3NDU2eW4zZ3ZpaWd4aHU0AA==
Fetching cursor: QQAAAAYAAAGVsC0iRmRpZDpwbGM6c2p0eXl4MmxnbXd4d2pydGltbDRlbzRxAA==
Fetching cursor: QQAAAAYAAAGVsC01zmRpZDpwbGM6NHp1ZzdrdzVscGlta3g1aXFnZ2NjZXpqAA==
Fetching cursor: QQAAAAYAAAGVsC1StmRpZDpwbGM6cmw3eGZyNmZ2bXUybHNhb3Bub3l2cmw0AA==
[...]

PR #1211, opened 2025-12-01
collectiondir: use base64url encoding for cursor param

bnewbold:
This should fix a corner-case bug with query param encoding with Swift.

Will need a deploy.

Closes: https://github.com/bluesky-social/indigo/issues/1084

### DIFF
--- a/cmd/collectiondir/pebble.go
+++ b/cmd/collectiondir/pebble.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/collectiondir/pebble.go
```diff
@@ -194,7 +194,7 @@ func (pcd *PebbleCollectionDirectory) ReadCollection(ctx context.Context, collec
 		return nil, "", fmt.Errorf("collection id err, %w", err)
 	}
 	if cursor != "" {
-		lower, err = base64.StdEncoding.DecodeString(cursor)
+		lower, err = base64.RawURLEncoding.DecodeString(cursor)
```
damz: Note that these are not compatible with each other: cursors encoded with `base64.StdEncoding` (`+` and `/` for characters 63 and 64, with `=` padding) will not decode with `base64.RawURLEncoding` (which expects `-` and `_` for characters 63 and 64 and will error out seeing the padding character).

If this lack of backwards compatibility is an issue, maybe implement an utility function that can decode old-style tokens (an easy way to discriminate would be to use the presence of `+`, `/` or `=` as a signal to use the old-style decoder).

bnewbold:
(this has been deployed)

PR #1214, opened 2025-12-01
relay: enforce initial seq as a minimum

bnewbold:


cmd/relay/stream/persist/diskpersist/diskpersist.go
```diff
@@ -214,6 +214,11 @@ func (dp *DiskPersistence) resumeLog() error {
 
 	dp.log.Info("loaded seq", "seq", seq, "now", time.Now().UnixMicro())
 
+	if seq < dp.initialSeq {
+		dp.log.Warn("forcing higher seq", "found", seq, "initialSeq", dp.initialSeq)
+		seq = dp.initialSeq
+	}
```
damz, line 220: If these are the desired semantics, maybe rename the flag and update the documentation?

The current flag is documented in several places as applying only if there is no persisted data, notably:

* https://github.com/bluesky-social/indigo/blob/81f317e322bcb558136ad9a6bd729032e3ce56a1/cmd/relay/stream/persist/diskpersist/diskpersist.go#L82-L83
* https://github.com/bluesky-social/indigo/blob/81f317e322bcb558136ad9a6bd729032e3ce56a1/cmd/relay/main.go#L141

bnewbold: Good catch, updated in those places. I didn't change the name of the flag itself.

Issue #1161, opened 2025-09-30
Relay: `active` state gets desynchronized, repos blocked


uniphil:
Right now there's nothing in the spec for eventual consistency of repo `active` state stored by middle network services, so (perhaps unsurprisingly) it *doesn't* stay synchronized in practice in `relay`. Maybe surprisingly it seems to happen a lot.

The current relay behaviour is to ignore events from accounts that the relay thinks are not active, so anyone impacted gets effectively just banned from that relay.

It seems to me like this could be fixed in may possible places

- the spec could fix this by adopting some kind of eventual consistency strategy for repo `active` state.
- PDSs could fix this by re-emitting their `active` repo states periodically (gross?)
- Relay could fix this by polling upstream if it sees unexpected commits from a repo it things is not `active` every so often

and probably more.

---

Original research showing Bluesky's sync1.1 relays accidentally banning 7% of self-hosters (12% for my relays eesh): https://bsky.app/profile/bad-example.com/post/3lwzghbs6322x . Script for running the scan included in that thread if you want to repro.

Live demo to explicitly check status for accounts followed: https://compare.hose.cam/?deactivated

uniphil:
related to but different from #1143

bnewbold:
Thanks for raising! This is indeed a bad situation and one of the blockers for us switching over completely to the "new" relay implementation for the primary endpoint (and the bsky appviews).

The hope was that this situation would not be encountered during migration because the relay also ignores messages coming from upstream hosts (PDS) which are not the host indicated in the account's DID document. Eg, if an account migrates, i'd assume that account and identity events come out of the new PDS, and the relay sees those, and starts ignoring the old host. Then the account on the old host is deactivated, but that account event is ignored because it is no longer the active host. I'm not sure if there is a complete bug in this expected behavior, or a race-condition, or what.

A couple high-level thoughts:

- we have an expected behavior where if a commit signature is invalid, the receiver refreshes their identity cache to see if there is a new key, before they discard the event. we could do something similar with received commit and sync events when an account is inactive due to "upstream" status: verify the account hosting status before dropping the message. this would probably need some kind of caching window/timing
- the client read-after-write mechanism is in theory a way to catch this sort of situation: appview response proxied by the PDS can include a header describing the last repo rev the appview has seen. if the client (or PDS) sees that this is pretty far out of date, it could flag that to the user or try to intervene in some way. this is not a solution, just a signal.
[...]

PR #1215, opened 2025-12-02
relay: iterate on verification of account host and status

bnewbold:
This is intended to fix the stuck account status described in: https://github.com/bluesky-social/indigo/issues/1161

The previous logic was:

- all events were checked against cached hosting location, and event dropped if not from expected account PDS
- if an account had upstream inactive status, that status would persist until a new `#account` event was emitted
- if an event is received from a host, and that host doesn't match the cached account location (host), code will check if the account location needs to be updated

The new logic:

- identity events immediate purge the cache for that account, and generally pass-through for any account from any host. This is to ensure that identity events always "get through" and purge caches, especially if they are the first event from "new" PDS during a migration
- if an event is received from a host, and the host doesn't match, the identity cache will be purged and the host checked again with the newly resolved identity
- if an event is received from the correct host, and the account has inactive status upstream, the relay will check for the current account status on the PDS host to check the upstream status (and will update the database/cache if needed) 

It is still the case that:

- account host information never gets created or updated unless an event is received from the current active host. that is, an event from the "old" PDS host during a migration is not sufficient; the "new" PDS host needs to emit an event (any event will do) and have it received by the relay. This is partially to prevent accounts from claiming to be hosted on a PDS which they are not (bi-directional verification)
- an "old" PDS host could emit an `#account` event marking an account as inactive. If there was no `#identity` event processed before this, this could result in the account status getting updated. however, the new logic should make it much more likely for the account status to "converge", and for that to happen without events getting dropped
- an account can get "stuck" active at the relay. If a PDS marked an account inactive (eg, takendown), but the relay never received that event (dropped, bug, or connected later), then it will have active account status at the relay, and will pass-through events. This is a pretty corner-case situation: a PDS presumably should not be sending such events if the account is inactive.

cc: @uniphil

### DIFF
--- a/cmd/relay/relay/ingest.go
+++ b/cmd/relay/relay/ingest.go
@@ -210,9 +210,14 @@ func (r *Relay) processIdentityEvent(ctx context.Context, evt *comatproto.SyncSu
 
 	// optimistically process event. might create a row in account table if this event came from the current host
 	handle := evt.Handle
-	_, _, err = r.preProcessEvent(ctx, evt.Did, hostname, hostID, logger)
+	_, ident, err := r.preProcessEvent(ctx, evt.Did, hostname, hostID, logger)
 	if err != nil {
 		// don't pass-through handle if there was a problem with event (eg, account on another host, or inactive status)
+		handle = nil
+	}
+
+	// check that handle at least matches that in the DID document (if available)
+	if ident != nil && handle != nil && ident.Handle.String() != *handle {
 		handle = nil
 	}
 


cmd/relay/relay/account.go
```diff
@@ -182,13 +193,51 @@ func (r *Relay) EnsureAccountHost(ctx context.Context, acc *models.Account, host
...
+		return err
+	}
+
+	if acc.UpstreamStatus != status {
+		// NOTE: r.UpdateAccountUpstreamStatus purges account cache
+		if err := r.UpdateAccountUpstreamStatus(ctx, did, acc.UID, status); err != nil {
+			return err
+		}
+		acc.UpstreamStatus = status
+	}
```
uniphil, line 226: yayyyyy

cmd/relay/relay/ingest.go
```diff
@@ -197,26 +195,32 @@ func (r *Relay) processIdentityEvent(ctx context.Context, evt *comatproto.SyncSu
...
+
+	// NOTE: not doing other validation or processing here; eg not strictly checking account host mapping or account status. Any PDS host can emit an identity event for any account, and it will be passed through
 
 	// Broadcast the identity event to all consumers
 	err = r.Events.AddEvent(ctx, &stream.XRPCStreamEvent{
 		RepoIdentity: &comatproto.SyncSubscribeRepos_Identity{
 			Did:    did.String(),
 			Time:   evt.Time,   // TODO: update to now?
-			Handle: evt.Handle, // TODO: we could substitute in our handle resolution here
+			Handle: evt.Handle, // TODO: we could substitute in our own handle resolution here
```
uniphil: is there a slightly increased risk from broadcasting the unmodified handle, now that forwarding `#identity` events is more permissive?

seems plausible that some downstream consumers might accidentally assume the `handle` field is trustworthy, opening themselves up to handle ownership confusion, with wider impact after this change. (it's not ambiguous in the spec but it's not called out with a specific warning not to trust it, not everyone reads specs, etc)

getting outside the scope of this PR, but it's also not clear to me why the handle is included with `#identity` at all. my guesses for the intent:

- to help detect redundant identity events, to filter/ignore them?
- to allow optimistically starting both halves of bidirectional re-validation at once?

i wonder if it might be safer to either
- just drop the handle always, or
- drop the handle unless the host is the same one we currently have cached

bnewbold: Yup, I agonized over this a bit and agree with what you are saying here. A bad actor PDS could blast out identity events with incorrect handles for any account in the network.

I'm pretty opinionated that the relay should not be doing handle resolution, at least to start. Doing handle resolution at scale has caused a number of performance problems during incidents in the past (it is fine when everything is fine, but causes pain when needing to catch up, or under flakey network conditions).

What I had missed earlier was that the `handle` field on `#identity` events is simply optional. I thought it was required and i'd have to stick `handle.invalid` in there to strip the value. What should be possible is to capture the `err` from `preProcessEvent`. If `err` is nil, that implies the event is probably coming from the expected host and the account is active, and we can keep the handle on the event. If it is not-nil, the problem could in theory be many things, but likely is one of those conditions failing, and we can remove the handle before re-broadcasting.

Will wait for @dholms to weigh in.

cmd/relay/relay/account.go
```diff
@@ -182,13 +193,51 @@ func (r *Relay) EnsureAccountHost(ctx context.Context, acc *models.Account, host
...
 
+// Checks that account is active. If it is not active upstream, double-checks with the host and updates status if needed; otherwise returns an error.
+func (r *Relay) EnsureAccountActive(ctx context.Context, acc *models.Account) error {
+
+	// short-circuit in common case
+	if acc.IsActive() {
+		return nil
+	}
+	// NOTE: this is checking local takedown
+	if acc.Status != models.AccountStatusActive {
```
dholms, line 204: I believe this is already checked in `acc.IsActive()`

cmd/relay/relay/ingest.go
dholms: We went back and forth on if the handle should even be on the identity event or not. I think we decided to keep it because it can be used as an optimistic check to see if full handle revalidation needs to occur. 

I think I'd be in favor of only populating handle if it matches the purported handle in the DID doc. This means that you do not need to do full handle resolution but will at least prevent a bunch of misleading events. 

At the very least I think we probably shouldn't pass along the handle value from non-canonical hosts.

dholms (approved):
lookin good 💎

cmd/relay/relay/account.go
bnewbold: `acc.IsActive()` checks both "local" status (`acc.Status`: eg, takedown on the relay itself) and "upstream" status on the PDS (`acc.UpstreamStatus`).

Here we are untangling upstream vs local status.

cmd/relay/relay/ingest.go
bnewbold: oh, good call to check against the DID document (which are are resolving). Added that.

bnewbold:
(this has been deployed to `relay1.us-west.bsky.network` and `relay1.us-east.bsky.network`)

PR #1217, opened 2025-12-02
relay: remove SSRF protections for insecure localhost connections

### LINKED ISSUE #1216: Relay refuses to connect to new localhost PDS over http
If with `serve --allow-insecure-hosts`, relay will ultimately refuse to invoke `/xrpc/com.atproto.server.describeServer` on a localhost PDS because is over http.

The issue appears to be that `HostChecker.CheckHost` does not respect the supplied flag.

bnewbold:
Closes: https://github.com/bluesky-social/indigo/issues/1216


cmd/relay/handlers.go
```diff
@@ -54,7 +54,14 @@ func (s *Service) handleComAtprotoSyncRequestCrawl(c echo.Context, body *comatpr
 		hostURL = "http://" + hostname
 	}
 
-	if err := s.relay.HostChecker.CheckHost(ctx, hostURL); err != nil {
+	if !s.config.AllowInsecureHosts {
+		err = s.relay.HostChecker.CheckHost(ctx, hostURL)
+	} else {
+		hc := relay.NewHostClient("")
```
damz, line 58: It sounds simpler to set `hc.Client = http.DefaultClient` in `NewRelay()` if `config.AllowInsecureHosts` is set.

bnewbold: That would impact all use of the HostChecker though. The current setup only allows the bypass when doing the request using admin authentication.

I flipped the ordering and added another `admin` boolean check to make it clearer to read and stricter in implementation.

damz: According to the code intelligence, `relay.HostChecker` is literally only used on that line, do you know of other users?

damz: Oh, I see. Yeah, with the additional check on `admin` it makes sense :+1:

damz: But maybe use `NewHostClient(s.config.UserAgent)` like in `NewRelay()`?

bnewbold: This PR starts using some of the *account* checking behavior: https://github.com/bluesky-social/indigo/pull/1215. I guess that will probably not work with localhost stuff now that I think of it.

This is honestly a big messy tangle, I kind of want to just be secure all the time and not support the insecure/localhost mode at all.

ThisIsMissEm:
The way I thought of handling this was like the safeFetch implementation in the atproto repo.

vmx:
Can this be merged please? It took my quite a while to find out what the problem was and then found this PR.

I can confirm that I can now add a local PDS via:

```
curl -X POST http://localhost:2470/admin/pds/requestCraw
  -H "Content-Type: application/json" \
  -u "admin:localdev" \
  -d '{"hostname": "localhost:2583"}'
```

And see new records arriving via:

    websocat 'ws://localhost:2470/xrpc/com.atproto.sync.subscribeRepos'

PR #1218, opened 2025-12-03
`syntax.AtIdentifier` as simple string

bnewbold:
This was a clever struct wrapped, but that was weird and inconsistent with all the other types.

This updates it to be a simple string alias type.

This is (slightly) breaking, and will require updates in downstream packages. Particularly those doing OAuth.


jcalabro (approved):
Alright, I don't feel super strongly. I think it's a relatively small thing, but unfortunate

bnewbold:
def appreciate the concern, and this is a bit borderline/grey to me.

I asked in-app (https://bsky.app/profile/bnewbold.net/post/3mcutnvk3xs2d) and in discord in case it was going to be rough on anybody and got fairly positive "go ahead" responses.

PR #1220, opened 2025-12-05
syntax: clarify behaviors around year zero

bnewbold:
See: https://github.com/bluesky-social/atproto-website/pull/479


atproto/syntax/testdata/datetime_parse_invalid.txt
```diff
@@ -5,3 +5,6 @@
 1985-04-12T25:20:50.123Z
 1985-04-12T23:99:50.123Z
 1985-04-12T23:20:61.123Z
+
+# ISO-8601, but normalizes to a negative time
```
ThisIsMissEm, line 9: Perhaps specifically call out that the timezone when converted to UTC results in a date before year zero?

There are of course use-cases for pre-year zero like registering events pre year zero like the building of The pyramids of geological events, but for those you should define an application specifically date type that explicitly handles distant historical events & serializing them correctly.

matthieusieben (approved):

PR #1234, opened 2025-12-17
Tap: allow custom plc directory

boltlessengineer:
One might need custom plc directory endpoint other than <https://plc.directory> for testing purposes or for plc mirrors.

As tap supports full network backfill, it would be better to test that on sandboxed atproto infra than public infra with ~40M repos.

So I added a `plc-url` flag and `TAP_PLC_URL` env var.
Same as relay, `ATP_PLC_HOST` env var can also be used instead.


bnewbold (approved):
seems reasonable to me

dholms (approved):
This looks great. Thanks!

PR #1241, opened 2025-12-22
Tap Go Library

jcalabro:
This complements #1170 and https://github.com/bluesky-social/atproto/pull/4290 by creating a new go library for working with the `tap` websocket.

This explicitly does not create HTTP handler helpers, but the `tap.Event` type can be used to support the plain HTTP request/response paradigm supported by `tap`.

It's pretty important that the  websocket reconnection and message retry logic be robust, if ever want to say replace inferno with this library, so I went ahead and did what feels like over-engineering, but I don't actually think it is. I would not want to ship this to our prod appview backend without very robust retry logic that mimics what we already do internally (infinite retry on transient processing failures).

There are mocked tests here, and I'm using this library in a new system I'm working on and it appears to be working well (not public yet, but it will be! :eyes:)


tap/event.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	default:
+		return fmt.Errorf("unknown event type %q", event.Type)
+	}
+
+	return nil
+}
+
+func (e Event) MarshalJSON() ([]byte, error) {
+	event := struct {
```
dholms, line 76: Can we use the `Event` type here?

tap/websocket.go
```diff
@@ -0,0 +1,320 @@
...
+
+	addr     string
+	sendAcks bool
+	maxErrs  int
+
+	connectTimeout time.Duration
+	readTimeout    time.Duration
+	writeTimeout   time.Duration
+
+	handler WebsocketHandlerFunc
```
dholms, line 50: Something I wondered when working with our firehose processing code is why do we use callbacks/handlers? These seems like a pretty natural use case for channels.

In this particular case, it seems like we miss out on parallel processing of events since we call the handler serially

dholms (commented):
Nice this looks good to me! Thanks for putting this together.

Before merging, we should add at least a barebones README. It could just have the example from `doc.go` in it. But more importantly, it should link to `./cmd/tap` which has the fuller documentation for tap. 

I can just see devs looking for tap, coming across this first (since it's in the root dir) and not knowing what to do with it. Actually to that end, maybe we rename this package `tap-client` or similar?

tap/websocket.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+var (
+	initialBackoff = 500 * time.Millisecond
+)
+
+// A thin error wrapper that indicates to the tap client consumer loop that a message
+// should not be retried (i.e. invalid user input that will surely fail again on retry).
+type NonRetryableError struct {
+	err error
```
bnewbold, line 24: why not expose the inner error? I feel like there is probably an idiom around this sort of error wrapper which i'm not remembering.

tap/websocket.go
bnewbold: I'm also interested in channel-based APIs for stream processing. there are some complexities:

- error and ACK handling can be simple with callbacks
- you lose ordering guarantees (eg, per DID) unless you add some sharding or locking mechanism (eg, like the parallel scheduler jaz implemented for firehose)
- might be some perf overhead around GC? but you get concurrency which seems like it should be an overall win

I know that why generally avoids channel APIs, and I think that is the reason we don't have many. I think we probably want to keep a callback/handler API for most of our packages; i'd just *also* like the option to use channels. message queing packages (kafka, nsq, nats, etc) have good examples.

tap/websocket.go
```diff
@@ -0,0 +1,320 @@
...
+		ws.sendAcks = true
+	}
+}
+
+// Defines an option for the tap websocket consumer. A nil error indicates that an ACK will be sent to tap
+// if WithAcks() is provided.
+type WebsocketHandlerFunc func(context.Context, *Event) error
+
+// Initializes a tap websocket consumer
+func NewWebsocket(addr string, handler WebsocketHandlerFunc, opts ...WebsocketOption) (*Websocket, error) {
```
bnewbold, line 109: I don't love `Websocket` as the noun in these APIs. `Consumer`? `Channel`?

tap/websocket.go
```diff
@@ -0,0 +1,320 @@
...
+	return func(ws *Websocket) {
+		ws.sendAcks = true
+	}
+}
+
+// Defines an option for the tap websocket consumer. A nil error indicates that an ACK will be sent to tap
+// if WithAcks() is provided.
+type WebsocketHandlerFunc func(context.Context, *Event) error
+
+// Initializes a tap websocket consumer
```
bnewbold, line 108: doc comment should explain that `addr` is expected to be a `wss://` URL. naming it `websocketURL` might also help

tap/websocket.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(ws)
+	}
+
+	return ws, nil
+}
+
+// Connects to and beings the main tap websocket consumer loop
```
bnewbold, line 146: "begins"

bnewbold (commented):
thanks for hacking on this!

I think especially for a package like this (vs a daemon), it is good/fine to do just doc.go, which renders nicely on pkg.go.dev. Might be worth a small README which just links to that, I think the current doc.go could be improved though: link to the `tap` command README, or even the announcement blog post. Have a section talking about how to do clean shutdown. Maybe talk about some design properties, like how backpressure is expected to work, what kind of performance (evt/sec) is expected with/without ACKs, etc.

I'm not super in to having the Makefile build the binaries in each cmd directory. I'm open to rethinking how it works, but I don't think the current setup is obviously wrong, and think we should discuss it in a separate PR not here.

I think the module should be named `tapclient`, not `tap`. I'm kind of ambivalent about all the options of where to put it: top level is fine ok but gets busy; `cmd/tap/tapclient` isn't idiomatic or accessible but feels a bit tidier; maybe `pkg/tapclient` but if we do that we should be more consistent. top-level is probably best for now.

I'd be curious to iterate on this more together, but also think this is good and helpful as is and don't think we need to block merging.

Issue #1239, opened 2025-12-20
Tap: resyncer doesn't honor its own backoff


damz:
In `claimResyncJob()`, the resyncer doesn't honor the backoff it has set in the `retry_after` field:

https://github.com/bluesky-social/indigo/blob/034a2c019e641b355481b177e5696a12120edd43/cmd/tap/resyncer.go#L76-L85

Because of the way SQLite executes this query (in order of state, retry_after because it follows the `idx_repos_state_retry` index), it tends to immediately re-pick jobs that have failed (because "desynchronized" sorts before "pending"), resulting in a tight retry loop in case a PDS returned a rate limit error.

PR #1247, opened 2025-12-23
Tap: respect backoff in resyncer

dholms:
Resyncer was setting `retry_after` but then not actually respecting it when claiming a job 

Closes https://github.com/bluesky-social/indigo/issues/1239

### DIFF
--- a/cmd/tap/resyncer.go
+++ b/cmd/tap/resyncer.go
@@ -73,16 +73,18 @@ func (r *Resyncer) claimResyncJob(ctx context.Context) (string, bool, error) {
 	defer r.claimJobMu.Unlock()
 
 	var did string
+	now := time.Now().Unix()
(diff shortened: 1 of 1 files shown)


dholms:
No this is unrelated. This is about retrying resyncs. For instance, if you discover a new repo and have to fetch the whole thing (`sync.getRepo`) and that fails, then this will cause it to wait some amount of time then retry the fetch.

I responded on Bluesky to that thread but the short of it is there are particular delivery guarantees that Tap provides. If you sync a whole repo (with `sync.getRepo`) then those events are all considered to be created "at the same time" (and tracked with the same rev) so they're all delivered concurrently and ordering is not enforced across them. Live events off the firehose have ordering enforced on a per-repo basis.

You can read more on the exact ordering guarantees here: https://github.com/bluesky-social/indigo/tree/main/cmd/tap#per-repo-ordering-rules

PR #1254, opened 2025-12-27
tap: add kafka mode

haileyok:
With this set of changes, I was able to get from ~6k evt/s over the websocket to ~ 9k evt/s when using SQLite, and roughly 17k evt/s when using Postgres + tuning. this might not be a thing for merging, but showing an example of how we can increase the backfill speed significantly under the right conditions...


cmd/tap/tap.go
```diff
@@ -117,9 +120,29 @@ func NewTap(config TapConfig) (*Tap, error) {
...
+		var err error
+		// the producer from atkafka is the same as https://github.com/bluesky-social/go-util, except it allows for producing JSON rather than
+		// protobufs, which is what we want here
+		kafkaProducer, err = atkafka.NewProducer(
+			context.TODO(),
+			logger.With("component", "kafkaProducer"),
+			config.KafkaBootstrapServers,
+			config.KafkaOutputTopic,
+			atkafka.WithEnsureTopic(true),   // ensures that the topic has been created
+			atkafka.WithTopicPartitions(24), // TODO: partition count env var
```
haileyok, line 136: these should actually be configurable, probably

cmd/tap/resyncer.go
```diff
@@ -148,11 +148,10 @@ func (r *Resyncer) doResync(ctx context.Context, did string) (bool, error) {
...
 
-	r.pdsBackoffMu.RLock()
-	backoffUntil, inBackoff := r.pdsBackoff[pdsURL]
-	r.pdsBackoffMu.RUnlock()
-	if inBackoff && time.Now().Before(backoffUntil) {
-		return false, nil
+	if backoffUntil, ok := r.pdsBackoff.Load(pdsURL); ok {
+		if time.Now().Before(backoffUntil) {
+			return false, nil
+		}
```
haileyok, line 154: since we are almost exclusively doing reads out of this map, swapping it out with sync.Map instead of that mutex

PR #1257, opened 2025-12-30
Tap: restore nextID counter on restart to prevent unique constraint errors

### LINKED ISSUE #1253: Tap: fails with `UNIQUE constraint failed: outbox_buffers.id` after restart
When using `tap`, a `UNIQUE constraint failed: outbox_buffers.id` error occurs after restarting the process.

### Steps to Reproduce
1. Start tap using:
   ```sh
   ./tap run
   ```
2. Add a specific repository via `repos/add`.
3. Stop tap.
4. Start tap again.
5. Add a record to the repository.

Then the following error occurs:
```
{"time":"2025-12-26T10:23:08.436143942Z","level":"INFO","msg":"retrying timed out event","system":"tap","component":"outbox","id":405}
{"time":"2025-12-26T10:23:10.207663188Z","level":"ERROR","msg":"failed to update repo state","system":"tap","component":"firehose","did":"did:plc:myrepodid","rev":"myreporev","error":"UNIQUE constraint failed: outbox_buffers.id"}
{"time":"2025-12-26T10:23:10.207785689Z","level":"ERROR","msg":"event handler failed","system":"parallel-scheduler","err":"UNIQUE constraint failed: outbox_buffers.id"}
```

### Workaround
it seemed that `nextID` was not being restored from the last loaded event during `EventManager` startup, and restoring it avoided the error.
Not sure if this is the best fix, I tried the following small change and the issue no longer reproduced for me:

https://github.com/bluesky-social/indigo/blob/598fbf0e146e26a0121abc89ecff1114c4630c4c/cmd/tap/event_manager.go#L114-L148
```go
eventCacheSize.Set(float64(len(em.cache)))
if len(dbEvts) > 0 {
    maxId := dbEvts[len(dbEvts)-1].ID
    em.nextID.Store(uint64(maxId + 1))
}
```

heathdutton:
When Tap restarts, the EventManager loads existing events from the database but doesn't initialize the nextID counter. This causes new events to be assigned IDs starting from 0, which collide with existing database records.

The fix restores the nextID counter as events are loaded, ensuring new IDs are always greater than any existing database IDs.

Fixes #1253

### DIFF
--- a/cmd/tap/event_manager.go
+++ b/cmd/tap/event_manager.go
@@ -140,6 +140,9 @@ func (em *EventManager) loadEventPage(ctx context.Context, lastID int) (int, err
 	eventCacheSize.Set(float64(len(em.cache)))
 	em.cacheLk.Unlock()
 
+	maxID := dbEvts[len(dbEvts)-1].ID
+	em.nextID.Store(uint64(maxID + 1))
+
 	for i := range dbEvts {
 		em.pendingIDs <- dbEvts[i].ID
 	}


dholms (approved):
Good find. Thank you!

PR #1262, opened 2026-01-04
Tap: Expose repo status via getRepoInfo

ThisIsMissEm:
I noticed I'd accidentally been storing the repository state in the `status` field on my Account model, as identity events give you a status field, but if you need to ensure you have a Account record for a given DID, you probably need to call the getRepoInfo endpoint, which returns a "state" field, which I mixed up with "status"

As the AccountStatus is indeed on the Repo model, I think it makes sense to return it here?


ThisIsMissEm:
I went with status for parity with identity events, but I agree in both cases account_status would be better

PR #1273, opened 2026-01-19
pds: remove duplicated import

davecheney:
github.com/bluesky-social/indigo/api/atproto was imported twice, the second time under an alias. This commit removes the duplicated import.


bnewbold (approved):
thanks!

in the mid-term i'd like to get the `api/atproto` package renamed to `api/comatproto` to be more atproto-idiomatic (at some go-idiomatic cost), but for now definitely better not to have duplicate imports!

PR #1278, opened 2026-01-22
atproto/atclient: password auth refresh should use method post

karitham:
Fix `PasswordAuth#Refresh` using the wrong method.


bnewbold (approved):
oof, good catch thanks

PR #1288, opened 2026-01-27
cmd/tap: move ownership of ^C'able context outside runTap

davecheney:
Rather than making runTap responsible for wrapping the context.Context passsed to it in a signal handler, push that responsibility to the caller as SIGTERM'ness is really a property of the process wrapped around tap, rather than runTap itself managing its own lifecycle. The logic is now runTap is pass a parent context, when that context is done, everyone shuts down. The why it was cancelled is not something runTap has to manage.

### DIFF
--- a/cmd/tap/main.go
+++ b/cmd/tap/main.go
@@ -26,6 +26,9 @@ func main() {
 }
 
 func run(args []string) error {
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
 	app := &cli.Command{
 		Name:    "tap",
 		Usage:   "atproto sync tool",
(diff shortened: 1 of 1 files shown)


cmd/tap/main.go
```diff
@@ -26,6 +26,10 @@ func main() {
 }
 
 func run(args []string) error {
+	ctx := context.Background()
+	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
```
jcalabro: very, very small nit, but we can avoid a variable shadow with

```go
ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```

davecheney: even better, will do.

PR #1289, opened 2026-01-27
cmd/tap: exit runResyncWorker when parent context is cancelled

davecheney:
runResyncWorker takes a context for lifecycle management, however if that context is cancelled -- err != nil -- then the rsync worker will sleep for a second, hit WaitForReady, which returns immediately because the context is cancelled, then claimResyncJob will also return immediately with context.Cancelled, err still not nil, so the worker sleeps for a second and tries again, etc.

Fix the loop by explicitly checking for the context.Cancelled case and exiting the worker.

No test yet, mia culpa.

### DIFF
--- a/cmd/tap/resyncer.go
+++ b/cmd/tap/resyncer.go
@@ -49,22 +49,34 @@ func (r *Resyncer) runResyncWorker(ctx context.Context, workerID int) {
 
 	for {
 		r.events.WaitForReady(ctx)
-		did, found, err := r.claimResyncJob(ctx)
-		if err != nil {
-			logger.Error("failed to claim resync job", "error", err)
-			time.Sleep(time.Second)
-			continue
-		}
 
-		if !found {
-			time.Sleep(time.Second)
-			continue
-		}
+		// Before calling claimResyncJob, check for context cancellation.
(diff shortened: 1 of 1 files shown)


cmd/tap/resyncer.go
```diff
@@ -49,22 +49,34 @@ func (r *Resyncer) runResyncWorker(ctx context.Context, workerID int) {
...
+				time.Sleep(time.Second)
+				continue
+			}
 
-		logger.Info("processing resync", "did", did)
-		err = r.resyncDid(ctx, did)
-		if err != nil {
-			logger.Error("resync failed", "did", did, "error", err)
+			if !found {
+				time.Sleep(time.Second)
```
jcalabro, line 71: not related to this PR obviously, but I feel like this sleep should be a bit more intelligent

davecheney: Yes, this is something that i was wondering. The "sleep and try again" pattern isn't that common inside tap, especially in this case where there is a "wait for a thing" right before it.

I think I spotted a backoff timer in the indigo repo so I'll try to find that again and look to integrating it so at least there is backoff with resonable jitter

PR #1292, opened 2026-01-27
cmd/tap: split tap service from command

davecheney:
This PR explores seperating the cmd/tap, the binary that holds main(), from tap, the service (or perhaps component if one was so keen as to run tap as part of a larger application).

Some plusses; `main.NewTap` becomes `tap.New()`, `main.TapConfig` becomes `tap.Config`, which is always pleasing to see. The downside is the shadowing of `tap, err := tap.New(...)`, which is perfectly cromulent, but the shadowing may not be to everyone's taste. This could be resolved by `t, err := tap.New(...)` if you're ok with the brevity, or `theTap, err := tap.New(...)` if you feel like pushing the boat out.

This also gives names to the other tap components, eg, `tap.Crawler` which is a hint that it may not belong in the tap package. Another clue is tap.Crawler (crawler.go) doesn't import any of the usual suspects like `../tap/models`. This is something to explore in the future.

Not really for review, more for discussion


.gitignore
```diff
@@ -37,7 +37,6 @@ test-coverage.out
 /sonar-cli
 /stress
 /supercollider
-/tap
```
jcalabro, line 38: why remove this in its current form?

jcalabro (approved):
one question, otherwise I'm down with the pattern

I was heavily using `./internal` in https://github.com/bluesky-social/indigo/tree/jc/cask

.gitignore
davecheney: oops, sorry, i'll revert that. before I understood that builds dumped into the root of the project i was scratching my head thinking "why is git being so obstinent'

PR #1293, opened 2026-01-27
gosky: refactor identity code

bnewbold:
This removes usage of the 'did' and 'handles' packages for resolution in gosky. The motivation is to clear the way for those packages to be fully removed.

The 'plc' package is still used for the command to create a PLC identity. Replacing that will require importing go-didplc, and some code to support reading secret keys from JWK format.

### DIFF
--- a/cmd/gosky/debug.go
+++ b/cmd/gosky/debug.go
@@ -20,7 +20,6 @@ import (
 	"github.com/bluesky-social/indigo/api/bsky"
 	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/atproto/syntax"
-	"github.com/bluesky-social/indigo/did"
 	"github.com/bluesky-social/indigo/events"
 	"github.com/bluesky-social/indigo/events/schedulers/sequential"
 	lexutil "github.com/bluesky-social/indigo/lex/util"
@@ -469,7 +468,7 @@ var debugFeedGenCmd = &cli.Command{
 			return err
 		}
 
-		didr := cliutil.GetDidResolver(cctx)
+		dir := identity.DefaultDirectory()
 
 		uri := cctx.Args().First()
 		puri, err := util.ParseAtUri(uri)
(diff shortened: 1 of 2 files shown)


cmd/gosky/admin.go
```diff
@@ -489,19 +490,21 @@ var enableInvitesCmd = &cli.Command{
...
-		handle := cctx.Args().First()
-		if !strings.HasPrefix(handle, "did:") {
-			phr := &handles.ProdHandleResolver{}
-			resp, err := phr.ResolveHandleToDid(ctx, handle)
+		username, err := syntax.ParseAtIdentifier(cctx.Args().First())
+		if err != nil {
+			return fmt.Errorf("must provide a valid DID or handle")
+		}
+		if username.IsHandle() {
+			bdir := identity.BaseDirectory{}
```
davecheney, line 498: creating a zero value identify.BaseDirectory to call ResolveHandle on it might be a suggestion that an indentity.ResolveHandle(...) is called for

cmd/gosky/admin.go
```diff
@@ -698,16 +703,22 @@ var queryModerationStatusesCmd = &cli.Command{
 		adminKey := cctx.String("admin-password")
 		xrpcc.AdminToken = &adminKey
 
-		did := cctx.Args().First()
-		if !strings.HasPrefix(did, "did:") {
-			phr := &handles.ProdHandleResolver{}
-			resp, err := phr.ResolveHandleToDid(ctx, did)
+		username, err := syntax.ParseAtIdentifier(cctx.Args().First())
```
davecheney, line 706: possible future suggestion, this could be wrapped up in a helper.

did, err := <somepackage>.Resolve(did/handle)

as this pattern has ben repeated a few times.

davecheney (approved):
LGTM with some weakly held suggestions for possible future cleanups.

cmd/gosky/admin.go
bnewbold: Yeah, there is a bit of type/interface/caching friction in the 'identity' package from how it evolved. The `Lookup*` methods try to force/ensure that handle resolution happens bi-directionally, because that is needed in most app code and easy to forget.

But the identity package has mostly been used for infra services, where the `Resolve*` methods are appropriate.

cmd/gosky/admin.go
bnewbold: Yup, maybe `ResolveUsername` or `ResolveIdentifierString` which goes from raw string to DID doc. And something similar for `Lookup()`

PR #1295, opened 2026-01-27
[draft] cmd/tap: use internal/group for managing the various goroutine components of a tap instance

davecheney:
This is a worked example of switching from using `go` to a group like structure to manage the lifecycle of the various workers inside a tap instance.

Because of the way some components, kick off their own child goroutines, this required stacking groups -- not something I'd done before -- but worked surprisingly well. I think the size of the change could be reduced by a preparatory PR which moves more of the worker creation into tap.Run.


jcalabro (commented):
I like! Can we just import github.com/pkg/group rather than copy+pasting though?

PR #1298, opened 2026-01-28
all: do not initalize mutex zero value

davecheney:
sync.Mutex and RWMutex leverage the zero value such that their uninitalised state is a valid mutex. Thus we can avoid initialising them during construction.

Additionally, a sync.Mutex's zero value is valid, whereas a *sync.Mutex _needs_ to be initialised to avoid an nil ptr panic. In extremely rare cases a field may be *sync.Mutex if it is sharing that mutex with something else, but this is not the case in terms of MockDirectory, so the initialisation can be simplified.

### DIFF
--- a/atproto/identity/mock_directory.go
+++ b/atproto/identity/mock_directory.go
@@ -12,8 +12,8 @@ import (
 // A fake identity directory, for use in tests
 type MockDirectory struct {
 	mu         sync.RWMutex
-	Handles    map[syntax.Handle]syntax.DID
-	Identities map[syntax.DID]Identity
+	handles    map[syntax.Handle]syntax.DID
+	identities map[syntax.DID]Identity
 }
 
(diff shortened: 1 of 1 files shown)


cmd/relay/relay/relay.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/bluesky-social/indigo/cmd/relay/stream/eventmgr"
 
 	"github.com/RussellLuo/slidingwindow"
-	"github.com/hashicorp/golang-lru/v2"
+	lru "github.com/hashicorp/golang-lru/v2"
```
davecheney, line 13: Sorry about this, my editor is insistent on adding this. It causes no changes in the rest of the file as /v2's package decl is `lru`.

jcalabro: Mine does the same (helix)

PR #1299, opened 2026-01-28
atproto/identity: unexport MockDirectory internals

davecheney:
MockDirectory.mu was not exported, by Handles and Indentities were. While I was not able to find any examples of this in indigo, this arrangement could permit a caller of MockDirectory to violate the lock invariant and reach into Handles/Identities directly.

Unexporting these two fields showed no breaks in any of the tests, which is excellent. If it was necessary to probe the internals of the MockDirectory, that could be done with additional methods exposed via a type asserting from Directory to *MockDirectory.

### DIFF
--- a/atproto/identity/mock_directory.go
+++ b/atproto/identity/mock_directory.go
@@ -12,8 +12,8 @@ import (
 // A fake identity directory, for use in tests
 type MockDirectory struct {
 	mu         sync.RWMutex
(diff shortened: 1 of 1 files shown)


bnewbold (approved):
I wrote this code, and don't remember why I exposed these publicly.

I checked in our other go repositories and don't see anything reaching in there, so this should be safe to merge.

PR #1300, opened 2026-01-28
Extract cmd/tap to service/tap and expose each tap component as first class citizen

davecheney:
This PR is an exploration of the idea _what would happen if tap became a service_. To explain this, let me lay a virtual end state.

- cmd/tap becomes the binary that is wrapped around the tap service
- service/tap, represents the components of the tap service. 
- other components, libraries; loggers, db, etc live elsewhere

Not pictured, but highly suggested, this pattern can be extended for other commands, cmd/relay, etc.

The big change here is not simply moving files around between packages, but, as expressed in the final commit, is tap.Tap is dissolved and the configuration responsibility moved to main.runTap. As expressed this is a little heavy handed, forcing the construction and gluing together of each component the responsibility of cmd/tap. But equally it demonstrates that alternative cmd/tap use cases are possible. 

Importantly this shows that, although service/tap components are wedded to the service/tap/models directory for the types to interact with the database, that is not actually a hard requirement. This opens the door to components like the crawler, the resyncer, etc moving to their own packages and becoming reusable components in their own right.

By removing the tap.Tap type it's demonstrated that the tap "service" could be sliced along various aspects. For example TapServer, which runs the admin interface and web socket requests, only has a weak dependency on crawler, firehose, etc, because of the stats reporting routes. The strong dependency is on Outbox, which depends on EventManager so that may provide a path to creating a multi process tap installation using the database as the point of co-ordination.

### DIFF
--- a/cmd/tap/main.go
+++ b/cmd/tap/main.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 
+	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/service/tap"
 	_ "github.com/joho/godotenv/autoload"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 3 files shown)


boltlessengineer:
@davecheney May I ask why this PR has been closed? Seems pretty useful.

PR #1301, opened 2026-01-28
remove old relay and supporting code: bigsky, bgs, pds

bnewbold:
This is roughly the scope of what can be removed now that the old relay (bigsky) is deprecated.

I might pull out one or two of these commits as separate PRs to merge first. Other chunks (like the handles, did, and plc top-level packages) can be removed after this.

My current plan is to create a fork of the indigo repo with just bigsky and it's supporting code, as a public archive. The code would also be in git history of this repo, but it is more visible in a fork.


jcalabro (approved):
sickosyes

PR #1303, opened 2026-01-28
copy pds/data models to testing/pdsmodels

bnewbold:
The motivation is to remove the dependency between the event persister tests and the pds/data package, so the later can be removed.


bnewbold:
(this is pretty low-stakes, i'm going to self-merge)

PR #1304, opened 2026-01-28
copy indexer.KeyManager to repomgr/keymgr

bnewbold:
The motivation is to clear the way to removing the 'indexer' directory. The only current non-deprecated usage is from cmd/supercollider.

We'll separately need to refactor KeyManager to use `identity.Directory` instead of the 'did' package.

### DIFF
--- a/repomgr/keymgr/keymgr.go
+++ b/repomgr/keymgr/keymgr.go
@@ -26,7 +26,7 @@ func NewKeyManager(didr DidResolver, k *did.PrivKey) *KeyManager {
 	return &KeyManager{
 		didr:       didr,
 		signingKey: k,
-		log:        slog.Default().With("system", "indexer"),
+		log:        slog.Default().With("system", "keymgr"),
 	}
 }
 


repomgr/keymgr/keymgr.go
```diff
@@ -0,0 +1,74 @@
...
+
+type DidResolver interface {
+	GetDocument(ctx context.Context, didstr string) (*did.Document, error)
+}
+
+func NewKeyManager(didr DidResolver, k *did.PrivKey) *KeyManager {
+	return &KeyManager{
+		didr:       didr,
+		signingKey: k,
+		log:        slog.Default().With("system", "indexer"),
```
davecheney: s/indexer/key_manager

repomgr/keymgr/keymgr.go
```diff
@@ -0,0 +1,74 @@
...
+	defer span.End()
+
+	k, err := km.getKey(ctx, did)
+	if err != nil {
+		return err
+	}
+
+	err = k.Verify(msg, sig)
+	if err != nil {
+		km.log.Warn("signature failed to verify", "err", err, "did", did, "pubKey", k, "sigBytes", sig, "msgBytes", msg)
```
davecheney, line 44: could this be log.Error?

bnewbold: I think i'd keep it as a warn: signatures failing to verify are something that can and will happen in the open network. They are abnormal and logging a bit louder than 'info' is good for debugging.

I'd want to keep 'error' level for truely unexpected scenarios or operational problems.

We don't have a written style guide around this and maybe should.

repomgr/keymgr/keymgr.go
bnewbold: good catch!

PR #1306, opened 2026-01-28
atproto/identity: return *CacheDirectory from constructor

davecheney:
This PR is sadly a breaking change, but probably one that is valuable to make. NewCacheDirectory was returning CacheDirectory, the value, not *CacheDirectory, the reference. This is problematic because CacheDirectory includes two sync.Map fields which should not be copied.

I'm not sure why returning the CacheDirectory from NewCacheDirectory by value did not trigger a linter warning -- returning a value is the same as passing a parameter, its a copy -- it could be a special case, but the result is the value return from NewCacheDirectory is tricky to use correctly.

Firstly it forces anyone who wants to use the result to take its address, because that is what satisfies the interface. Also, not taking the address and passing it as a parameter or assinging the result will trigger the lint warning
```
sync.Map contains sync.noCopy (govet)
```
I cannot find a defensible reason that NewCacheDirectory should continue to return a value, so I have updated it to return a pointer type which resolves a number of places which were complicated by not being able to take the address of a return value without first assinging it.

However, this remains a breaking change, and probably one that will be visible to downstream indigo users. I suspect they will welcome it, as I believe it is the correct thing to do, but never the less, this is not a transparent change.

### DIFF
--- a/atproto/identity/cache_directory.go
+++ b/atproto/identity/cache_directory.go
@@ -37,8 +37,8 @@ type identityEntry struct {
 var _ Directory = (*CacheDirectory)(nil)
 
 // Capacity of zero means unlimited size. Similarly, ttl of zero means unlimited duration.
-func NewCacheDirectory(inner Directory, capacity int, hitTTL, errTTL, invalidHandleTTL time.Duration) CacheDirectory {
-	return CacheDirectory{
+func NewCacheDirectory(inner Directory, capacity int, hitTTL, errTTL, invalidHandleTTL time.Duration) *CacheDirectory {
+	return &CacheDirectory{
 		ErrTTL:           errTTL,
 		InvalidHandleTTL: invalidHandleTTL,
 		Inner:            inner,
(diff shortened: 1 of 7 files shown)


bnewbold (approved):
I made this original mistake, and am in favor of the change.

We should probably update `NewMockDirectory` and `NewAPIDirectory` at the same time, for consistency (all implementing the `identity.Directory` interface). Could be the same PR, or paired PRs. I haven't checked, but I might have made similar mistakes elsewhere in `indigo:atproto/*`.

This is breakage enough that we should at least do a small announcement to our core go ecosystem devs before merging. I can handle that... might be good to roll up any other breaking changes as a mini-release?

I can also help update our dependent repos at the same time (it looks like several will be impacted).

davecheney:
> I made this original mistake, and am in favor of the change.
> 
> We should probably update `NewMockDirectory` and `NewAPIDirectory` at the same time, for consistency (all implementing the `identity.Directory` interface). Could be the same PR, or paired PRs. I haven't checked, but I might have made similar mistakes elsewhere in `indigo:atproto/*`.
> 
> This is breakage enough that we should at least do a small announcement to our core go ecosystem devs before merging. I can handle that... might be good to roll up any other breaking changes as a mini-release?
> 
> I can also help update our dependent repos at the same time (it looks like several will be impacted).

Ok, let me park this PR for the moment and cast a border net. It would probably be best to hit all these changes at the same time in one commit, and use that as the place to announce the fix.

jcalabro (approved):
yeah as long as we communicate it, this is actually a good and important thing to do I think

PR #1311, opened 2026-01-29
gosky: refactor use of 'plc' package

bnewbold:
This would allow removal of the 'did' and 'plc' top-level packages in indigo.

Not trivial to merge for two reasons:

- small deprecation here where gosky wants to parse signing (secret) key from JWK format.
- creates a repo-level import loop between go-didplc and indigo. I think go-didplc should potentially vendor the crypto and syntax bits it needs. though then there would be two crypto packages to work with, hrm.

Though I might move forward anyways to complete the 'did' and 'plc' cleanups.

### DIFF
--- a/util/cliutil/util.go
+++ b/util/cliutil/util.go
@@ -16,8 +16,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/bluesky-social/indigo/did"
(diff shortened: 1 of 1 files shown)


bnewbold:
handled this in a different PR

Issue #1313, opened 2026-01-29
atproto: some constructors return uncopyable values


davecheney:
As originally observed here https://github.com/bluesky-social/indigo/pull/1306, `atproto/identity.NewCacheDirectory` returns `CacheDirectory`. The issue is `CacheDirectory` contains several `sync.Map` fields which are marked `nocopy`. In practice all uses of this return value immediately take its address, because `*CacheDirectory` implements `identity.Directory`. `go vet` and other linters will spot any attempt to copy a `CacheDirectory` by value, ie `cdir := identity.NewCacheDirectory(..); somestruct.dir = cdir`, but we shouldn't lean on linters to catch this. I think we should correct the return type to be a reference.

@bnewbold suggested that there may be other occurrences which we should address in one commit because changing the return type is a breaking change, albeit a valuable one. An assay of atproto/ highights:

- `identity.NewCacheDirectory`
- `apidir.NewAPIDirectory` which requires the caller to refence the return value to fit the `APIDirectory` interface.
- `identify.NewMockDirectory`, as above, also contains a lock and map fields, which represents a correctness issue if the lock is copied.

- `syntax.NewTIDClock` returns `TIDClock`. `TIDClock` contains a `sync.Mutex` which protects the `lastUnixMicro` field. Copying a TIDClock copies the lock, which is a party foul, but it also copies lastUnixMicro. There are some questions about the correctness of the use of this type.

Additionally

- `atproto/lexicon.NewBaseCatalog` and `atproto/lexicon.NewResolvingCatalog` return by value. Inside indigo the result is used immediately, but returing by value means the result does not satisfy the `lexicon.Catalog` interface. This may be why `ResolvingCatalog` references its `Base` field as `BaseCatalog`, not the interface type.
- `identify.ParseIdentity` returns an `Identity` by value which is a moderately complex object. Benchmarking might be useful to determine if this is more expensive to copy vs gc the reference later. This may also apply to `Identity.DIDDocument`

davecheney:
@bnewbold https://github.com/bluesky-social/indigo/pull/1318 addresses the first three items.

PR #1318, opened 2026-01-30
identify.NewCacheDirectory, apidir.NewAPIDirectory, identity.NewMockDirectory return ptr values

davecheney:
Updates https://github.com/bluesky-social/indigo/issues/1313

This PR addresses the first three items recorded in https://github.com/bluesky-social/indigo/issues/1313. `identify.NewCacheDirectory`, `apidir.NewAPIDirectory`, and `identity.NewMockDirectory` now return a reference to their respective values. In all cases this is a quality of life improvement, to implement the `identify.Directory` and `identity.Resolver` interfaces, a pointer receiver is required, forcing the caller to store, then reference the temporary value.

In the case of `identify.NewCacheDirectory` and `identity.NewMockDirectory`, returning by value was a correctness issue as those types contain fields which should not be copied.

### DIFF
--- a/atproto/atclient/password_auth_test.go
+++ b/atproto/atclient/password_auth_test.go
@@ -131,7 +131,7 @@ func TestPasswordAuth(t *testing.T) {
 
 	{
 		// simple GET requests, with token expire/retry
-		c, err := LoginWithPassword(ctx, &dir, syntax.Handle("user1.example.com").AtIdentifier(), "password1", "", nil)
+		c, err := LoginWithPassword(ctx, dir, syntax.Handle("user1.example.com").AtIdentifier(), "password1", "", nil)
 		require.NoError(err)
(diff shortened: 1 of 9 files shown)


cmd/bluepages/main.go
```diff
@@ -184,7 +186,12 @@ func configLogger(cmd *cli.Command, writer io.Writer) *slog.Logger {
 	return logger
 }
 
-func configClient(cmd *cli.Command) apidir.APIDirectory {
+type directory interface {
```
bnewbold, line 189: nice! might fold some form of this in to the identity package itself at some point.

davecheney: Yup, it's quite a common pattern. Most places that take a directory & identity shaped thing end up using the concrete type because neither Directory nor Identity describe their use case completely.

PR #1322, opened 2026-02-01
per PDS backfill ratelimiting

whyrusleeping:
A global rate limit is a little bad, so I added some retries on 429's, and also a per-PDS rate limiter


backfill/backfill.go
```diff
@@ -118,7 +123,9 @@ type BackfillOptions struct {
 	ParallelRecordCreates int
 	NSIDFilter            string
 	SyncRequestsPerSecond int
-	RelayHost             string
+	// Per-PDS rate limit (requests per second per host). If 0, defaults to 2.
```
bnewbold: this logic is a bit weird to me, i'd expect the default to be 2, but 0 to mean no limit?

backfill/backfill.go
```diff
@@ -338,37 +373,66 @@ func (b *Backfiller) fetchRepo(ctx context.Context, did, since, host string) (io
...
+		}
+
+		req.Header.Set("Accept", "application/vnd.ipld.car")
+		req.Header.Set("User-Agent", fmt.Sprintf("atproto-backfill-%s/0.0.1", b.Name))
+		if b.magicHeaderKey != "" && b.magicHeaderVal != "" {
+			req.Header.Set(b.magicHeaderKey, b.magicHeaderVal)
+		}
+
+		// Wait on global rate limiter
+		b.syncLimiter.Wait(ctx)
```
bnewbold: should handle the possible `err` returned here (and two lines below). those would happen if the ctx is cancelled. otherwise the error will come from the `client.Do()`, and it would naively look like an HTTP request timeout.

backfill/backfill.go
```diff
@@ -338,37 +373,66 @@ func (b *Backfiller) fetchRepo(ctx context.Context, did, since, host string) (io
...
-		return nil, &FetchRepoError{
+		if resp.StatusCode == http.StatusOK {
+			instrumentedReader := &instrumentedReader{
+				source:  resp.Body,
+				counter: backfillBytesProcessed.WithLabelValues(b.Name),
+			}
+			return instrumentedReader, nil
+		}
+
+		resp.Body.Close()
```
bnewbold, line 454: this is the HTTP non-success status code path. I think you need to drain the body here, or the connection can't be reused (eg, HTTP/2), which is relevant b/c you're probably making many requests to a PDS, and there will be some suspended/deleted accounts.

could also be nice to fetch error response JSON body and log it or something. doesn't need to happen in this PR but a TODO comment would help track it.

PR #1323, opened 2026-02-03
cmd/tap: small string optimizations

James-Gilbert-:
Slight speed improvements in utility functions

 - `matchesCollection()`: using CutSuffix instead of HasSuffix + TrimSuffix (~23% faster)
 - `userAgent()`: caching via package-level var (avoids Short()'s slice allocs per call)


dholms (approved):
This looks good. Thanks!

PR #1324, opened 2026-02-03
add arm64 to tap docker build

axfelix:
Need to kick off the CI to test this, I *think* this is all that's needed


axfelix:
Well OK, it is definitely a bit slow now, but this did work. We would probably not want QEMU arm64 builds for the entire repo to take this long, I think we should invest in dedicated runners if we need them for *more than* just tap (pending indigo repo decoupling). Right now this is all I've heard the community ask for.

PR #1344, opened 2026-02-25
relay: `Allow-Methods: GET, POST` for CORS requestCrawl

uniphil:
Motivation: web-based PDS tooling like https://debug.hose.cam is currently unable to offer sumitting a `requestCrawl` directly, due to CORS restrictions (which may or may not be intentional?)

Preflight response headers from the relay include `Allow: GET, POST`, but lack an `Access-Control-Allow-Methods` header permitting the same, causing POST requests (like `requestCrawl`) to fail from browsers.

The `echo` CORS middleware seems to imply that preflights should have the cors header set,

>  If `allowMethods` is left empty, this middleware will fill for preflight request `Access-Control-Allow-Methods` header value from `Allow` header that echo.Router set into context.
>
> _—echo's [middleware/cors.go](https://github.com/labstack/echo/blob/1753170a74959596a69735c553f3fe5a4bd07715/middleware/cors.go#L64-L66)_

but, this doesn't happen for requests to the relay:

```bash
$ curl -iXOPTIONS https://bsky.network/xrpc/com.atproto.sync.requestCrawl

HTTP/2 204
allow: OPTIONS, POST
vary: Origin
date: Wed, 25 Feb 2026 15:48:29 GMT
strict-transport-security: max-age=6307200
```

I haven't dug in further about it within echo.

This change explicitly configures the allowed methods CORS header to include `GET` and `POST`, allowing web-based PDS tooling to submit `requestCrawl` requests to relays.


bnewbold:
I'm at least on the fence about this.

"Request Crawl" was originally intended to be a request that a PDS would make to a relay, not to allow any party on the internet to tell the relay to reach out any make HTTPS and WebSocket requests to any arbitrary other host on the internet. If we had an existing way to authenticate requests between PDS and relay, I wouldn't be against that.

Certainly there needs to be a way for relay operators to add arbitrary PDS instances, but I think the existing admin API endpoints allow for that.

It would be helpful for individual accounts to be able to poke at the PDS/relay connection and request retries. My intuition is that this should be an authenticated endpoint on the PDS (maybe in the PDS web interface), and should pair with the ability to emit no-op `#identity`, `#account`, and `#sync` events on a per-account basis.

It's already possible for a bad actor to spam relays with crawl requests (which then result in the relay sending out many requests, which could get the relay rate-limits / blocked / in trouble). But making that possible from a web browser reduces friction further. As an ecosystem we've gotten by with fairly low friction, but there are folks poking at all this, and I think we should continue to balance concerns and retain a bit of friction.

uniphil:
i agree with everything except the friction argument.

the existing (well known i think? trivial) public relay requestcrawl DOS (and server-side-request driving stuff) is already very-low-friction, and i'm not sure i see how this makes it much lower.

to me the friction balance is backwards right now: debug.hose.cam mostly gets frequent use from **PDS operators**, who will click a button to fix their own PDS being eg., missing on third-party relays, but the friction of leaving to figure out how to do requestcrawls by other means will stop many of them, leaving the network in a worse state.

this change tries to slightly shift the balance of friction for today's (unfortunate) requestcrawl world in favour of indie operators. it's definitely not a long-term-thinking move.

(agreed the existing authenticated endpoint for _relay admins_ is already fine)

an authenticated PDS endpoint makes a lot of sense, but i hope it's not designed to *only* be part of the PDS's own interface. a tight oauth scope (or even a redirect with relay hint param) for tools like debug.hose.cam would be great. (otherwise it would be following the same bluesky-privileging pattern as requestcrawl imo -- would bluesky ever ship a relay status check that knows about `relay.fire.hose.cam` in the reference pds? etc)

---

anyway this is probably already too much bikeshedding! if this is rejected, i'd like to follow up to remove `POST` from preflight, so it doesn't look like an accidental restriction as it currently does.

PR #1345, opened 2026-02-25
identity.CacheDirectory should not cache context.Canceled / context.DeadlineExceeded errors

arushibandi:
I'm using the `identity.DefaultDirectory()` and noticed that I was seeing failed lookups due to `context cancelled`. This also seemed to coincide with logins/oauth redirect flows or refreshed. Restarting my process fixed this. After digging in, it seems that this is because the CacheDirectory caches all errors, not filtering for context cancelled / deadline exceeded.

My workaround for this is always pass in `context.Background()` into the `dir.Lookup()` function, but this behavior seems unintended.

The tests fail before the fix and pass after.

closes https://github.com/bluesky-social/indigo/issues/1346


atproto/identity/cache_directory.go
arushibandi, line 218: this should not be cached right?

PR #1367, opened 2026-04-13
automod: limit blob processing concurrency per record

bnewbold:
The motivation is that we are looking at doing a "gallery" embed type on bsky posts, and want to constrain the increase in concurrent goroutines. In particular, the number of concurrent HTTP requests being made by automod to PDS instances to download the blobs: we don't want automod to get rate-limited if it is doing dozens of concurrent blob requests to a small/simple PDS due to a single record; or to DDoS that type of PDS instance if it does not have limits on concurrent requests in place.

Note that this patch does *not* add concurrency limits on processing rules for individual blobs. If an automod process had 20 blob rules, and concurrent blob processing of 8x, it might start 160 processing goroutines at the same time. This is expected to be fine because blob rules are usually "per service", and these requests would be going to separate backend services. The best way to constrain concurrency there is to limit the overall number of worker queues.

Note that i'm not using `errgroup.WithContext` (doesn't seem needed), and that i'm doing the old-fashioned `blob := blob` variable re-assignment in the for-loops out of habit, even though loop vars were fixed in Go 1.22.

Mostly requesting review from Rafael, but also tagging Jim for review (if he has time) just for idiomatic Go concurrency patterns.

### DIFF
--- a/automod/engine/ruleset.go
+++ b/automod/engine/ruleset.go
@@ -3,11 +3,14 @@ package engine
 import (
 	"bytes"
 	"fmt"
-	"sync"
 
 	appbsky "github.com/bluesky-social/indigo/api/bsky"
 	lexutil "github.com/bluesky-social/indigo/lex/util"
+
+	"golang.org/x/sync/errgroup"
 )
+
+var DEFAULT_BLOB_PROCESSING_CONCURRENCY = 8
 
 // Holds configuration of which rules of various types should be run, and helps dispatch events to those rules.
 type RuleSet struct {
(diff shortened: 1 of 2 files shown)


jcalabro (approved):

automod/engine/ruleset.go
```diff
@@ -121,60 +124,36 @@ func (r *RuleSet) fetchAndProcessBlobs(c *RecordContext) error {
...
-	// check for errors
-	for err := range errChan {
-		if err != nil {
-			return err
-		}
+			return r.processBlob(c, blob, data)
+		})
 	}
-	return nil
+	return wg.Wait()
```
rafaeleyng, line 144: This keeps the same behavior of returning the first non-nil error. I just want to ensure that's the behavior we want here (e.g., instead of returning an aggregate error).

automod/engine/ruleset.go
```diff
@@ -121,60 +124,36 @@ func (r *RuleSet) fetchAndProcessBlobs(c *RecordContext) error {
...
+	var wg errgroup.Group
+	for _, brf := range r.BlobRules {
+		brf := brf
+		wg.Go(func() error {
+			return brf(c, blob, data)
+		})
 	}
-	return nil
+
+	return wg.Wait()
```
rafaeleyng, line 158: Same question.

rafaeleyng (approved):

automod/engine/ruleset.go
bnewbold: I think this is fine (and consistent with how it was working before). The important thing is to attempt processing all the blobs, and not interrupt/cancel on the first error encountered.

PR #1368, opened 2026-04-14
atproto/auth: more flexible 'aud' validation; stricter 'lxm' validation

bnewbold:
NOTE: this contains a minor breaking API change: public struct field renamed and changed from string to array

These are updates to the service auth SDK to support more specific/flexible audience (`aud`) validation, verifies the `kid` header field (if present), and makes `lxm` field non-optional.

Here is the proposal this relates to: https://github.com/bluesky-social/proposals/tree/main/0014-service-auth-revised

This SDK already only supported a bare DID in the `iss` field. This PR does not include support for configurable `kid` values and key lookup in the DID document. That can be added later as a non-breaking change.

Some things to consider in reivew:
- is the degree of API breakage here acceptable given the current (in)stability of this package/repo?
- should I add issuer `kid` configurable handling now? adds complexity/delay and like 99.9% of services should just be accepting `#atproto` (or empty, the default)
- might it be important to pass back either the full `aud` or the issuer `kid` in addition to the issuer DID? eg, calling code (service) might care about that context


bnewbold:
~hrm, I'm remembering now that the way we do feedgen service auth is weird/janky and this package probably needs to support that. In those cases the audience doesn't match, and the `lxm` might not match either (eg, `getFeed` vs `getFeedSkeleton`)~

(actually this should work fine on the feedgen side; it is PDS instances and to a lesser degree AppViews which need hacks for the feedgen double-proxy pattern)

PR #1389, opened 2026-05-15
Add standard site lexicons

estrattonbailey:
Adds new `make lexgen-ext` and corresponding `ext.json` config for generating external lexicons and their dependencies. It seems that it's necessary to include the `atproto` lexicons within this config file in order for them to be resolved when building the external lexicons.

Two main decisions needed:
- Do we want a separate `ext.json` config or should we chuck it all in the `bsky.json` config?
- How's the naming convention of `standardsite` here? Any better suggestions?

Also important note: in order to get this to build, I needed to add a `#placeholder` def to the `document`'s `content` field. This does not match real-world use cases of course, but for our current purposes, we are ignoring this field. Anyone else consuming these types will still be able to treat the `content` field as an open union, and would need to install other lexicons as appropriate e.g. Leaflet's content data structures.


cmd/lexgen/ext.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "package": "atproto",
+    "prefix": "com.atproto",
+    "outdir": "api/atproto",
+    "import": "github.com/bluesky-social/indigo/api/atproto"
+  },
```
blackmichael, line 7: Should this be removed since it's in `bsky.json` already?

estrattonbailey: The build fails without it, since SS lexicons rely on atproto lexicons. I don't think our lexgen command supports anything to help with this, but let's run it by some more folks this week, might be unnecessary.

devinivy (approved):

PR #1406, opened 2026-06-13
lexlint: demote unlimited-string and unlimited-bytes to 'info' level

bnewbold:
These warnings show up a ton with `goat lex lint` and I feel are fine to ignore, so demoting them from "warn" to "info".

This will pair with an update to `goat` to allow filtering by level and ignoring specific names.

The other one we could consider demoting is "endpoint-output-undefined".

Tagging divy as a lexicon style thinker.


devinivy:
I'm cool with that.  An alternative is to not output warns by default, and only worry about errors.  The warns definitely create noise for us right now in CI (e.g. see https://github.com/bluesky-social/bsky/pull/26#issuecomment-5281916701), particularly because fixing the lint issue is often going to be a breaking change that just isn't going to happen.  Most notably in the case of records.

> The other one we could consider demoting is "endpoint-output-undefined".

When I am creating a lexicon for the first time I definitely want at least a warn on this one— it can be a bit of a trap to omit it by mistake since adding one later may be considered breaking.  Permitting long strings isn't quite a trap in the same way that this one is.

One pattern I'm noticing— I want stricter linting when I am creating a lexicon for the first time versus editing a published one.  When editing a published one, I don't want to be asked to make breaking changes.

PR #1421, opened 2026-07-22
cmd/tap: Use is_active instead of isActive in README

lilasta:
Update the identity event example in `cmd/tap/README.md` to use `is_active` instead of `isActive`.

This matches the JSON tag on `IdentityEvt` in `cmd/tap/types.go`:
https://github.com/bluesky-social/indigo/blob/dfe5578fd537e71dd1ed884dd09b876dd768dfd5/cmd/tap/types.go#L55

(Sorry, Copilot left an unintended review comment because of my configuration mistake. Please ignore it)


bnewbold (approved):
thanks!

PR #1428, opened 2026-08-11
Adds content visibility lexicon

cuducos:
See bluesky-social/atproto#5372


vineyardbovines (approved):

gen/main.go
```diff
@@ -73,6 +73,7 @@ func main() {
 		bsky.FeedPostgate_DisableRule{},
 		bsky.GraphVerification{},
 		bsky.ActorStatus{},
+		bsky.ActorContentVisibility{},
```
vineyardbovines: oh- should we change this to `bsky.ActorContentVisibilityDeclaration`?

cuducos: 100% missed that
tks (fixing it now)

PR #1449, opened 2026-09-01
backfill: route repo fetches through a jttp client

treethought:
Repo fetches now use a jttp client that skips private and reserved network targets. UnsafeDisableSSRF allows them for local development.


bnewbold (approved):
nice, thanks for the changes from earlier

PR #1452, opened 2026-09-01
harden identity package

bnewbold:
Adds some extra safety checks to `identity.DefaultDirectory` and `BaseDirectory`, and adds some doc comments.

These change *do* have behavioral impact for code which uses this package. By splitting out the PLC http client we hope the most common local dev impacts will be limited. But folks with localhost did:web service accounts might encounter issues. 

Updates to other services which use the identity package will be a separate PR.

Note that we intend to switch to [jttp](https://github.com/jcalabro/jttp) soon for some of this HTTP hardening.

### DIFF
--- a/atproto/identity/did.go
+++ b/atproto/identity/did.go
@@ -14,6 +14,9 @@ import (
 	"github.com/bluesky-social/indigo/atproto/syntax"
 )
 
+// 5 MByte as a somewhat arbitrary value to start (should probably be smaller)
+var maxDIDBodySize int64 = 5 * 1024 * 1024
+
 // Resolves a DID to a parsed `DIDDocument` struct.
 //
 // This method does not bi-directionally verify handles. Most atproto-specific code should use the `identity.Directory` interface ("Lookup" methods), which implement that check by default, and provide more ergonomic helpers for working with atproto-relevant information in DID documents.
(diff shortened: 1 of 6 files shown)


atproto/identity/did.go
```diff
@@ -116,15 +119,15 @@ func (d *BaseDirectory) resolveDIDWeb(ctx context.Context, did syntax.DID) ([]by
...
 		return nil, fmt.Errorf("%w: did:web HTTP status 404", ErrDIDNotFound)
 	}
 	if resp.StatusCode != http.StatusOK {
-		io.Copy(io.Discard, resp.Body)
+		io.Copy(io.Discard, io.LimitReader(resp.Body, maxDIDBodySize))
 		return nil, fmt.Errorf("%w: did:web HTTP status %d", ErrDIDResolutionFailed, resp.StatusCode)
 	}
 
-	return io.ReadAll(resp.Body)
+	return io.ReadAll(io.LimitReader(resp.Body, maxDIDBodySize))
```
treethought, line 137: Just noting using a LimitReader here will cause a truncated response to be returned rather than an error. Noticed we do this elsewhere in the repo, and a valid did document wouldn't be this large, but may be good to review these LimitReader cases later

atproto/identity/directory.go
```diff
@@ -63,15 +64,20 @@ var DefaultPLCURL = "https://plc.directory"
...
 	base := BaseDirectory{
 		PLCURL: DefaultPLCURL,
 		HTTPClient: http.Client{
 			Timeout: time.Second * 10,
 			Transport: &http.Transport{
-				Proxy: http.ProxyFromEnvironment,
-				// would want this around 100ms for services doing lots of handle resolution. Impacts PLC connections as well, but not too bad.
-				IdleConnTimeout: time.Millisecond * 1000,
-				MaxIdleConns:    100,
+				Proxy:             http.ProxyFromEnvironment,
```
treethought, line 75: may be worth a comment saying the ssrf dialer only validates the proxy if one is set

treethought (approved):

atproto/identity/directory.go
bnewbold: hrm, yeah it is kind of convoluted (the DialContext protections apparently do apply to the connection to the HTTP proxy itself?). I guess we could have custom code which tries to check if a proxy is being invoked (and skip the SSRF stuff), but that feels fraught.

I at least added a doc string mentioning this for now.

atproto/identity/did.go
bnewbold: mmm!

I added a check on `Content-Length`, and opened an issue here to track the more subtle corner-case:
https://github.com/bluesky-social/indigo/issues/1460

bnewbold:
I rebased on main to get the `gttp` move/rename, and did a force-push

treethought (approved):

PR #1456, opened 2026-09-01
harden more untrusted network fetches

bnewbold:
This covers a number of network fetches to untrusted hosts, other than `identity.Directory` usage (covered in other PRs).

These changes may have an impact on downstream code. For example, the OAuth change could impact some local/offline development workflows (though it should not break simple development with "localhost" `client_id`).

This only uses `jttp` a bit; future work is to use that more broadly.

### DIFF
--- a/atproto/lexicon/resolve.go
+++ b/atproto/lexicon/resolve.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"net/http"
+	"time"
 
 	"github.com/bluesky-social/indigo/api/agnostic"
 	"github.com/bluesky-social/indigo/atproto/atclient"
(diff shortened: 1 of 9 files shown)


atproto/lexicon/resolve.go
```diff
@@ -73,7 +74,9 @@ func resolveLexiconJSON(ctx context.Context, dir identity.Directory, nsid syntax
 func fetchRecordJSON(ctx context.Context, ident identity.Identity, aturi syntax.ATURI) (*json.RawMessage, error) {
 
 	slog.Debug("fetching record", "did", ident.DID.String(), "collection", aturi.Collection().String(), "rkey", aturi.RecordKey().String())
+	// the this is an untrusted endpoint URL
 	client := atclient.NewAPIClient(ident.PDSEndpoint())
+	client.Client.Transport = ssrf.PublicOnlyTransport()
```
treethought: client.Client here is the `http.DefaultClient`, so this applies the new transport globally for requests made with DefaultClient

bnewbold: good catch! fixed by creating a fresh `http.Client`. might be nice to have a shared/global-ish safe client within some of these packages (eg, internal to identity pkg and atclient pkg)

treethought: yeah think that's a good idea 👍. I think there are a few places where we are creating a fresh client, when it'd make sense to use an existing shared one

bnewbold:
I rebased on main to get the `gttp` move/rename, and did a force-push

atproto/lexicon/resolve.go
bnewbold: tracking issue: https://github.com/bluesky-social/indigo/issues/1461

bnewbold:
CI failures were sporadic external network failures. I retried them and all green now.

treethought (approved):

PR #1458, opened 2026-09-01
oauth: validate all auth server endpoints as HTTPS URLs

treethought:
Previously only the authorization endpoint was checked for being a well-formed HTTPS URL. Now the pushed-authorization, token, and optional revocation endpoints are validated too, rejecting non-HTTPS schemes, query/fragment components, and hostless opaque forms like "https:example.com".


bnewbold (approved):
nice, and thanks for adding test coverage

PR #1459, opened 2026-09-01
bluepages: return generic identity error messages to reduce verbosity

treethought:
Handle and DID resolution failures now log the underlying error server-side and return a generic message to clients.


bnewbold (approved):
nice, thanks

PR #1462, opened 2026-09-02
mst: validate PrefixLen range when decoding node data

haileyok:
## Changes
- `atproto/repo/mst/encoding.go`: `Node()` signature changed from `Node(c *cid.Cid) Node` to `Node(c *cid.Cid) (Node, error)`; each entry's `PrefixLen` is checked against the previous key length before being used as a slice bound, returning a wrapped `ErrInvalidTree`. Caller `loadNodeFromStore` propagates the error
- `mst/mst_util.go` (`deserializeNodeData`): same range check before the `make()` preallocation
- One test caller updated for the new `Node()` signature (`mst_interop_test.go`).
- No behavior change for well-formed data: `PrefixLen == 0` (first entry) and `PrefixLen == len(prevKey)` (later entries) remain valid, covered by boundary tests.

## Review notes
- `Node()` is an exported API change (new error return) — any external consumers of `atproto/repo/mst` will need a mechanical update; in-repo callers are all updated here.
- Defense-in-depth follow-up (not in this PR): a `recover()` in the relay parallel scheduler's worker loop so future parser panics degrade to dropped events rather than process death. Deliberately excluded to keep this fix minimal and reviewable.

### DIFF
--- a/atproto/repo/mst/encoding.go
+++ b/atproto/repo/mst/encoding.go
@@ -90,7 +90,10 @@ func (n *Node) NodeData() NodeData {
 // Tansforms an encoded `NodeData` to `Node` data structure format.
 //
 // c: optional CID argument for the CID of the CBOR representation of the NodeData
-func (d *NodeData) Node(c *cid.Cid) Node {
+//
+// Returns an error if any entry's PrefixLen is out of range, since NodeData
+// comes from untrusted input and PrefixLen is used as a slice bound.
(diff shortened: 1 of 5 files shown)


bnewbold (approved):
looks good, thanks for hitting both versions with this patch.

this technically breaks the API surface, but this is a relatively internal method so I don't think it should be a problem.

PR #1468, opened 2026-09-02
automod: improve account metadata fetch logic

bnewbold:
This PR changes the logic and codeflow for how the automod engine fetches account metadata from ozone and/or the bsky appview (with rare fallback to an entryway API).

The motivation for these changes are to be resilient to failure to fetch info from the bsky appview in particular. It is not uncommon for the appview to return and HTTP error because an account is suspended, but the account continues to emit events. In these situations, the metadata from ozone is sufficient.

At the same time, this keeps things working in situations where an ozone instance is not configured.

Adds a couple additional metrics, and drops a "new account retry" sleep down to 2 seconds. this is also now against ozone, so we don't expect failures if the account is simply suspended; and will end up caching the overall fetch as long as the appview does respond (for example).

For reviewers: a concern to keep in mind is that the account metadata cache has an expiration of 24 hours. Having bad/empty metadata cached could be bad in some situations. Ozone events and account/identity events can purge that cache for individual accounts. And not all the metadata is critical (eg, the profile info isn't often used in automod rules; profile records themselves have rules on the record themselves).

### DIFF
--- a/automod/engine/fetch_account_meta.go
+++ b/automod/engine/fetch_account_meta.go
@@ -15,15 +15,15 @@ import (
 	"github.com/bluesky-social/indigo/xrpc"
 )
 
-var newAccountRetryDuration = 3 * 1000 * time.Millisecond
+var newAccountRetryDuration = 2 * 1000 * time.Millisecond
 
 // Helper to hydrate metadata about an account from several sources: PDS (if access), mod service (if access), public identity resolution
 func (e *Engine) GetAccountMeta(ctx context.Context, ident *identity.Identity) (*AccountMeta, error) {
 
 	logger := e.Logger.With("did", ident.DID.String())
 
-	// fallback in case client wasn't configured (eg, testing)
-	if e.BskyClient == nil {
+	// fallback in case upstream clients are not configured (eg, testing)
+	if e.BskyClient == nil && e.OzoneClient == nil {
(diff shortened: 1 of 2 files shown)


automod/engine/fetch_account_meta.go
```diff
@@ -152,42 +110,122 @@ func (e *Engine) GetAccountMeta(ctx context.Context, ident *identity.Identity) (
...
+		}
+	}
+
+	// fetch other account metadata from bsky AppView
+	// this only gets public info, which can include profile contents as well as public labels
+	pv, err := appbsky.ActorGetProfile(ctx, e.BskyClient, ident.DID.String())
+	if err != nil {
+		logger.Warn("account profile lookup failed (from bsky appview)", "err", err)
+		var xrpcError *xrpc.Error
+		if errors.As(err, &xrpcError) {
```
treethought, line 172: Just noting the PR says dropped the retry sleep to 2 seconds, but the retry is actually removed

treethought (approved):

automod/engine/fetch_account_meta.go
bnewbold: I moved the retry logic from the Appview API request (where an HTTP 4xx might mean a takedown) to the Ozone API request (where an HTTP 4xx should only mean unknown/unseen account, not a takedown).

PR #1469, opened 2026-09-03
mst: reword recent comments and error message

bnewbold:
This PR doesn't change behaviors, just rephrases some copy text.


haileyok (approved):
thanks!

